PR #1, opened 2017-12-18
Tracker Networks Top Offenders - Pill Version

MariagraziaAlastra:
**Reviewer:** @laurengarcia 

## Description:
Updating top offenders card in the main popup with the new pills UX.
Continuing from [here](https://github.com/duckduckgo/chrome-zeroclickinfo/pull/232)

screen shot 2017-12-19 at 10 59 00 pm
screen shot 2017-12-19 at 10 57 20 pm

## Steps to test this PR:
1. run make
2. load or reload extension in Chrome or FF
3. after browsing 10 websites, the main popup should show 3 pills with the top offenders icons
4. top offenders subview should be unchanged

## Automated tests:
- [ ] Unit tests
- [ ] Integration tests

###### Reviewer Checklist:
- [ ] **Ensure the PR solves the problem**
- [ ] **Review every line of code**
- [ ] **Ensure the PR does no harm by testing the changes thoroughly**
- [ ] **Get help if you're uncomfortable with any of the above!**
- [ ] Determine if there are any quick wins that improve the implementation

###### PR Author Checklist:
- [ ] Get advice or leverage existing code
- [ ] Agree on technical approach with reviewer (if the changes are nuanced)
- [ ] Ensure that there is a testing strategy (and documented non-automated tests)
- [ ] Ensure there is a documented monitoring strategy (if necessary)
- [ ] Consider systems implications

PR #3, opened 2017-12-19
Privacy practices and Grade Scorecard

andrey-p:
**Reviewer:** @laurengarcia 

## Description:

Adding the privacy practices view, still WIP.

screenshot from 2017-12-27 15-10-40

## Steps to test this PR:

1. Remove the `is-hidden` on this line: https://github.com/duckduckgo/duckduckgo-privacy-extension/blob/7dbadaa5526ca9c0eaf41d81796149fba98109c1/shared/js/ui/templates/site.es6.js#L50
2. Visit a bunch of websites (e.g. steampowered.com, gmail.com, duckduckgo.com, etc) and open the privacy practices subview to see what this looks like.

## Automated tests:
- [ ] Unit tests
- [ ] Integration tests

###### Reviewer Checklist:

###### PR Author Checklist:


shared/scss/popup.scss
```diff
@@ -666,6 +666,8 @@ body {
     }
 }
 
+@import "views/privacy-practices";
```
andrey-p: I added this new `views` folder where we can place styles that are specific to views. It feels messy to have them all in the bigger files. What do you think?

Also, I tried to minimize nesting as much as possible, to get the full benefit out of using BEM. Should probably try and enforce that elsewhere.

shared/js/ui/views/privacy-practices.es6.js
```diff
@@ -0,0 +1,44 @@
...
+    this.setupClose();
+    this.renderAsyncContent();
+}
+
+PrivacyPractices.prototype = $.extend({},
+    ParentSlidingSubview.prototype,
+    {
+        renderAsyncContent: function () {
+            const random = Math.round(Math.random()*100000)
+            this.currentModelName = 'site' + random
```
andrey-p: I copied this from the `GradeDetails` view. The prospect of collisions worries me, this is bound to cause a problem for someone at some point.

We've got a bit of code in the core code that generates a UUID for each model as it's created. Should we port this to here?

shared/js/ui/views/privacy-practices.es6.js
```diff
@@ -0,0 +1,44 @@
...
+            this.model = new SiteModel({
+                modelName: this.currentModelName
+            })
+
+            this.model.getBackgroundTabData().then(() => {
+                this.$overview.html(overviewTemplate(
+                    this.model.domain,
+                    this.model.tosdr
+                ))
+                this.$details.html(detailsTemplate(this.model.tosdr))
```
andrey-p: Again, I looked at the `GradeDetails` view as an example when putting this together. I decided to handle things in a different way, though. The pattern used there (use the template twice, once to render the view, once as a thing to append) was really unclear to me at first. It seems unnecessarily coupled to the implementation of the view, and when the model is created.

What do you think to this approach?

laurengarcia: Yeah its a better pattern for sure. Alot of the UI was thrown together quite quickly, improvements welcome.

shared/scss/views/_privacy-practices.scss
```diff
@@ -0,0 +1,47 @@
+.privacy-practices {
+    min-height: 100%;
+    box-sizing: border-box;
```
laurengarcia: This made me realize that our normalize file is not making border-box the default. Maybe I should do that. Hm.

shared/scss/views/_privacy-practices.scss
```diff
@@ -0,0 +1,125 @@
...
+
+.privacy-practices__overview__icon {
+    width: 48px;
+    height: 65px;
+    margin: 0 auto;
+    background-repeat: no-repeat;
+}
+
+// TODO dry these up also
+.privacy-practices__overview__icon--good {
```
MariagraziaAlastra: @andrey-p Oh now I know what you meant earlier. Yes, this is a good usecase for doing something like [this](https://github.com/duckduckgo/duckduckgo-privacy-extension/pull/1/files#diff-2dff5e5386b765d8477d1bbb024acc20)

shared/scss/views/_privacy-practices.scss
```diff
@@ -0,0 +1,125 @@
...
+    .is-browser--chrome & {
+        background-image: url($icon__hero_ribbon__good--chrome);
```
MariagraziaAlastra: Actually this could be improved even further. I see a lot of these .is-browser--chrome and .is-browser--firefox around, we should probably have a mixin for browser-specific styles...
In addition to that, we could put the base extension URLs in a variable, so instead of having 30 variations of $icon-1--chrome and $icon-1--firefox, we'd just pass along the icon name and we'd have something like:
<pre>
@mixin browser_icons_path($icon-name) {
     .is-browser--moz & { 
       background-image: url($moz-url$icon-name);
    }
    .is-browser--chrome & { 
       background-image: url($chrome-url$icon-name);
    }
}
</pre>

Just an idea, and not critical ofc, but I think it'd save us time with the other layouts

shared/scss/_mixins.scss
```diff
@@ -150,6 +150,16 @@
...
+/* Cross-platform background image */
+@mixin cross_platform_bg_image ($image) {
+    $asset-path: "__MSG_@@extension_id__/img/";
+    .is-browser--chrome & {
+        background-image: url("chrome-extension://#{$asset-path}#{$image}");
+    }
+    .is-browser--moz & {
+        background-image: url("moz-extension://#{$asset-path}#{$image}");
+    }
+}
```
andrey-p: @MariagraziaAlastra I only just saw your comment - great minds think alike as they say :smile:

shared/js/site.js
```diff
@@ -52,10 +53,41 @@ class Score {
...
+
+                let reasons = tosdrData.match
+
+                if (reasons.good) {
+                    reasons.good = reasons.good.map(utils.capitalizeFirstLetter)
+                }
+
+                if (reasons.bad) {
+                    reasons.bad = reasons.bad.map(utils.capitalizeFirstLetter)
+                }
```
andrey-p: I'm not sure if this stuff is better done in the script that generates TOS;DR data in the first place.

shared/js/site.js
```diff
@@ -52,10 +53,41 @@ class Score {
...
+                        message = constants.tosdrMessages.mixed
+                    } else if (tosdrData.score > 0 ) {
+                        message = constants.tosdrMessages.bad
+                    }
+                }
+
+                let reasons = {
+                    good: matchGood.map(utils.capitalizeFirstLetter),
+                    bad: matchBad.map(utils.capitalizeFirstLetter)
+                }
```
andrey-p: The capitalizing could probably also be done in the script, where `tosdr.json` is generated.

shared/js/ui/templates/privacy-practices.es6.js
```diff
@@ -0,0 +1,32 @@
...
+            <div class="privacy-practices__explainer padded border--bottom--inner">
+                Privacy practices indicate how much the personal information
+                that you share with a website is protected.
+            </div>
+            <div class="privacy-practices__details padded border--bottom--inner
+                js-privacy-practices-details">
+                ${details(tosdr)}
+            </div>
+            <div class="privacy-practices__attrib padded">
+                Privacy Practice results from <a href="https://tosdr.org/" target="_blank">ToS;DR</a>
```
andrey-p: This formatting of this deviates from the design mockups but fits how they do it on their own website: https://tosdr.org/about.html

shared/scss/views/_privacy-practices.scss
```diff
@@ -0,0 +1,132 @@
...
+    .padded {
+        padding: 20px;
+    }
+}
+
+.privacy-practices__overview,
+.privacy-practices__explainer,
+.privacy-practices__attrib {
+    text-align: center;
+}
```
andrey-p: TODO use util class for this

shared/js/ui/views/privacy-practices.es6.js
andrey-p: As I was working on https://github.com/duckduckgo/duckduckgo-privacy-extension/pull/7 I realized I can just pass down the global site model here. I'll rejig this.

laurengarcia: Yes, inheriting the one Site model would be preferred, but the other PR is still appreciated and will be merged.

shared/js/utils.js
```diff
@@ -36,6 +36,12 @@ require.scopes.utils = ( () => {
         }
     }
 
+    function capitalizeFirstLetter (string) {
+        if (typeof string !== 'string') { return '' }
+
+        return string[0].toUpperCase() + string.substring(1)
+    }
+
```
laurengarcia: There's a way better way to do this -- am using npm `titleize` in the `site` template. Please use that.

shared/scss/_vars.scss
```diff
@@ -105,4 +105,3 @@ $icon__settings--moz: "moz-extension://__MSG_@@extension_id__/img/settings-gear@
 /* Close modal "x" icon */
 $icon__close--chrome: "chrome-extension://__MSG_@@extension_id__/img/close.svg";
 $icon__close--moz: "moz-extension://__MSG_@@extension_id__/img/close.svg";
```
laurengarcia: Can you fix your code editor so that it does not remove the blank line at the end of files? Thanks. Details are in the "Code editor setup" task in the "O-1 Meta" project.

shared/js/utils.js
```diff
@@ -36,6 +36,12 @@ require.scopes.utils = ( () => {
         }
     }
 
+    function capitalizeFirstLetter (string) {
+        if (typeof string !== 'string') { return '' }
+
+        return string[0].toUpperCase() + string.substring(1)
+    }
```
laurengarcia: In the Site template, I am using an npm package called `titleize` for this. Please use that instead on the front end. This utils file is for the background process.

shared/js/ui/views/privacy-practices.es6.js
laurengarcia: Inheriting the Site model is preferred to init'ing a new one here. The other PR is needed though so that will still get merged obvi.

shared/js/site.js
laurengarcia: Please just do it at the template level if you can. More comments below.

shared/js/utils.js
andrey-p: Titleize does every word though?

shared/scss/_vars.scss
andrey-p: Ah, I did this manually when I removed some variables from here. I'll readd.

shared/js/site.js
andrey-p: :+1: on moving this to the template - this is a presentation concern.

shared/js/utils.js
andrey-p: Maybe we can just switch to an all-in-one solution, e.g. https://www.npmjs.com/package/change-case

shared/js/site.js
```diff
@@ -52,10 +52,39 @@ class Score {
...
+                    message = constants.tosdrMessages.mixed
+                } else {
+                    if (tosdrData.score < 0) {
+                        message = constants.tosdrMessages.good  
+                    } else if (tosdrData.score === 0 && (matchGood.length || matchBad.length)) {
+                        message = constants.tosdrMessages.mixed
+                    } else if (tosdrData.score > 0 ) {
+                        message = constants.tosdrMessages.bad
+                    }
+                }
```
andrey-p: Hmm, this is getting me a `unknown privacy practices` for facebook.com even though there's two negative reasons shown. Let me investigate further.

andrey-p: Fixed it, turns out the constant values for `tosdrMessages.good` and `tosdrMessages.bad` were missing.

shared/js/ui/templates/shared/privacy-practices-overview.es6.js
```diff
@@ -0,0 +1,31 @@
...
+    status = tosdr.message.toLowerCase()
+  } else {
+    subtitle = `Unknown Privacy Practices`
+    status = `unknown`
+  }
+
+  return bel`<div>
+    <a href="#" class="privacy-practices__overview__close js-sliding-subview-close">
+      <span class="icon icon__arrow icon__arrow--left">
+      </span>
```
andrey-p: TODO this needs a bigger chevron icon, ask design for it next week.

shared/scss/popup.scss
```diff
@@ -748,6 +728,9 @@ body {
     }
 }
 
+@import "views/hero";
```
andrey-p: Since this isn't a standalone view, maybe it doesn't make sense to have it in a `views` directory. Any names you prefer? We use `objects` in the core repo, but that's somewhat nebulous. `subviews` is used elsewhere to mean something else. `components`?

andrey-p: Also, maybe we want to abstract away the template for the hero section into its own thing, but that's not trivial because of the `isCalculating` logic. Maybe we can do it in a separate PR (or file it away for a later date).

shared/scss/views/_hero.scss
```diff
@@ -0,0 +1,35 @@
...
+    padding-top: 6px;
+    @include uppercase_label();
+}
+
+// TODO this needs a bigger icon
+.hero__close {
+    position: absolute;
+    left: 10px;
+    top: calc(50% - 20px);
+}
```
andrey-p: This unifies any reused code between the hero section in the main popup view and in the subview.

I initially said I was going to do this in a separate PR, but then that would've been a lot of PRs flying about in a short period of time, so I decided against it.

andrey-p:
Two things irk me about this PR at the moment.

1. When the Site model updates, this view redraws only the hero section and the detail list. Whereas the main popup view and the new scorecard view (#8) rerender completely.

    What I've got here is more performant but I think I might make this view also rerender completely, so things are consistent.

2. I've abstracted away the styles for the hero section, but ideally it'd have its own template. I could do this here, but I'm worried this PR is becoming a mammoth. (I could add it if you don't mind adding yet another thing to review :) )

laurengarcia:
1. re: rerender - one of the main things here is to make sure that if the subview is scrollable, when you rerender its going to cause a "jump" back to the top. i havent pulled this one down yet so am not sure if its scrollable in any use-case, but just giving you a heads-up. 

I think the rerendering is more of an issue in the other PR you have going for Grade Details than here -- because I think the privacy practices just gets looked up once and then there's nothing to recalculate, whereas Grade Details can be constantly updating due to additional trackers found on page.

One thing to note, I've not pulled this down but: the Site model updates quite alot for reasons stated above, and I'm careful not to re-render the subview (or component within subview) unless there's a change to the model that pertains to that section. Please rerender as sparingly as possible.

2. I dont mind reviewing an abstraction for the hero section, in fact, it might be preferable because then you could just re-render that element in a targeted way if/when need be rather than the whole subview.

andrey-p:
Very good points, thank you.

I'll do the hero abstraction here so @MariagraziaAlastra can pull it into her PR. Then I'll update the other PR to be as sparing with re-renders as possible.

shared/js/ui/templates/shared/hero.es6.js
```diff
@@ -0,0 +1,18 @@
...
+    <div class="hero__icon hero__icon--${ops.status}">
+    </div>
+    <h1 class="hero__title">
+      ${ops.title}
+    </h1>
+    <h2 class="hero__subtitle">
+      ${ops.subtitle}
+    </h2>
+  </div>`
+}
```
andrey-p: cc @MariagraziaAlastra here's a hero template you can now use. You'll likely need to override some styles to set the individual styles, e.g. https://github.com/duckduckgo/duckduckgo-privacy-extension/blob/94e66350f6792b05e892d509ed098c47c77ee353/shared/scss/views/_privacy-practices.scss#L14-L28

shared/scss/_mixins.scss
```diff
@@ -36,7 +36,6 @@
 @mixin is_disableable_link () {
     body.is-disabled & {
         cursor: default;
-        opacity: 0.5;
```
andrey-p: Removed this, otherwise the text was opacified twice.

laurengarcia:
This looks great. Love the live trackers updates inside of the grade details view, thx for the extra effort there.

Found ONE minor bug - when I disabled "Site Privacy Protection" (aka whitelist a site), the copy in the grade details subview read "18 trackers _blocked_". The main site view reads that property and determines whether the extension "blocked" or "found" XX number of trackers (depending if whitelisted). Pushed a fix for this and one other minor alignment issue in the interest of time.

Really excellent stuff! Thank-you.

andrey-p:
Awesome! Excellent find with the blocked/found thing, thanks for jumping in and fixing it.

PR #4, opened 2017-12-21
Port to Chrome: Synchronous https lookups

laurengarcia:
**Reviewer:**
@jdorweiler @russellholt 

## Description:
This PR adds an additional step to onBeforeRequest handler and the https module for Chrome, which does not obey asynchronous request resolvers. To verify this works, run the unit tests in both Chrome and Firefox, and in Chrome you'll want to double-verify that the upgrade was obeyed by 
looking at network tab. See attached screenshot for desired behavior.

Unit tests added for both cases (synchronous and asynchronous).

chrome-https-synchronous

## Automated tests:
- [X] Unit tests
- [ ] Integration tests

###### Reviewer Checklist:

###### PR Author Checklist:

### DIFF
--- a/shared/js/background.js
+++ b/shared/js/background.js
@@ -232,7 +232,6 @@ chrome.webRequest.onBeforeRequest.addListener(
         // Fetch upgrade rule from db (synchronous -- Chrome only)
         if (utils.isChromeBrowser()) {
             console.log('IS CHROME BROWSER')
-            // TODO
 
         // Fetch upgrade rule from db (asynchronous -- Firefox/Mozilla)
         } else {
--- a/shared/js/background.js
+++ b/shared/js/background.js
@@ -227,16 +227,23 @@ chrome.webRequest.onBeforeRequest.addListener(
         // since we aren't sure if we're in Chrome or Firefox at this point
         if (!https.isReady) return
 
+        // Is this request from the tab's main frame?
         const isMainFrame = requestData.type === 'main_frame' ? true : false
 
         // Fetch upgrade rule from db (synchronous -- Chrome only)
         if (utils.isChromeBrowser()) {
-            console.log('IS CHROME BROWSER')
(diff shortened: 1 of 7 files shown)


shared/js/utils.js
```diff
@@ -36,6 +36,12 @@ require.scopes.utils = ( () => {
         }
     }
 
+    function isChromeBrowser () {
```
jdorweiler: We're doing this in background.js too. Do you want to generalize that and make it a util function? https://github.com/duckduckgo/duckduckgo-privacy-extension/blob/master/shared/js/background.js#L30

shared/js/background.js
```diff
@@ -224,11 +223,28 @@ chrome.webRequest.onBeforeRequest.addListener(
...
-            const isMainFrame = requestData.type === 'main_frame' ? true : false
+        // Make sure https module is ready, check synchronous .isReady property
+        // since we aren't sure if we're in Chrome or Firefox at this point
+        if (!https.isReady) return
+
+        // Is this request from the tab's main frame?
+        const isMainFrame = requestData.type === 'main_frame' ? true : false
+
+        // Fetch upgrade rule from db (synchronous -- Chrome only)
+        if (utils.isChromeBrowser()) {
```
jdorweiler: This doesn't change. What about setting it as a background variable to avoid calling this so often?

shared/js/utils.js
laurengarcia: Yeah, I thought about that but rejected it and opted for utils.isChromeBrowser because it is synchronous. The other one you point to is async, and the boot up/init for this is tricky. Want sync for this, thus new fn in utils.

shared/js/background.js
laurengarcia: Sounds reasonable.

laurengarcia:
Merging this so its easier for Brian to deal with on develop.

PR #8, opened 2017-12-29
Grade scorecard

andrey-p:
**Reviewer:** @laurengarcia 

**CC:** @russellholt 
**Depends on:** https://github.com/duckduckgo/duckduckgo-privacy-extension/pull/3

## Description:

Adds a grade scorecard, accessible by clicking on a button next to the hero icon on the main view. Branched off of #3, I've set the PR to merge back into #3's branch so I can see the diff but we can change that depending on the order you prefer to merge these in.

screenshot from 2017-12-29 14-39-37

## Automated tests:
- [ ] Unit tests
- [ ] Integration tests

###### Reviewer Checklist:

###### PR Author Checklist:


shared/js/ui/models/site.es6.js
```diff
@@ -44,6 +45,7 @@ Site.prototype = window.$.extend({},
                 this.domain = backgroundTabObj.site.domain
                 this.fetchSiteRating()
                 this.tosdr = backgroundTabObj.site.score.tosdr
+                this.isaMajorTrackingNetwork = backgroundTabObj.site.score.isaMajorTrackingNetwork
```
andrey-p: Adding this so I can pass it to the grade view.

shared/js/ui/templates/grade-scorecard.es6.js
```diff
@@ -0,0 +1,65 @@
...
+    ${detailItems}
+  </ul>`
+}
+
+function renderItem (modifier, item, highlight) {
+  return bel`<li class="status-list__item status-list__item--${modifier}
+      bold ${highlight ? 'is-highlighted' : ''}">
+    ${item}
+  </li>`
+}
```
andrey-p: There's now some common markup between this and the `privacy-practices-details`. I should probably consolidate those in separate shared template.

shared/scss/popup.scss
```diff
@@ -237,6 +237,12 @@ body {
             height: 64px;
         }
 
+        .site-info__rating__open {
+            position: absolute;
+            right: 10px;
+            top: calc(50% - 20px);
+        }
```
andrey-p: Pending a better icon from design.

shared/js/ui/views/grade-scorecard.es6.js
```diff
@@ -0,0 +1,90 @@
...
+}
+
+GradeScorecard.prototype = window.$.extend({},
+  Parent.prototype,
+  {
+    _render: function (ops) {
+      // doing this so the template can get access to the reasons
+      // TODO put them on a model? allow them to be passed via the arguments
+      // to _render like we do in core?
+      this.reasons = this._getReasons()
```
andrey-p: Looking at this again, it's probably best to move all this to a model. I'll do that next week.

shared/js/ui/models/grade-scorecard.es6.js
```diff
@@ -0,0 +1,75 @@
...
+
+    _getReasons: function () {
+      const site = this.site
+      let reasons = []
+
+      // grab all the data from the site to create
+      // a list of reasons behind the grade
+
+      // encryption status
+      let httpsStatusText = site.httpsStatusText
```
laurengarcia: The way I've been working with `let` vs. `const` is: if its a variable that gets updated after being declared then its a `let`. Otherwise, it should be a `const`. In this case, `httpsStatusText` appears to be a const. Please review the rest of the `let`s in this function likewise.

shared/js/ui/models/grade-scorecard.es6.js
```diff
@@ -0,0 +1,75 @@
...
+  this.reasons = this._getReasons()
+}
+
+GradeScorecard.prototype = window.$.extend({},
+  Parent.prototype,
+  {
+    modelName: 'gradeScorecard',
+
+    _getReasons: function () {
+      const site = this.site
```
laurengarcia: I don't get why you would rewire `site` up as a `const` here instead of just checking `this.site` throughout the function. If we bulk up the models with cruft like this they will become difficult to refactor later on.

shared/js/ui/models/grade-scorecard.es6.js
```diff
@@ -0,0 +1,75 @@
...
+      if (isPartOfMajorTrackingNetwork) {
+        reasons.push({
+          modifier: 'bad',
+          msg: `Site Is Part of a Major Tracker Network`
+        })
+      }
+
+      // privacy practices from tosdr
+      let privacyMessage = site.tosdr && site.tosdr.message
+      if (privacyMessage && privacyMessage !== 'Unknown') {
```
laurengarcia: Just a niggle, but please compare the `.toLowerCase()` versions of things coming from background process.

shared/js/ui/views/grade-scorecard.es6.js
```diff
@@ -0,0 +1,18 @@
+const Parent = require('./sliding-subview.es6.js')
+
+function GradeScorecard (ops) {
+  this.model = ops.model
+  this.template = ops.template
+
+  Parent.call(this, ops)
+
+  this.setupClose()
+}
```
laurengarcia: I think you'll want to make sure you subscribe to site model changes and re-render this view/template dynamically as things like tracker numbers change/trickle in.

shared/scss/views/_status-list.scss
```diff
@@ -0,0 +1,70 @@
...
+    .status-list__item--#{$rating} {
+        &:before {
+            @include cross_platform_bg_image("status--#{$rating}.svg");
+        }
+
+        &.is-highlighted {
+            color: $color;
+        }
+    }
+}
```
laurengarcia: Nice, thanks.

shared/js/ui/templates/grade-scorecard.es6.js
```diff
@@ -0,0 +1,65 @@
...
+      )}
+    </div>
+    ${getReasons(this.model.reasons)}
+    ${getGrades(this.model.site.siteRating)}
+  </div>
+  </section>`
+}
+
+function getReasons (reasons) {
+  if (!reasons.length) { return }
```
laurengarcia: You don't need curly braces here.

laurengarcia: Also, please check `!reasons && !reasons.length`

shared/js/ui/templates/grade-scorecard.es6.js
```diff
@@ -0,0 +1,65 @@
...
+
+  return bel`<ul class="status-list status-list--right padded border--bottom--inner">
+    ${reasons.map(item => { return renderItem(item.modifier, item.msg) })}
```
laurengarcia: es6 syntax would be:
`${reasons.map(item => renderItem(item.modifier, item.msg) )}`

The curly braces and `return` are implied in this form.

shared/js/ui/templates/grade-scorecard.es6.js
```diff
@@ -0,0 +1,65 @@
...
+  </ul>`
+}
+
+function getGrades (rating) {
+  const before = rating.before
+  const after = rating.after
+
+  if (!before || !after) {
+    return
+  }
```
laurengarcia: Please consolidate to one line without curly braces. Be terse.

shared/js/ui/models/grade-scorecard.es6.js
```diff
@@ -0,0 +1,75 @@
+const Parent = window.DDG.base.Model
+
+function GradeScorecard (attrs) {
+  Parent.call(this, attrs)
+
+  this.reasons = this._getReasons()
```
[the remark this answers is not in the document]
andrey-p: That's a very good point. Let me update.

shared/js/ui/models/grade-scorecard.es6.js
andrey-p: I've generally done this to reduce the number of times I have to type `this.`, since it's used multiple times in the function, plus it allows us to shave a few bytes off in minifiable code. Whether it makes refactoring more difficult or not largely depends on what kind of refactoring you need to do, e.g. if you change how `site` is accessed you may only need to edit one line.

I'm not strongly attached to the way I did it. I'll update the code to use `this.site`.

shared/js/ui/models/grade-scorecard.es6.js
andrey-p: Thank you for pointing this out - I'm still internalizing how to use `let` vs `const`!

shared/js/ui/models/grade-scorecard.es6.js
andrey-p: Looking at this again - what would you say to comparing it to the value in `constants` instead?

shared/js/ui/templates/grade-scorecard.es6.js
andrey-p: I've got into the habit of using curly braces even for tiny guards. Less worry if the guard ever becomes multiline or if I need to add errors/logs/debuger lines inside the guard.

Not too strongly attached to having them. I'll remove them here and the other place you pointed out.

shared/js/ui/models/grade-scorecard.es6.js
```diff
@@ -0,0 +1,75 @@
...
+        let modifier = 'bad'
+
+        if (httpsStatusText === 'Secure') {
+          connectionMsg = 'Encrypted'
+          modifier = 'good'
+        }
+
+        reasons.push({
+          modifier,
+          msg: `${connectionMsg} Connection`
```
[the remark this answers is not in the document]
andrey-p: I definitely see where you're coming with this, but doing this will require the templates to be able to differentiate between different kinds of reasons. This means spreading the idea of what a "reason" is between multiple entities, which probably makes this model unnecessary.

Wouldn't you say it's better to keep this model as the sole source of truth for what the scorecard reasons are, and keep the template as dumb and logic-less as possible?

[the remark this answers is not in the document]
andrey-p: (I'm not strongly attached to this BTW, if you're still not convinced I'll do it your way.)

shared/js/ui/models/grade-scorecard.es6.js
andrey-p: Changed to use the value in `constants`.

shared/js/ui/models/grade-scorecard.es6.js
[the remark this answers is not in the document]
andrey-p: That said, by the same token I should move the grades to this model as well. Let's see what we decide regarding this model generally, then move them or not.

andrey-p:
Addressed all your comments apart from the model thing (see above). I've also merged all the duplicate `status-list` markup into its own template.

andrey-p:
All right, this is now at a place where I'm really happy with it.

I applied the new abstracted hero templates to both the grade scorecard and the main popup header. I left the styles/templates that were previously shared between the main popup header and the grade details subview untouched. We'll need to remove them when we remove the grade details subview.

I also made sure the relevant parts of the grade scorecard subview only rerender when necessary.

Re: the model thing we discussed above, I think it comes down to fat models + skinny templates vs skinny models + fat templates. In the core codebase, we've generally gone for fat models as that's easier to unit test.

shared/js/ui/models/grade-scorecard.es6.js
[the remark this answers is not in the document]
laurengarcia: Sounds like a template mixin could be the single source of truth for message mappings then? Saying this because at some point if/when we internationalize, i want all of that logic strictly in the templates and not spread out all over the place in models and views.

shared/js/ui/views/grade-scorecard.es6.js
```diff
@@ -0,0 +1,73 @@
...
+    _rerenderReasons: function () {
+      this.$reasons.replaceWith(reasonsTemplate(
+        this.model.getReasons()
+      ))
+    },
+
+    _onSiteChange: function (e) {
+      if (e.change.attribute === 'siteRating') {
+        this._rerenderHero()
+        this._rerenderGrades()
```
laurengarcia: Two thumbs up here.

laurengarcia:
Left a message way above (scroll up) re: single source of truth for message mappings.

shared/js/ui/models/grade-scorecard.es6.js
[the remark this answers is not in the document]
andrey-p: OK, I buy that.

I may end up shifting this entire thing in the template then. The vast majority of this model is mapping values from the site model to a message.

andrey-p:
All right, I'm merging this into `andrey/privacy-practices`. Thanks for the input!

PR #12, opened 2018-01-04
Make https puts into indexedDB async

bsstoner:
**Reviewer:**
@laurengarcia 

**CC:** 
@nilnilnil 

## Description:
Making the puts async seems to have solve the issues with Chrome freezing and corrupting the db. Loading the large 200k list took a while and hogged CPU...we may want to throttle the puts even more so they can just run in the background over time?

It seems to all be working and upgrading the sites for me? (though I'm not sure how to fully test).

FYI -- I branched off #4 (Chrome https sync changes), not master.

## Automated tests:
- [ ] Unit tests
- [ ] Integration tests

###### Reviewer Checklist:

###### PR Author Checklist:

### DIFF
--- a/shared/js/indexed-db.js
+++ b/shared/js/indexed-db.js
@@ -36,7 +36,7 @@ class IndexedDBClient {
         this.db = null
 
         this.serverUpdateUrls = {
-            https: 'http://brian.duckduckgo.com/contentblocking.js?l=https-200k'
+            https: 'http://duckduckgo.com/contentblocking.js?l=https'
             // ...add more here
         }
         this.serverUpdateFails = 0
@@ -246,7 +246,7 @@ const fetchServerUpdate = {
 
                         putNextRecord.call(this).then(() => {
                             counter++
-                            console.log(`IndexedDBClient: ${counter} records added, ${records.length} left to add`)
+                            //console.log(`IndexedDBClient: ${counter} records added, ${records.length} left to add`)
                             putRecords.call(this)
                         }, () => {
                             // on error, just skip and go to the next record
--- a/shared/js/indexed-db.js
+++ b/shared/js/indexed-db.js
@@ -36,7 +36,7 @@ class IndexedDBClient {
         this.db = null
 
         this.serverUpdateUrls = {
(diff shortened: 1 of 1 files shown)


laurengarcia:
PS -- Thanks for branching off #4, just merged that to develop. Sorry for lag there.

bsstoner:
Added a 20ms setTimeout throttle, but I don't think it's good enough...

On my machine it used about ~10% of CPU, and at that rate would take ~66min to load all 200k records. I'm concerned about letting it use that much CPU for so long right after install. I could increase it, but at 50ms it uses ~5% of CPU for almost 3 hours.

I could batch it, and have it use a lot of CPU in short bursts, but I'm not sure that's any better.

laurengarcia:
Batching should significantly help and 5-10% CPU is pretty paltry.

bsstoner:
What do you mean by batching?

What's the plan for when the db gets half loaded and the user closes their browser?

nilnilnil:
From where I am sitting, this just isn’t scaling right. Either we need to understand why such a supposedly powerful thing is struggling, or we should move on to another approach. Yesterday we spoke about throwing work away, but this potentially looks like it may never work well for our purposes.

As far as investigating why, it seems like we already did some investigation on async va. non-async puts. Do we really want to put this more at risk for the sake of using what was already here?

nilnilnil:
Maybe the best thing to do is hop back on a call?

laurengarcia:
5-10% of CPU is pretty weak and this doesn't even attempt batching. Was thinking something like 100 batches of 2000 PUTs punctuated by 100ms would get us there in 5 seconds. To give you a sense of scale: before i optimized the homepage SVGs they were running up to 100-110% of CPU and I optimized them back to max of 70% of CPU for first iteration of animation, which i think is on the order of a few seconds long. Try the batching recipe above.

laurengarcia:
If you're tired of working on this or aren't super interested I can pick this up later today. Should be pretty easy.

laurengarcia:
Just documenting: This depends on (the 200k dataset from server) https://dub.duckduckgo.com/duckduckgo/ddg/pull/11741/files

PR #16, opened 2018-01-09
Use chrome.storage.local for https list

bsstoner:
**Reviewer:** @laurengarcia 
@nilnilnil 

## Description:
I think we should consider removing all the complexity around indexedDB and do something way simpler for now.

What this does is:
1. Each time the browser loads, in the background it tries to get an updated list from the server, if successful it's saved to local storage as a single item.
2. Saves the etag for next time to avoid reloading the list when there's no change.
3. Removes the white list concept (all updates should just go through the list, right?)
4. Removes the async logic since the lookup is done sync in memory
5. Packages a small list of domains with the extension, and uses that as a fallback if nothing can be loaded from local storage/server (this one is maybe not necessary, we could initialize with an empty array until the server download completes)

This approach seems to work fine in Chrome/Firefox with the big 200k list. It reads from and writes to storage immediately with no noticeable overhead.

Maybe easier to jump on a call and discuss at some point today.

## Automated tests:
- [ ] Unit tests
- [ ] Integration tests

###### Reviewer Checklist:

###### PR Author Checklist:


bsstoner:
Updated with changes to make it work across FF/Chrome/Safari.

IMO we should go with this approach for now and revisit indexedDB in the future when we have an explicit need for it. There's too many unknowns and additional complexity associated with using it when we don't actually need it and still have a ton of other things to get done over the next few weeks.

laurengarcia:
Will pull this down and take a look. One thing that bothers me is that chrome.storage.local is limited to 5.2MB in size (altho its nice bc its sync) and one of the stated goals at the outset of this project is a fairly large dataset (eventually). I did manage to get the indexed db writes down into a reasonable range with batching/throttling in #18. Definitely some trade-offs to consider.

bsstoner:
Are you sure there's a 5.2MB limit? From the docs it seems like as long as you specify the `unlimitedStorage` permission you can store as much as you want: https://developer.chrome.com/extensions/storage#property-local

And in Safari it seems like the limit is 100MB:
https://developer.apple.com/library/content/documentation/Tools/Conceptual/SafariExtensionGuide/ExtensionSettings/ExtensionSettings.html

shared/data/constants.js
```diff
@@ -8,7 +8,8 @@ let constants = {
     "entityMap": "data/tracker_lists/entityMap.json",
     "easylists": ["privacy", "general"],
     "blocking": ["Advertising", "Analytics"],
-    "httpsWhitelist": "data/httpsWhitelist.json",
+    "httpsUpgradeList": "data/httpsUpgradeList.json",
+    "httpsUpgradeListUrl": "https://brian.duckduckgo.com/contentblocking.js?l=https2",
```
nilnilnil: This part isn't mergeable for now. Demarcating PR as WIP.

shared/data/constants.js
```diff
@@ -8,7 +8,8 @@ let constants = {
     "entityMap": "data/tracker_lists/entityMap.json",
     "easylists": ["privacy", "general"],
     "blocking": ["Advertising", "Analytics"],
-    "httpsWhitelist": "data/httpsWhitelist.json",
+    "httpsUpgradeList": "data/httpsUpgradeList.json",
```
nilnilnil: I'm not sure this is a good idea, because I'm concerned about a user getting this bundled list, having a busted redirect forever loop on a site they care about, and just having a terrible experience. If we think this is really not probable, and our preference is not breaking the web -- we should probably just try again at some point or improve that logic in the next release.

Additionally, I'm not sure that removing the whitelist is ok. @laurengarcia has more knowledge here. I don't think it's safe to say today that we don't need this. I think it's much safer to say that we should have the whitelist come from the server though.

shared/data/defaultSettings.js
```diff
@@ -19,5 +19,6 @@ let defaultSettings = {
     "trackersWhitelistTemporary-etag": null,
     "trackersWhitelist-etag": null,
     "generalEasylist-etag": null,
-    "privacyEasylist-etag": null
+    "privacyEasylist-etag": null,
+    "https-etag": null
```
nilnilnil: I really enjoy how easy this is, nice.

shared/js/https.js
```diff
@@ -1,228 +1,211 @@
...
+const load = require('load')
 const settings = require('settings')
 const utils = require('utils')
-let knownMixedContentList
 
-settings.ready().then(() => {
-    load.JSONfromLocalFile(constants.httpsWhitelist, (wl) => knownMixedContentList = wl)
-})
+// number of chunks to split up the data
+// when storing to Safari to avoid over quota issues:
```
nilnilnil: Technical doc link for future reference?

shared/js/background.js
```diff
@@ -230,33 +230,15 @@ chrome.webRequest.onBeforeRequest.addListener(
...
-                return {redirectUrl: url}
-            } else {
-              return
-            }
-
-        // Fetch upgrade rule from db (asynchronous -- Firefox/Mozilla)
+        // Fetch upgrade rule from https module:
+        const url = https.getUpgradedUrl(requestData.url, thisTab, isMainFrame)
+        if (url.toLowerCase() !== requestData.url.toLowerCase()) {
+            console.log('HTTPS: upgrade request url to ' + url)
```
nilnilnil: Do we want these in production?

shared/js/https.js
```diff
@@ -1,228 +1,211 @@
...
 
-    ready () {
-        return this._ready
-    }
-
-    pipeRequestUrl (reqUrl, tab, isMainFrame) {
+    getFromStorage (fn) {
+        return new Promise((resolve, reject) => {
+            // For Chrome/Firefox:
+            if (window.chrome) {
```
nilnilnil: Isn't there a util function for this? The same one used previously for setting `isSync`. Why not use that?

shared/js/https.js
```diff
@@ -1,228 +1,211 @@
...
+            this.getFromStorage().then((items) => {
+
+                // if items were found in local storage,
+                console.log("HTTPS: init() found existing list in storage with " + items.length + " items")
+                httpsUpgradeList = items
+                return resolve()
+
+            }, () => {
+
+                // if it fails to find anything in local storage,
```
nilnilnil: Again, are we sure we want to do this? We get no recourse if something is broken in there. We literally have to wait for a release. The whitelist was different because it was the other way, explicitly saying ignore this one.

shared/js/https.js
```diff
@@ -1,228 +1,211 @@
...
+        load.loadExtensionFile({
+            url: constants.httpsUpgradeListUrl,
+            source: 'external',
+            etag: etag
+        }, (data, res) => {
+            // this only gets called if the etag is different
+            // and it was able to get a new list
+            console.log("HTTPS: updateList() got updated list from server")
+
+            let newEtag = res.getResponseHeader('etag') || ''
```
nilnilnil: Does `load.loadExtensionFile` handle the eTag? Namely, where does the tag get checked / skipped everything? Is it a nop from loadExtensionFile if the tag doesn't pass muster?

shared/js/https.js
nilnilnil: I suppose that was for isChromeBrowser. In any case, I think it's worth using what we have or changing those.

shared/js/https.js
```diff
@@ -1,228 +1,211 @@
...
+                    }
 
-            // Only deal with http calls
-            const protocol = utils.getProtocol(reqUrl).toLowerCase()
-            if (!protocol.indexOf('http:') === 0) {
-                if (this.isSync) { return reqUrl } else { return resolve(reqUrl) }
-            }
+                    try {
+                        let parsedList = JSON.parse(results['https-upgrade-list'])
+                        resolve(parsedList)
```
nilnilnil: Can you explain why some of these have a `return` call and some don't? Obviously execution continues, but I'm not sure what happens after?

shared/js/https.js
```diff
@@ -1,228 +1,211 @@
...
+                    console.log("HTTPS: getFromStorage() error parsing JSON from localStorage", e)
+                    reject()
                 }
+            } else {
+                reject()
             }
+        })
+    }
+
+    saveToStorage (data) {
```
nilnilnil: I'm fine with parameterizing this, but I don't like saving a data object to a hardcoded key. Put another way, I can associate any data with a specific key here. It's not likely; but it's still not my favorite interface.

shared/js/https.js
```diff
@@ -1,228 +1,211 @@
...
-        const testHosts = [
-            '1337x.to',
-            'submit.pandora.com',
-            '*.api.roblox.com',
-            'thump.vice.com',
-            'yts.ag'
-        ]
+        // Check for upgrades
+        for (let i=0; i<hosts.length; i++) {
+            if (httpsUpgradeList.indexOf(hosts[i]) > -1) {
```
nilnilnil: Comment this check or make it more obvious with another construct please?

shared/js/https.js
```diff
@@ -1,228 +1,211 @@
...
+                return reqUrl.replace(/^(http|https):\/\//i, 'https://')
```
nilnilnil: So we only operate on the front of the URL here. I wish we had a better way if we had to make further modifications like subdomains and path fragments. I'll look but assume it was the same before.

shared/data/constants.js
bsstoner: Agree, I will remove the local upgrade list.

I don't understand what value the whitelist adds in it's current form. Can't we just remove the entries from the larger list on the server and have it sync without shipping an update?

I could see a case for a whitelist that does updates during browser session since changes to the full list only sync on new browser sessions? But that seems out of scope for this PR.

shared/data/constants.js
bsstoner: We should discuss how to handle the backend piece. This PR won't work with the existing endpoint unless we modify it to do so.

The existing endpoint uses a JSON structure that makes working with the data more difficult than it needs to be and unless there's any reason not to, I'd like to change that along with deploying the larger list.

My preference would be to get the latest list from Zach, merge it with this PR and deploy it ahead of deploying this extension change: https://dub.duckduckgo.com/duckduckgo/ddg/pull/11741

shared/js/background.js
bsstoner: Agree we probably should remove in production, but doesn't seem critical.

They're useful for debugging and there's a lot of logging in the extension now, so I was following convention. We could do something similar to what we do internally where we strip them out as part of a production build? @laurengarcia do you have an existing plan for how to remove log statements?

shared/js/https.js
bsstoner: agree, will remove.

shared/js/https.js
bsstoner: will add it

shared/js/https.js
bsstoner: yes, the comment a few lines above was trying to say that, will make it clearer

shared/js/https.js
bsstoner: no, this is different. It's doing feature detection, not browser detection. chrome.storage.local is supported by FF and Chrome (and anyone that implements that same extension spec).

bsstoner: I'll make the comment clearer here.

shared/data/constants.js
nilnilnil: I think you can consider them both problematic, sure. My point is that the whitelist exists and was explicitly made to resolve issues. If that list is for sure incorporated by @zachthompson 's work on the Backend, then you'd be 100% right. I'm not sure we have captured the same edge case information there yet. We can push it as part of the dataset investigations over there though.

shared/js/https.js
bsstoner: I was only using returns when I needed to return early. Should be unnecessary here because it's either successfully finishing the try or falling to the catch and there's nothing afterwards.

shared/js/https.js
bsstoner: I was trying to encapsulate that in Chrome/FF it's one entry, in Safari it has to be chunked to multiple key/val entries...

shared/data/constants.js
nilnilnil: I added an asana comment to that effect linking back here. Thanks.

shared/data/constants.js
nilnilnil: Yup. Just saying so it's not merged by anyone yet.

shared/js/https.js
bsstoner: I'm not sure what this function was used for, I was just trying to make it functionally equivalent to the way it was before. @laurengarcia was this just for your debugging? Is it used as part of some automated tests somewhere?

shared/js/https.js
nilnilnil: Alright, is it worth adding that feature detection to utils then and calling that more clearly? Or is this OK and that would be annoyingly redundant? I really don't know.

bsstoner: yea, it could be in a util, I'm not sure how Jason is handling it, he probably had to do something similar elsewhere.

shared/js/https.js
nilnilnil: I think we're talking across each other here, but maybe not. I'm saying `saveToStorage` should either be without parameter, and use 2 hardcoded / instance variables in the function, or have 2 key, value.

```JavaScript
// This jawn saves the in memory rules set to local storage.
saveToStorage()

//This jawn is a generic interface to save stuff to local storage
saveToStorage(localStorageKey, localStorageData)
```

Perhaps I'm being pedantic though.

shared/js/https.js
nilnilnil: Fair enough. Bigger fish right now.

shared/js/background.js
russellholt: These log statements in particular go to the background process, so it's not visible in the normal dev panel.

bsstoner:
Made some updates in response to @nilnilnil's comments:
- removed local upgrade list, now it just uses an empty array until the list can be downloaded from the server
- pointed the endpoint to the production list (the small one), so that we can merge this and handle the larger list/api change independently
- removed the isReady/async init logic, because it wasn't serving a purpose, it will just run with the empty array until it can load a list from somewhere
- cleaned up the safari/localStorage chunking logic and added more comments
- found an issue in Firefox where the settings weren't loading before I was trying to get the latest etag, so I added a setTimeout to wait until settings are ready before trying to get the etag and check the server.

Other comments could use input from @laurengarcia or others, not sure they're necessary to address right now:
- Is there a plan for removing console.log statements?
- Do we need that test function at the bottom? If so, does it work or should it be changed?
- There's probably a better way to handle chrome.storage.local vs. localStorage, a util function or an interface class, should we do something different here? Or wait until we see what the rest of the Safari port looks like?

bsstoner:
^^ Added one more commit, to recursively call the updateList() function that checks the etag and update the list from the server if it's changed every 30min. Not sure that's the right frequency, but it can easily be changed.

nilnilnil:
I'm good here with the caveat I think @laurengarcia is likely a better reviewer overall.

shared/js/background.js
laurengarcia: Yeah, we can wrap these on a logger fn that strips for production but they are very useful and not really visible to users unless they really jump thru some hoops.

laurengarcia:
OH almost forgot -- we should probably nuke delete the indexed db instances in production for now.

bsstoner:
Made most of the updates:
- Updated the tests and debug functions, unit tests pass
- Switched to settings.ready(), that's a nice interface, thanks
- Added a step to delete indexedDB in the onInstalled handler in background.js so that it should only run on update I think? Could use help validating that will work as expected.

One question on the chrome.alarm thing. How will it work in Safari? Won't we still need the setTimeout for that case? How is that being addressed in abp.js for Safari?

laurengarcia:
Docs update: looks like Chrome changed their permissions policies at some point during the year and they added `unlimitedStorage` to the list of apis that don't trigger additional warnings/perms dialogs:

from March 2018 - https://cmsdk.com/javascript/unlimitedstorage-not-supported-by-chrome-optional-permission-api-how-to-update-extension-with-the-permission.html

https://developer.chrome.com/extensions/permission_warnings#nowarning

As always when working with the extension -- if/when we get there, we should package it up and test it as a publishable bundle (i.e. .crx file) in all browsers to double check, because if there's one thing we've learned its that browser api docs are frequently wrong/out of date/and-or don't apply to other browsers.

laurengarcia:
re: updates 

- sounds like jason is doing chrome.alarms as setIntervals in Safari, just pinged him and asked. eventually it would be nice to polyfill these things so we have dependable interfaces, but i guess we can just go with that for now. yick.

- thx for fixing tests

- will validate the onInstalled handler works. that event is probably one of the more tricky use-cases to test for and its annoying that we don't have a "set" way of dealing with it properly yet. tbd.

shared/js/https.js
laurengarcia: +1 Would be good to consolidate feature + browser detection + polyfills in one place. Added a task in bug queue to track for later.

nilnilnil:
What needs to happen here to get this merged? Conflicts obviously, but what else?

bsstoner:
Resolved conflicts

bsstoner:
Can we merge this so that it can get merged into Jason's Safari port branch and then we can start implementing the https upgrades for Safari?

I think the only thing left was uncertainty about how to adequately test indexedDB deletion. I don't think there's risk it will break anything, worst case it doesn't delete the old db for everyone and we have to modify and try again.

laurengarcia:
Sorry I was concentrating on UI work and GA surrogates yesterday. I never got a chance to pull this down and validate that it was working before going to my bday dinner ;) ... but it doesn't seem to be upgrading any of the sites on our list. Have only tested Chrome so far but wanted to raise this before EOD bc its later in the day on the east coast.

laurengarcia:
How to validate that https upgrades are actually working:

A good way to test is to use one single tab in private browsing mode -- the browser can't access its own in-memory cache of the redirects to https in a fresh single tab in private browsing mode.
Either that or you have to restart the browser to clear that cache.

The two things i look for (in chrome) that indicate our upgrading is working:
1 - output message in out background process that says "'HTTPS: upgrade request url to.... "
2 - double check(!!!) outgoing request in the network tab, the very first call to the main_frame will be an "internal redirect" in chrome
*NOTE that firefox it doesnt consider this an "internal redirect" in the network tab, so you have to just look at the address of the outgoing request (should be "https://")

Fakeouts:
Also note that you CANNOT just trust the console output of the background tab(!), esp in chrome. You must validate the background process output with the network tab.

You'll also want to distinguish it from the site's own auto-upgrades(!) which can fool you
Right now "adobe.com" is not auto-upgrading so its a good one to test, its in our upgrade list but i can def. type in "adobe.com" to address bar and the first request goes out unimpeded _without_ https and returns 200.

PR #18, opened 2018-01-10
Batch https 200k into indexed db

laurengarcia:
**Reviewer:**
@bsstoner @nilnilnil 

## Description:
This batch-writes the larger 200k https ruleset into indexed db by re-using the same transaction for each batch and throttling them. Also switches the PUTs to ADDs.

Perf: Writing the 200k records to disk is clocking in at ~6.5 mins on Chrome here for my setup. CPU never goes above ~52% . These numbers are fairly easy to tweak if you think that's too much CPU. [EDIT: see comment below re: Dexie.js and onsuccess() handlers etc]

(Making the large xhr call ramps CPU up at first, then as rules start writing it settles down into 45-52% CPU range for duration of batch writes). We could easily slow it down by tweaking one variable and use even less CPU (and more time). 

Or we could use #16 (chrome.storage.local) for https ruleset instead. If long-term we want more than 5MB of data overall (5MB is the api limit for chrome.storage.local), then #16 is probably throwaway work. [EDIT: looks like chrome at least updated their permission warnings policy sometime during 2017 and we can (maybe) just specify `unlimitedStorage` in the manifest w/o triggering a special dialog.

from March 2017 - https://cmsdk.com/javascript/unlimitedstorage-not-supported-by-chrome-optional-permission-api-how-to-update-extension-with-the-permission.html

https://developer.chrome.com/extensions/permission_warnings#nowarning

TODOs:
If we decide that this is the way to go then there's some TODOs left in the PR to flip the initialization for Chrome so it reads the rules into memory before starting the writes to disk in the background (i tested it and it takes 17s to read the 200k out of indexed db on subsequent startups). We also need to make sure we delete the old db every time we get a new etag. Eventually it would be nice to not have to do that and be more efficient about ingesting updates/new rules -- later.

## Automated tests:
- [ ] Unit tests
- [ ] Integration tests

###### Reviewer Checklist:

###### PR Author Checklist:

### DIFF
--- a/shared/js/indexed-db.js
+++ b/shared/js/indexed-db.js
@@ -1,3 +1,9 @@
+// TODO: delete OLD top500 entries in production!
+// TODO: flip chrome init sequence so it reads xhr data first to memory
+//       then starts the write here
+// TODO: optimize https.syncRuleCache so its structure is more simple:
+//       this.syncRuleCache = { 'hackernews.com': true }
+
 (function () {
 const utils = require('utils')
 
@@ -94,9 +100,9 @@ class IndexedDBClient {
                 }
                 cursor.continue()
             } else {
+                console.timeEnd('logRecords')
                 if (!optionalLogObject) {
                     console.log(`IndexedDBClient: logAllRecords() No more entries for objectStore: ${objectStore}`)
-                    console.timeEnd('logRecords')
                 }
             }
         }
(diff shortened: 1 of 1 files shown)


laurengarcia:
Apparently an additional optimization can be made here: much of the CPU suck is apparently from the DB .add() requests generating onsuccess() and onerror() functions. Last night I was looking at an IndexedDB library wrapper called Dexie.js which has bulk methods that somehow do things like .add() without generating those fns and that significantly optimizes writes. 

Seems like this is for later. 

#16 -- using chrome.storage.local -- seems sufficient for our "80%" dataset if its ~3.5MB, esp. bc it is a synchronous api.

This indexed db solution would probably be great for lazy-loading the long-tail dataset and the supporting the opportunistic strategy we discussed as a possible follow-up project later.

PR #19, opened 2018-01-10
Fix GA tracker issues

dharb:
**Reviewer:** @jdorweiler 

## Description:

This PR sets up the extension to detect outbound requests to GA and return data redirects to noop functions in their place a la https://raw.githubusercontent.com/uBlockOrigin/uAssets/master/filters/resources.txt

As of now we're only addressing a handful of necessary cases from that list, but I've structured this so that we may add more without needing to change any code here.

## Steps to test this PR:
1. npm run dev-firefox (or chrome)
2. reload extension
3. in the console, check that surrogateList exists and is properly structured data
4. nuke whitelist `whitelists.trackersWhitelist.isLoaded = false`
5. visit broken site (I've been using raspberrypi.org, follow steps to buy)
6. request to GA should be intercepted, and a redirect to a base64 data url should be returned.
7. most importantly, the tracker is blocked and the page still functions.

## Automated tests:
- [ ] Unit tests
- [ ] Integration tests

###### Reviewer Checklist:

###### PR Author Checklist:

### DIFF
--- a/shared/data/constants.js
+++ b/shared/data/constants.js
@@ -14,6 +14,7 @@ let constants = {
     "trackersWhitelist": "https://duckduckgo.com/contentblocking/trackers-whitelist.txt",
     "generalEasylist": "https://duckduckgo.com/contentblocking.js?l=easylist",
     "privacyEasylist": "https://duckduckgo.com/contentblocking.js?l=easyprivacy",
+    "trackersSurrogateList": "https://dave.duckduckgo.com/contentblocking.js?l=surrogates",
     /**
      * Major tracking networks data:
      * percent of the top 1 million sites a tracking network has been seen on.
--- a/shared/data/constants.js
+++ b/shared/data/constants.js
@@ -14,7 +14,7 @@ let constants = {
     "trackersWhitelist": "https://duckduckgo.com/contentblocking/trackers-whitelist.txt",
     "generalEasylist": "https://duckduckgo.com/contentblocking.js?l=easylist",
     "privacyEasylist": "https://duckduckgo.com/contentblocking.js?l=easyprivacy",
-    "trackersSurrogateList": "https://dave.duckduckgo.com/contentblocking.js?l=surrogates",
+    "trackersSurrogateList": "https://duckduckgo.com/contentblocking.js?l=surrogates",
     /**
      * Major tracking networks data:
      * percent of the top 1 million sites a tracking network has been seen on.
--- a/shared/js/abp-preprocessed.es6.js
+++ b/shared/js/abp-preprocessed.es6.js
@@ -171,7 +171,7 @@ function fetchSurrogateCode () {
             // take identifier from first line
             let pattern = firstLine.split(' ')[0]
             // create regular expression for it
-            let regex = new RegExp(pattern.replace(/\//g,'\\/').replace(/\./g,'\\.').concat('$'),'g')
+            let regex = new RegExp(pattern.replace(/\//g,'\\/').replace(/\./g,'\\.').concat('$'))
             // convert to base 64 string
             let b64surrogate = btoa(lines.join('\n'))
             surrogateList[pattern] = {regex: regex, snippet: b64surrogate}
--- a/shared/js/abp-preprocessed.es6.js
+++ b/shared/js/abp-preprocessed.es6.js
@@ -102,42 +102,7 @@ function updateLists () {
         trackersWhitelistTemporary = listData.trim().split('\n')
     })
 
-    let trackersSurrogateListEtag = settings.getSetting('trackersSurrogateList-etag') || ''
-    // reset etag to get a new list copy if we don't have data
-    if (!trackersSurrogateList || !trackersSurrogateListEtag) trackersSurrogateListEtag = ''
-    // load surrogates list
-    load.loadExtensionFile({url: constants.trackersSurrogateList, etag: trackersSurrogateListEtag, source: 'external'}, (listData, response) => {
-        const newTrackersSurrogateListEtag = response.getResponseHeader('etag') || ''
-        const clientChecksum = crypto.createHash('md5').update(listData).digest('base64')
-
-        settings.updateSetting('trackersSurrogateList-etag', newTrackersSurrogateListEtag);
-
-        //  check that etag hash matches hash of file received
-        if (!clientChecksum || clientChecksum.substring(0,6) !== newTrackersSurrogateListEtag) {
-            console.log("Checksum didn't match")
-        } else {
-            // split response using ** delimiter and remove first element
-            // since it's a comment
-            trackersSurrogateList = listData.trim().split('\n\n')
-
-            for (let surrogate of trackersSurrogateList) {
-                // remove comment lines
-                let lines = surrogate.split('\n').filter((line) => {
-                    return !(/^#.*/).test(line)
-                })
-                console.log('lines', lines)
-                // remove first line, store it
-                let firstLine = lines.shift()
-                // take identifier from first line
-                let pattern = firstLine.split(' ')[0]
-                // convert to base 64 string
-                let b64surrogate = btoa(lines.join('\n'))
-
-                surrogateList[pattern] = b64surrogate
-            }
-        }
-
-    })
+    fetchSurrogateCode()
 }
 
 // Make sure the list updater runs on start up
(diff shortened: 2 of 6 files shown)


shared/js/abp-preprocessed.es6.js
```diff
@@ -140,3 +143,41 @@ function getVersionParam () {
...
+	            let firstLine = lines.shift()
+	            // take identifier from first line
+	            let pattern = firstLine.split(' ')[0]
+	            // create regular expression for it
+	            let regex = new RegExp(pattern.replace(/\//g,'\\/').replace(/\./g,'\\.').concat('$'),'g')
+
+	            // convert to base 64 string
+        	    let b64surrogate = btoa(lines.join('\n'))
+
+    	        surrogateList[pattern] = {regex: regex, snippet: b64surrogate}
```
jdorweiler: Maybe check that we're putting valid data in here. I'm just thinking of some weird situation where the file is messed up and the `pattern` is parsed out as `.*` or something bad like that.

shared/js/trackers.js
```diff
@@ -64,13 +66,15 @@ require.scopes.trackers = (function() {
             // tracker is related to the current site. If this is the case we consider it to be the
             // same as a first party requrest and return
             var trackerByParentCompany = checkTrackersWithParentCompany(blockSettings, urlSplit, currLocation)
+            var surrogateTracker = checkSurrogateList(urlToCheck)
```
jdorweiler: I'd only do this after we have decided to block the request. Look up the surrogate and add the b64 to the data returned from isTracker, then in background.js make the cancel vs redirect to surrogate decision. So `isTracker` might return something like:
```
{ 
    parentCompany: "Google", 
    url: "google-analytics.com", 
    type: "Analytics", 
    block: true 
    surrogate: XXXXXXX
}
```

shared/data/constants.js
```diff
@@ -14,6 +14,7 @@ let constants = {
     "trackersWhitelist": "https://duckduckgo.com/contentblocking/trackers-whitelist.txt",
     "generalEasylist": "https://duckduckgo.com/contentblocking.js?l=easylist",
     "privacyEasylist": "https://duckduckgo.com/contentblocking.js?l=easyprivacy",
+    "trackersSurrogateList": "https://dave.duckduckgo.com/contentblocking.js?l=surrogates",
```
nilnilnil: Demarcating as WIP until this is pointing at a production node.

shared/js/abp-preprocessed.es6.js
```diff
@@ -140,3 +143,40 @@ function getVersionParam () {
...
+	if (!trackersSurrogateList || !trackersSurrogateListEtag) trackersSurrogateListEtag = ''
+	// load surrogates list
+	load.loadExtensionFile({url: constants.trackersSurrogateList, etag: trackersSurrogateListEtag, source: 'external'}, (listData, response) => {
+	    const newTrackersSurrogateListEtag = response.getResponseHeader('etag') || ''
+	    const clientChecksum = crypto.createHash('md5').update(listData).digest('base64')
+
+	    settings.updateSetting('trackersSurrogateList-etag', newTrackersSurrogateListEtag);
+
+	    //  check that etag hash matches hash of file received
+	    if (!clientChecksum || clientChecksum.substring(0,6) !== newTrackersSurrogateListEtag) {
```
nilnilnil: Why is there this extra checksum step?

shared/js/abp-preprocessed.es6.js
```diff
@@ -140,3 +143,40 @@ function getVersionParam () {
 
     return versionParam
 }
+
+function fetchSurrogateCode () {
+    let trackersSurrogateListEtag = settings.getSetting('trackersSurrogateList-etag') || ''
+	// reset etag to get a new list copy if we don't have data
+	if (!trackersSurrogateList || !trackersSurrogateListEtag) trackersSurrogateListEtag = ''
+	// load surrogates list
+	load.loadExtensionFile({url: constants.trackersSurrogateList, etag: trackersSurrogateListEtag, source: 'external'}, (listData, response) => {
```
nilnilnil: Style comment: generally this would look like:
```javascript
load.loadExtensionFile({
  url: constants.trackersSurrogateList,
  etag: trackersSurrogateListEtag,
  source: 'external'
}, (listData, response) => {
......
```

shared/js/abp-preprocessed.es6.js
```diff
@@ -140,3 +143,40 @@ function getVersionParam () {
...
+	            })
+	            // remove first line, store it
+	            let firstLine = lines.shift()
+	            // take identifier from first line
+	            let pattern = firstLine.split(' ')[0]
+	            // create regular expression for it
+	            let regex = new RegExp(pattern.replace(/\//g,'\\/').replace(/\./g,'\\.').concat('$'),'g')
+
+	            // convert to base 64 string
+                let b64surrogate = btoa(lines.join('\n'))
```
nilnilnil: spacing

shared/js/trackers.js
```diff
@@ -233,6 +237,17 @@ require.scopes.trackers = (function() {
         return match
     }
 
+    function checkSurrogateList(url) {
+        const b64dataheader = 'data:application/javascript;base64,'
+
+        for (item of Object.keys(surrogateList)) {
+            if (surrogateList[item].regex.test(url)) {
+                return {redirectUrl: b64dataheader + surrogateList[item].snippet}
```
nilnilnil: So this is all being done locally instead of redirecting to another endpoint it seems? Sorry if I missed it, but why the deviation?

shared/js/abp-preprocessed.es6.js
dharb: We're checking that the contents of the file the extension receives match the contents of the file sent from our servers. Doing this to somewhat mitigate a potential MitM attack.

shared/js/abp-preprocessed.es6.js
dharb: works for me

shared/js/trackers.js
russellholt: @jdorweiler  are you sure? That requires those to be in the list and certain to block. I see these as separate. In other words, this implementation can handle many other cases that won't be in the block list. This should be super fast lookup.

shared/js/abp-preprocessed.es6.js
```diff
@@ -5,6 +5,7 @@
  * This will be browserifyed and turned into abp.js by running 'grunt'
  */
 abp = require('abp-filter-parser')
+const crypto = require('crypto')
```
laurengarcia: There is a native `window.crypto.subtle` api that obviates the need for this. Also keep in mind in the future that if you add an npm package you need to specify it in package.json. In this case tho, `window.crypto.subtle` api should do it without the need.

shared/js/trackers.js
russellholt: I think this was always the plan.

redirecting with an endpoint would simplify the client but I think we're talking about redirecting URLs from 75% of the web, which has negative privacy implications. I might be wrong but also aren't there cross domain issues with HTTPS? This way is also more transparent since we send resources.txt unchanged.

shared/js/abp-preprocessed.es6.js
```diff
@@ -140,3 +147,41 @@ function getVersionParam () {
 
     return versionParam
 }
+
+function fetchSurrogateCode () {
+    let trackersSurrogateListEtag = settings.getSetting('trackersSurrogateList-etag') || ''
```
laurengarcia: Looks like since this is happening on bootup and depends on `settings.getSetting()` which can be laggy especially on Firefox, you'll want to definitely invoke the call to this function with `settings.ready().then()` promise, otherwise the stored etag is never going to be read.

shared/js/trackers.js
nilnilnil: That's a good point for both CSP and CORS. I think we could always make a privacy implication out of something, but in this particular case, I think that's a witch hunt. All that said, the CORS and CSP constrains seem non trivial and per server configurable, which this could avoid.

I think this is, as programmed, way too brittle to be run on, as you say, 75% of the web. This stuff needs to be bullet proof IMO.

shared/js/abp-preprocessed.es6.js
```diff
@@ -140,3 +147,41 @@ function getVersionParam () {
...
+        etag: trackersSurrogateListEtag,
+        source: 'external'
+    }, (listData, response) => {
+        const newTrackersSurrogateListEtag = response.getResponseHeader('etag') || ''
+        const clientChecksum = crypto.createHash('md5').update(listData).digest('base64')
+
+        settings.updateSetting('trackersSurrogateList-etag', newTrackersSurrogateListEtag);
+        //  check that etag hash matches hash of file received
+        if (!clientChecksum || clientChecksum.substring(0,6) !== newTrackersSurrogateListEtag) {
+            console.log("Checksum didn't match")
```
laurengarcia: Dumb niggle but could you specify that the trackers surrogate checksum didnt match and use single quotes.

shared/js/trackers.js
laurengarcia: Am also wondering this^^. Perhaps its just the optics.

shared/js/trackers.js
laurengarcia: Yup, was about to ask a similar/relevant question: because it is possible to serve up GA from an endpoint that we don't recognize or even from local `file://` protocol, what's to stop someone from evading our surrogate by hosting the file(s) elsewhere?

https://support.google.com/analytics/answer/1032389?hl=en

jdorweiler: Good point. That sounds fine then. We're still counting them as trackers, right?

shared/js/abp-preprocessed.es6.js
dharb: @laurengarcia that was my first thought, but crypto.subtle doesn't support md5: https://developer.mozilla.org/en-US/docs/Web/API/SubtleCrypto/digest. Now, why we're using md5 server-side in the first place is a great question, but out of the scope of this PR.

Also, I was under the impression that the crypto package I used is included in node, so there's no need to specify anything new in package.json: https://nodejs.org/api/crypto.html#crypto_crypto ?

shared/js/abp-preprocessed.es6.js
russellholt: I think we should ditch the crypto checksum at this time. this is overcomplicating it at this point.

jdorweiler:
@dharb you can add unit tests for the surrogates here: https://github.com/duckduckgo/duckduckgo-privacy-extension/blob/master/test/tests/trackers.js

dharb:
Closing this PR as we're merging it with https://github.com/duckduckgo/duckduckgo-privacy-extension/pull/21

PR #28, opened 2018-01-12
Safari HTTPS upgrading

bsstoner:
**Reviewer:**
@jdorweiler @laurengarcia @nilnilnil @russellholt 

## Description:
This approach seems to work for upgrading main frame requests.
Diff is against Jason's jd/safari-build branch.

## Automated tests:
- [ ] Unit tests
- [ ] Integration tests

###### Reviewer Checklist:

###### PR Author Checklist:


browsers/duckduckgo.safariextension/js/background.js
```diff
@@ -106,4 +107,59 @@ var onBeforeRequest = function (requestData) {
...
+    // avoid redirect loops:
+    // TODO: this won't work in safari yet:
+    //if (thisTab.httpsRedirects[e.url] >= 7) {
+    //    console.log('HTTPS: cancel https upgrade. redirect limit exceeded for url: \n' + e.url)
+    //    e.preventDefault()
+    //    e.target.url = thisTab.downgradeHttpsUpgradeRequest(e)
+    //}
+
+    const url = e.url
+    const isMainFrame = true // always main frame in this handler
```
nilnilnil: this is safari related, so maybe say that more explicitly?

bsstoner: Probably not necessary, the whole file and directory is safari overrides

jdorweiler:
I'm seeing this message on startup `"HTTPS: init() failed to get existing list from storage"` and `httpsUpgradeList` is empty.

bsstoner:
That message is expected the first time. You should see it go on to load the list from the server: `HTTPS: updateList() got updated list from server`

I'm not sure about the `httpsUpgradeList`, what is that used for? If you do `https.getUpgradeList()` does it return the list?

jdorweiler:
I think the problem is here https://github.com/duckduckgo/duckduckgo-privacy-extension/blob/develop/shared/js/https.js#L53

You'll want to clear out the etag if you don't have https data. That will force it go grab the latest list.

bsstoner:
ohhh, maybe this is because you're upgrading from indexedDB? I think you're saying I need to clear the etag because you have the latest etag saved in settings but the data was in indexedDB not localStorage? I don't think I handled that upgrade case correctly when we migrated over. I will fix that in a separate PR.

bsstoner:
Merged the fix into this branch, @jdorweiler can you try again and see if it works now?

browsers/duckduckgo.safariextension/js/background.js
```diff
@@ -106,4 +107,59 @@ var onBeforeRequest = function (requestData) {
...
+ */
+var onBeforeNavigation = function (e) {
+    //console.log("beforeNavigation", e.url, e.target.url, e.eventPhase, e)
+
+    if (!e.url) return
+
+    const tabId = tabManager.getTabId(e)
+    let thisTab = tabManager.get({tabId: tabId})
+
+    if (!thisTab) {
```
nilnilnil: What case does this happen?

bsstoner: Most of them I think, `beforeNavigation` fires before the other `canLoad` event handler above where the tab was previously being created.

jdorweiler:
@bsstoner yeah it works now. Is it ready to go or still WIP?

bsstoner:
Cool, I think it's good to merge, we can test more on the Safari PR.

PR #29, opened 2018-01-12
Clear https-etag (forcing server load) when there's no local list

bsstoner:
**Reviewer:**
@jdorweiler @laurengarcia @nilnilnil @russellholt 

## Description:
See task for more detailed explanation: https://app.asana.com/0/72649045549333/523002315037057

## Automated tests:
- [ ] Unit tests
- [ ] Integration tests

###### Reviewer Checklist:

###### PR Author Checklist:

### DIFF
--- a/shared/js/https.js
+++ b/shared/js/https.js
@@ -33,18 +33,19 @@ class HTTPS {
         this.getFromStorage().then((items) => {
             console.log("HTTPS: init() found existing list in storage with " + items.length + " items")
 
-            // if there are already items in the list (e.g. the server update somehow
-            // finishes faster than the local one) don't overwrite what's there.
-            if (!httpsUpgradeList.length) {
-                httpsUpgradeList = items
-            }
-        },() => {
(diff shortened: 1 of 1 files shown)


laurengarcia:
Yes, this seems like it must have been what i was seeing last night.

laurengarcia:
This looks better. I think one additional improvement would be to add a `setTimeout` to line 39 - the init() case where it finds the (soon to be large) list in storage and loads it into memory. We're asking the extension to do alot of things immediately on startup and am concerned that adding an immediate (large) xhr call to the mix right after reading/loading a large object into memory is going to cause perf seizures on underpowered devices. Can we wait, say, 10 seconds in that case?

bsstoner:
Thanks for the suggestion, let's keep it in our back pocket if we hit performance issues with the large list. It seems unrelated to the purpose of this PR and I'd rather not add complexity to the code to solve for problems we may or may not have.

PR #30, opened 2018-01-12
Set timeout for rerendering the top blocked pills

MariagraziaAlastra:
**Reviewer:** @laurengarcia 

## Description:
Fix for Firefox flashing an empty state for the top blocked networks in the main popup.

## Steps to test this PR:
1. Build and reload extension
2. Visit heavy tracker sites, like alternet.org or msnbc.com
3. Open the popup and make sure the top blocked part doesn't jump or flash empty states
4. Reset global stats and repeat 

## Automated tests:
- [ ] Unit tests
- [ ] Integration tests

###### Reviewer Checklist:

###### PR Author Checklist:

### DIFF
--- a/shared/js/ui/views/top-blocked-truncated.es6.js
+++ b/shared/js/ui/views/top-blocked-truncated.es6.js
@@ -9,7 +9,7 @@ function TruncatedTopBlocked (ops) {
 
   this.model.getTopBlocked().then(() => {
     Parent.call(this, ops)
-    this.rerenderList()
+    this._setup()
   })
 
   this.bindEvents([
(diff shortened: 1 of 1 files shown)


shared/js/ui/views/top-blocked-truncated.es6.js
```diff
@@ -9,6 +9,7 @@ function TruncatedTopBlocked (ops) {
   Parent.call(this, ops)
 
   this.model.getTopBlocked().then(() => {
+    setTimeout(() => this.rerenderList(), 750)
```
MariagraziaAlastra: Note: I tried setting a timeout on `this._rerender()`, but it was breaking clicking on "all trackers" to go to the subview

laurengarcia:
Just pushed a slight change to this PR, I can see how there was a race condition. The data is retrieved async from the background via `this.model.getTopBlocked().then(() => {...` but `Parent.call()` triggers the first render and was being called before the data came back. If there is no data the template renders the the `${noData()}` function. Makes sense, and would only show up on really heavy sites.

PR #42, opened 2018-01-16
prevent https redirect loops in safari

bsstoner:
**Reviewer:**
@jdorweiler @laurengarcia 

## Description:
https://app.asana.com/0/72649045549333/524358648218923

## Steps to test this PR:
adobe.com --> should upgrade as usual
businessinsider.com --> shouldn't get stuck in redirect loop, should just fallback to using http

## Automated tests:
- [ ] Unit tests
- [ ] Integration tests

###### Reviewer Checklist:

###### PR Author Checklist:

### DIFF
--- a/browsers/duckduckgo.safariextension/js/background.js
+++ b/browsers/duckduckgo.safariextension/js/background.js
@@ -162,12 +162,24 @@ var onBeforeNavigation = function (e) {
     // skip upgrading broken sites:
     if (thisTab.site.isBroken) {
         console.log('HTTPS: temporarily skip upgrades for: ' + url)
+
+        // e.preventDefault() here prevents broken
+        // sites from loading via autocomplete. It comes
+        // with the downside that it resets the omnibar 
+        // back to what it's state was before they started typing when ,
+        // but this is probably better than autoloading a site?
+        e.preventDefault()
+
         return
     }
 
     // skip trying again if we've already tried upgrading this url
     if (thisTab.hasUpgradedUrlAlready(url)) {
         console.log('HTTPS: skipping upgrade to avoid redirect loops', url)
+
+        // same as the block above
+        e.preventDefault()
+
         return
     }
 
--- a/browsers/duckduckgo.safariextension/js/background.js
+++ b/browsers/duckduckgo.safariextension/js/background.js
@@ -134,11 +134,21 @@ var onBeforeRequest = function (requestData) {
 var onBeforeNavigation = function (e) {
     if (!e.url) return
 
+    //console.log(`onBeforeNavigation ${e.url} ${e.target.url}`)
+
     const url = e.url
     const isMainFrame = true // always main frame in this handler
     const tabId = tabManager.getTabId(e)
 
     let thisTab = tabId && tabManager.get({tabId: tabId})
+
+    // if a tab already exists, but the url is different,
+    // delete it and recreate for the new url
+    if (thisTab && thisTab.url !== url) {
+        tabManager.delete(tabId)
+        thisTab = null
+        console.log('onBeforeNavigation DELETED TAB because url did not match')
+    }
 
     if (!thisTab) {
         thisTab = tabManager.create(e)
@@ -162,24 +172,12 @@ var onBeforeNavigation = function (e) {
     // skip upgrading broken sites:
     if (thisTab.site.isBroken) {
         console.log('HTTPS: temporarily skip upgrades for: ' + url)
-
-        // e.preventDefault() here prevents broken
-        // sites from loading via autocomplete. It comes
-        // with the downside that it resets the omnibar 
-        // back to what it's state was before they started typing when ,
-        // but this is probably better than autoloading a site?
-        e.preventDefault()
-
         return
     }
 
     // skip trying again if we've already tried upgrading this url
     if (thisTab.hasUpgradedUrlAlready(url)) {
         console.log('HTTPS: skipping upgrade to avoid redirect loops', url)
-
-        // same as the block above
-        e.preventDefault()
-
         return
     }
 
--- a/browsers/duckduckgo.safariextension/js/background.js
+++ b/browsers/duckduckgo.safariextension/js/background.js
@@ -134,9 +134,9 @@ var onBeforeRequest = function (requestData) {
  * check whether we should upgrade to https
  */
 var onBeforeNavigation = function (e) {
-    if (!e.url || !e.target || e.target.url === 'about:blank' || e.url.match(/com.duckduckgo.safari/)) return
-
     //console.log(`onBeforeNavigation ${e.url} ${e.target.url}`)
+
+    if (!e.url || !e.target || e.target.url === 'about:blank' || e.url.match(/com.duckduckgo.safari/)) return
 
     const url = e.url
     const isMainFrame = true // always main frame in this handler
@@ -192,7 +192,6 @@ var onBeforeNavigation = function (e) {
 
         e.preventDefault()
         e.target.url = upgradedUrl
-        return
     }
 }
 
--- a/browsers/duckduckgo.safariextension/js/background.js
+++ b/browsers/duckduckgo.safariextension/js/background.js
@@ -139,22 +139,31 @@ var onBeforeNavigation = function (e) {
         return
     }
 
-    // avoid redirect loops:
-    // TODO: this won't work in safari yet:
-    //if (thisTab.httpsRedirects[e.url] >= 7) {
-    //    console.log('HTTPS: cancel https upgrade. redirect limit exceeded for url: \n' + e.url)
-    //    e.preventDefault()
-    //    e.target.url = thisTab.downgradeHttpsUpgradeRequest(e)
-    //}
-
     const url = e.url
     const isMainFrame = true // always main frame in this handler
     const upgradedUrl = https.getUpgradedUrl(e.url, thisTab, isMainFrame)
+
+    // avoid redirect loops:
+    if (thisTab.httpsRedirects[url]) {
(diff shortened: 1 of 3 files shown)


browsers/duckduckgo.safariextension/js/background.js
```diff
@@ -139,23 +139,32 @@ var onBeforeNavigation = function (e) {
         return
     }
 
-    // avoid redirect loops:
-    // TODO: this won't work in safari yet:
-    //if (thisTab.httpsRedirects[e.url] >= 7) {
```
bsstoner: @laurengarcia why were we waiting for 7 times before bailing? Why can't we bail after one failed attempt?

browsers/duckduckgo.safariextension/js/tabManager.js
```diff
@@ -33,13 +33,13 @@ class TabManager {
...
      */
     create(tabData) {
-        console.log(`CREATE TAB: ${tabData.url}`)
         // when it's created from a message event, use tabData.message.currentURL, but
         // when it's created from a beforeNavigation event, message object won't exist, so use tabData.url:
         let url = tabData.message ? tabData.message.currentURL : tabData.url
         let createTabData = {url: url, id: tabManager.getTabId(tabData)}
         createTabData.target = tabData.target
         
+        console.log(`CREATE TAB: ${url}`)
```
bsstoner: This was unrelated, but the log output was showing undefined since tabData.url isn't always defined

browsers/duckduckgo.safariextension/js/tabManager.js
```diff
@@ -101,6 +101,13 @@ class TabManager {
 var tabManager = new TabManager();
 
 var closeHandler = function (e) {
+    // only want to delete the tab
+    // when it's being closed. This handler will get
+    // called when the url is changing in the tab
+    // and we don't want to delete it here in those cases
+    // or it breaks things like https upgrades:
+    if (e.type !== 'close') return
```
bsstoner: I *think* this is what we want, right? What's currently happening on the safari branch is it's deleting and recreating the tab every time the url changes (including upgrades from http to https), which was making it impossible to track state and prevent redirect loops.

nilnilnil: What are the other types and when do they happen? Is there a doc online about it you used to find this?

browsers/duckduckgo.safariextension/js/tabManager.js
nilnilnil: Makes sense.

browsers/duckduckgo.safariextension/js/background.js
```diff
@@ -139,23 +139,32 @@ var onBeforeNavigation = function (e) {
...
     const url = e.url
     const isMainFrame = true // always main frame in this handler
     const upgradedUrl = https.getUpgradedUrl(e.url, thisTab, isMainFrame)
 
+    // avoid redirect loops:
+    if (thisTab.httpsRedirects[url]) {
+        const downgradedUrl = thisTab.downgradeHttpsUpgradeRequest(e)
+        console.log('HTTPS: cancel https upgrade. redirect limit exceeded for: ' + url)
+
+        if (url.toLowerCase() !== downgradedUrl.toLowerCase()) {
```
nilnilnil: This is a new case it seems. What's it for?

browsers/duckduckgo.safariextension/js/tabManager.js
bsstoner: I was working backwards from debugging. Actually, it might be this line where we're invoking it ourselves: https://github.com/duckduckgo/duckduckgo-privacy-extension/pull/42/files#diff-051311c865c94f354e947f669b5569f7R121

bsstoner: @jdorweiler do you remember why you were invoking `closeHandler()` on `unloadTab`? Why wasn't the `close` event sufficient for that?

browsers/duckduckgo.safariextension/js/background.js
bsstoner: preventing infinite loops, doing `e.preventDefault()` and `e.target.url = `, even when the url is the same seems to dispatch the `beforeNavigate` even which sends it back through the handler.

browsers/duckduckgo.safariextension/js/tabManager.js
jdorweiler: I'd double check but from what I remember, close fires on tab closes and unload fires whenever you navigate to a new page. In either case we want to clean up the old tab data. Maybe instead of unload we could do this in beforeNavigate?

Your PR seems to work but is also breaking the popup when you navigate to new pages. Is it from calling preventDefault in the beforeNavigate handler?

jdorweiler: Never mind, I wasn't looking at this actual line in the diff. Yes, we need to clean up the tabs with unload events or find some other way around that.

browsers/duckduckgo.safariextension/js/background.js
laurengarcia: Sorry for the captain obvious response but: because many of the ads on the web redirect several times before reaching final destination. The EFF's version of this waits 10 times. I cut it down to 7.

laurengarcia: In practical terms: if you lower this any more, you're going to break alot of non-tracker third party content on the web.

bsstoner:
Okay, tried a different approach above that seems to work. Could use help poking holes in it.

jdorweiler:
Can you merge the latest safari branch and test it again? I'm getting stuck in a redirect loop still.

jdorweiler:
If you're testing businessinsider for the redirect loop, make sure to remove it from the whitelist. It's in `trackersWhitelistTemporary`. This is an easy way to remove it.
```
trackersWhitelistTemporary.shift()
```

bsstoner:
okay, on it now

browsers/duckduckgo.safariextension/js/background.js
```diff
@@ -160,28 +161,23 @@ var onBeforeNavigation = function (e) {
...
-    const upgradedUrl = https.getUpgradedUrl(e.url, thisTab, isMainFrame)
+    const upgradedUrl = https.getUpgradedUrl(url, thisTab, isMainFrame)
 
     if (url.toLowerCase() !== upgradedUrl.toLowerCase()) {
         console.log('HTTPS: upgrade request url to ' + upgradedUrl)
         thisTab.upgradedHttps = true
-        thisTab.addHttpsUpgradeRequest(upgradedUrl)
+        thisTab.addHttpsUpgradeRequest(upgradedUrl, url)
 
-        e.preventDefault()
```
bsstoner: removing `e.preventDefault()` seems to have fixed autocomplete from redirecting on upgrade to a page that's pre-fetching and everything seems to still work fine without it as far as I can tell..

bsstoner:
@jdorweiler I think this is ready for another look.

I did what we discussed, saving redirect url's to the safari tab, and clearing them off after page load. That seems to work to prevent redirect loops.

I also removed the `e.preventDefault()` before setting the upgrade url, which seems to stop the autocomplete/redirect issue from happening (e.g. when you would start typing adobe.com and it would immediately upgrade and load the page)

There's one remaining issue that I've been unable to find a solution for, with redirect loops, the autocomplete issue still exists. What's happening is when it upgrades from http to https and then gets automatically redirected by the site back to http somehow the page is going from being prefetched in the background to being loaded on the screen. I can call a `e.preventDefault()` to prevent it from happening, but what happens is that prevents you from loading that url at all, even if you type it in explicitly and press enter. Wondering if we should just ignore this case for now and hope that the data is good enough that we don't have any of these in our new list? At least it won't get stuck in a loop and crash the browser now.

bsstoner:
Actually, I just found a bug, the popup is empty sometimes on opening sites from a new tab...
screen shot 2018-01-17 at 6 30 45 pm

bsstoner:
Sorry, there's still some other weirdness happening, I'm going to have to look at it again tomorrow.

bsstoner:
@jdorweiler I fixed the issues and think this could be merged now. The only things it's addressing now are https upgrade redirect protection and making the blocked sites check work. Since we changed the https list, the old test sites don't work anymore, but I was doing this in the console to add them back:
```
https.canUpgradeHost = function (host) { return host.match(/businessinsider\.com|adobe\.com/) || (this.getUpgradeList().indexOf(host) > -1) ? true : false }
```

jdorweiler:
I'm still getting auto redirected when https upgrades a prefetched page.

jdorweiler:
It's only for urls that autocomplete with http since we skip the upgrade if it's already https. For me it just happens with businessinsider. Maybe it's not a big issue if it's that isolated.

bsstoner:
yea, I think the auto redirect issue is going to need to be a product call, either people think it's not that bad or we find patterns for when it's most annoying and narrow the scope of upgrades or we ship the feature disabled behind a setting or something.

jdorweiler:
Seems fine to me as long as we're really sure the domains in the upgrade list are good.

bsstoner:
Thoughts on merging this and re-releasing safari to ddh6 so people can test the full https upgrade functionality and we can use that feedback to determine whether or not to ship it?

nilnilnil:
One possibility is to also make this have it's own HTTPs url that we can change the backend to no results without re-releaseing.

jdorweiler:
Putting this up for testing sounds good, I'll do that today.

jdorweiler:
@bsstoner added a parameter so we can serve a separate https list for safari if we need to. I'm going to merge this and put it up for testing.

PR #50, opened 2018-01-16
Fix page load bug

bsstoner:
**Reviewer:**
@jdorweiler 

## Steps to test this PR:
Open new tab
start typing 'goo'
icon should no longer update to D
navigate to duckduckgo.com (without going to google first)
icon should update to A for duckduckgo

## Automated tests:
- [ ] Unit tests
- [ ] Integration tests

###### Reviewer Checklist:

###### PR Author Checklist:

### DIFF
--- a/browsers/duckduckgo.safariextension/js/content-script.js
+++ b/browsers/duckduckgo.safariextension/js/content-script.js
@@ -51,14 +51,5 @@ function getLocation () {
 }
 
 window.onbeforeunload = ((e) => unload(e))
-window.onfocus = ((e) => sendLoadEvent(e))
 
 document.addEventListener('beforeload', onBeforeLoad, true);
-document.addEventListener("DOMContentLoaded", sendLoadEvent, true)
-        
-        
-function sendLoadEvent (event) {
-    if (window === window.top) {
-        safari.self.tab.dispatchMessage('tabLoaded', {mainFrameURL: mainFrameURL})
-    }
-}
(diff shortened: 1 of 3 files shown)


browsers/duckduckgo.safariextension/js/content-script.js
```diff
@@ -51,14 +51,5 @@ function getLocation () {
 }
 
 window.onbeforeunload = ((e) => unload(e))
-window.onfocus = ((e) => sendLoadEvent(e))
```
bsstoner: what was the onfocus handler for? That may be broken now if you needed that code to run on focus.

jdorweiler:
It's almost there. The badge doesn't update with the preloaded site grade but I still see wrong site data in the popup.

bsstoner:
Pushed the changes we discussed, it solves the first bug, but creates some weird scenarios. e.g.
1) navigate to duckduckgo.com
2) start typing google.com, but don't hit enter and then click out/press escape
3) open the popup (still has info for google showing)

It seems like there are 2 tabs (the visible and the one preloading), maybe we need to model it that way? Leave the visible tab alone, but route the preloading events to a preloadingTab that we keep separate until it becomes visible, then we upgrade it to the real tab?

FF/Chrome didn't have this issue?

laurengarcia:
This branch appears to fix the bug I was seeing earlier where private browsing tabs don't show the correct https state for sites like google.com and microsoft.com. I'll test it some more to make sure it hasn't broken other things.

jdorweiler:
Looks good, thanks for fixing this.

PR #56, opened 2018-01-17
Show post-install page on install or update event

MariagraziaAlastra:
**Reviewer:** @laurengarcia 

## Description:
Show the new post-install page (same as in #47) on install.
The page is now live, which makes this easier to test :smile: 

## Steps to test this PR:
1. Build and reload extension
2. Page should be loaded in a new tab
3. Try reloading again from the interface - nothing should happen this time
4. Uninstall and install again - page should show
5. Close and reopen the browser - nothing should happen

## Automated tests:
- [ ] Unit tests
- [ ] Integration tests

###### Reviewer Checklist:

###### PR Author Checklist:

### DIFF
--- a/shared/js/background.js
+++ b/shared/js/background.js
@@ -72,13 +72,18 @@ function Background() {
     // only run the following section on install and on update
     if (details.reason.match(/install|update/)) {
         ATB.onInstalled();
+    }
 
-        if (!localStorage['hasSeenPostInstall']) {
-          localStorage['hasSeenPostInstall'] = true;
-          chrome.tabs.create({
-            url: 'https://www.duckduckgo.com/app?post=1'
-          })
-        }
+    // only show post install page on install
+    if (details.reason.match(/install/)) {
+        settings.ready().then( () => {
+          if (!settings.getSetting('hasSeenPostInstall')) {
+            settings.updateSetting('hasSeenPostInstall', true)
+            chrome.tabs.update({
+              url: 'https://www.duckduckgo.com/app?post=1'
+            })
+          }
+        })
     }
 
     // blow away old indexeddbs that might be there:
--- a/shared/js/background.js
+++ b/shared/js/background.js
@@ -69,9 +69,16 @@ function Background() {
   });
 
   chrome.runtime.onInstalled.addListener(function(details) {
-    // only run the following section on install
+    // only run the following section on install and on update
     if (details.reason.match(/install|update/)) {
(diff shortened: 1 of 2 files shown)


shared/js/background.js
```diff
@@ -69,9 +69,16 @@ function Background() {
   });
 
   chrome.runtime.onInstalled.addListener(function(details) {
-    // only run the following section on install
+    // only run the following section on install and on update
```
russellholt: doesn't this mean it will show on upgrade? we do not want to do that at this time.

shared/js/background.js
```diff
@@ -69,9 +69,16 @@ function Background() {
...
     if (details.reason.match(/install|update/)) {
         ATB.onInstalled();
+
+        if (!localStorage['hasSeenPostInstall']) {
+          localStorage['hasSeenPostInstall'] = true;
+          chrome.tabs.create({
+            url: 'https://www.duckduckgo.com/app?post=1'
+          })
+        }
```
laurengarcia: In addition to Russell's comment above, i also think `hasSeenPostInstall` should go into settings, and should also have a default value in shared/data/defaultSettings.js.

shared/js/background.js
laurengarcia: I think you'll want to do 
`if (details.reason.match(/install/) { 
  settings.ready().then( () => { 
     settings.updateSetting( 'hasSeenPostInstall', true)
     chrome.tabs.update( ... )
}`

Double-check that everything above is correct but that's a rough idea of what should be happening here.

laurengarcia: (Sorry for formatting above)

MariagraziaAlastra:
@russellholt Ok now it only runs on first install
@laurengarcia Addressed your comments as well, I agree having hasSeenPostInstall as a setting makes more sense 👍

shared/js/background.js
```diff
@@ -69,11 +69,23 @@ function Background() {
...
   chrome.runtime.onInstalled.addListener(function(details) {
-    // only run the following section on install
+    // only run the following section on install and on update
     if (details.reason.match(/install|update/)) {
         ATB.onInstalled();
     }
 
+    // only show post install page on install
+    if (details.reason.match(/install/)) {
+        settings.ready().then( () => {
```
jagtalon: Wow arrow syntax you're living in the future! 🌵

laurengarcia:
Thanks!

PR #57, opened 2018-01-17
load tosdr in manifest

jdorweiler:
**Reviewer:**

## Description:

When you restart the browser with saved tabs there can be a race condition between loading tosdr data and tab loading. The result is that sometimes a saved tab will have the wrong grade assigned. 

To fix this I'm just loading tosdr from the manifest. This way we can guarantee it's loaded in time.

## Automated tests:
- [ ] Unit tests
- [ ] Integration tests

###### Reviewer Checklist:

###### PR Author Checklist:

### DIFF
--- a/shared/js/site.js
+++ b/shared/js/site.js
@@ -9,18 +9,10 @@
 var load = require('load')
 var settings = require('settings')
 
-let tosdr
 let tosdrRegexList
-let tosdrListLoaded
 let trackersWhitelistTemporary
 
-settings.ready().then(() => {
-    load.JSONfromLocalFile(constants.tosdr,(data) => {
-        tosdr = data
-        tosdrRegexList = Object.keys(tosdr).map(x => new RegExp(x))
-        tosdrListLoaded = true
-    })
-})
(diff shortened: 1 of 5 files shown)


shared/js/site.js
```diff
@@ -43,9 +35,6 @@ class Score {
     getTosdr() {
         let result = {}
 
-        // return if the list hasn't been built yet
-        if (!tosdrListLoaded) return result
-
         tosdrRegexList.some(tosdrSite => {
```
laurengarcia: Overall this makes sense and should be faster without waiting for `settings.ready()` but I would feel better about this if `tosdrRegexList` had a default value that won't error out if it hits `tosdrRegexList.some()` and its still undefined.

laurengarcia:
Not sure this will fix the bug that Gabriel was seeing tho -- ?. See task for details.

shared/js/site.js
laurengarcia: (I realize its sync but just in case someone comes along and moves things around...)

PR #66, opened 2018-01-18
fix re-renders in popup disabled state

laurengarcia:
**Reviewer:**

## Description:
The popup recently started re-rendering in the disabled state way too much due to `updateTabData` events from the background process. Is/was causing a memory leak for new tabs (disabled state). I've fixed the re-rendering issue, but think I might be only addressing the symptom here, but the memory leak seems to have gone away with this fix. 

Will keep an eye on this.

## Automated tests:
- [ ] Unit tests
- [ ] Integration tests

###### Reviewer Checklist:

###### PR Author Checklist:

### DIFF
--- a/shared/js/ui/views/site.es6.js
+++ b/shared/js/ui/views/site.es6.js
@@ -71,10 +71,12 @@ Site.prototype = window.$.extend({},
     rerender: function () {
       // console.log('[site view] rerender()')
       if (this.model && this.model.disabled) {
-        console.log('.addClass is-disabled')
-        this.$body.addClass('is-disabled')
-        this._rerender()
-        this._setup()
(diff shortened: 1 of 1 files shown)


shared/js/ui/views/site.es6.js
```diff
@@ -71,10 +71,12 @@ Site.prototype = window.$.extend({},
     rerender: function () {
       // console.log('[site view] rerender()')
       if (this.model && this.model.disabled) {
-        console.log('.addClass is-disabled')
-        this.$body.addClass('is-disabled')
-        this._rerender()
-        this._setup()
+        if (!this.$body.hasClass('is-disabled')) {
```
russellholt: that's better

MariagraziaAlastra:
Thanks! LGTM

PR #68, opened 2018-01-19
Andrey/trackers scrolling

andrey-p:

## Description:

This changes the tracker view to include the hero and explainer sections when scrolling, not just the tracker list. There was some weirdness with the container (see inline comments below) which is why the diff looks a bit crazier.

screenshot from 2018-01-19 15-48-41

## Steps to test this PR:

Sorry, this change can potentially all the subviews in a tiny way.

1. Check the tracker view scrolls OK (e.g. go to https://vice.com and open the trackers tab)
2. The privacy practices should theoretically be able to scroll as well. I fudged it by adding `for (let i = 0; i < 10; i++) { good.push(good[0]) }` around [this line](https://github.com/duckduckgo/duckduckgo-privacy-extension/blob/develop/shared/js/ui/templates/privacy-practices.es6.js#L58) and visiting duckduckgo.com.
3. All other views/subviews should look exactly the same.

## Automated tests:
- [ ] Unit tests
- [ ] Integration tests

###### Reviewer Checklist:

###### PR Author Checklist:

### DIFF
--- a/shared/scss/popup.scss
+++ b/shared/scss/popup.scss
@@ -239,7 +239,7 @@ body {
         }
 
         .site-info__rating-li {
-            padding-top: 0;
+            padding-top: 8px;
         }
 
         .site-info__rating {
--- a/shared/scss/popup.scss
+++ b/shared/scss/popup.scss
@@ -384,7 +384,11 @@ body {
     }
 
     .site-info--full-height {
-        bottom: 0;
+        position: relative;
+        margin-top: 0;
+        top: 0;
+        height: 100%;
+        overflow-y: auto;
     }
 
     // main popup view has a coloured hero section
--- a/shared/scss/popup.scss
+++ b/shared/scss/popup.scss
@@ -461,68 +461,6 @@ body {
                 font-size: 13px;
                 padding: 7.5px 20px 15px 20px;
             }
-
-            .site-info__trackers__company-list {
-                overflow-y: auto;
-                height: 230px;
-                padding: 0 20px 60px 20px;
-
-                    li {
-
-                        &.is-empty {
-                            padding: 15px 0 120px 0;
-                        }
-
-                        &:first-child {
-                            padding-top: 7.5px;
-                        }
-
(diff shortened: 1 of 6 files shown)


MariagraziaAlastra:
Great work @andrey-p! Only thing: maybe it'd be cool to have a max width for the tracker name?
Low priority though, doesn't have to be in this PR
screen shot 2018-01-19 at 1 59 41 pm

andrey-p:
That's a good call, might as well fix it here!

shared/scss/popup.scss
```diff
@@ -457,68 +461,6 @@ body {
...
-                        padding-top: 7.5px;
-                        font-size: 13px;
-                        color: $color--grey;
-                        position: relative;
-
-                        .url {
-                            display: block;
-                            width: 65%;
-                            text-overflow: ellipsis;
-                            overflow: hidden;
```
andrey-p: Max width was already here. It's just none of those styles were applying because this was nested into `.site-info--details`, which wasn't happening in the new view.

I just copied the bits over that made sense to keep, and deleted the rest since it wasn't used.

MariagraziaAlastra: @andrey-p I think the max-width and overflow logic needs a div to be effective:
screen shot 2018-01-19 at 2 27 59 pm

vs

screen shot 2018-01-19 at 2 27 40 pm

MariagraziaAlastra: .url is currently a span, second screenshot is the current one

andrey-p: Gah, sorry, I think I pushed without doublechecking. Lemme fix.

andrey-p:
@MariagraziaAlastra All right, fixed! I had to position the category absolutely, otherwise it was pushed down. I also added `white-space: nowrap` on the URL since it could wrap at a hyphen, and that looked weird.

MariagraziaAlastra:
Alright, LGTM! Thanks again @andrey-p!

PR #73, opened 2018-01-21
allow sites with missing url field to use first related url

dharb:
**Reviewer:** @jdorweiler 

**CC:** @russellholt 

## Description: 
Some sites in tosdr are missing the url field (see https://github.com/tosdr/tosdr.org/blob/master/services/kolabnow.json for example), yet have multiple items in the urls field. This allows us to take the first item from the urls field in place of the url field in this case.

## Steps to test this PR:

1. pull this down and run `grunt execute:tosdr --browser=chrome --type=dev`, then check shared/data/tosdr.js and search for 'kolab'. there should now be an entry.

## Automated tests:
- [ ] Unit tests
- [ ] Integration tests

###### Reviewer Checklist:

###### PR Author Checklist:

### DIFF
--- a/scripts/tosdr.js
+++ b/scripts/tosdr.js
@@ -62,6 +62,12 @@ function getSitePoints (sites) {
         let servicesUrl = `${githubRepo}/services/${site}.json`
         request.get(servicesUrl, (err, res, body) => {
             let data = JSON.parse(body)
+            // some sites lack the 'url' field, but have
+            // multiple items in the 'urls' field.
+            if (!data.url && relatedUrls) {
+                data.url = relatedUrls.shift()
+            }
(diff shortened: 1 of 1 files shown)


dharb:
@jdorweiler I tested locally and this seems to work, however I noticed that we're showing multiple of the same privacy practices:
virtualbox_2018-01-20_19-15-44

Looking at the tosdr data for kolabnow, they actually do have multiple entries for 'no third-party access without a warrant': https://raw.githubusercontent.com/tosdr/tosdr.org/master/api/1/service/kolabnow.json

Do you think we should add a unique check to the match list?

jdorweiler:
👍

PR #79, opened 2018-01-23
Don't open post install page if extension is installed from there

MariagraziaAlastra:
**Reviewer:** @andrey-p 

**CC:** @jagtalon 

## Description:
As @jagtalon pointed out, no need to open /app if the user installs the extension from there

## Steps to test this PR: 
1. Remove current extension
2. Build and load from this branch
3. Post install page should open as usual
4. Now change `domain` in background.js to be the same as `postInstallUrl`
5. Repeat from step 1 - this time the post install page should NOT open
6. Feel free to `console.log` the domain and inside the if statement for further verification

## Automated tests:
- [ ] Unit tests
- [ ] Integration tests

###### Reviewer Checklist:

###### PR Author Checklist:

### DIFF
--- a/shared/js/background.js
+++ b/shared/js/background.js
@@ -77,10 +77,12 @@ function Background() {
     // only show post install page on install
     if (details.reason.match(/install/)) {
         settings.ready().then( () => {
-          if (!settings.getSetting('hasSeenPostInstall')) {
+          const domain = window.location ? window.location.href : ''
+          const postInstallUrl = 'https://www.duckduckgo.com/app'
+          if ((!settings.getSetting('hasSeenPostInstall')) && (!domain.match(postInstallUrl))) {
             settings.updateSetting('hasSeenPostInstall', true)
             chrome.tabs.create({
-              url: 'https://www.duckduckgo.com/app?post=1'
+              url: postInstallUrl + '?post=1'
             })
           }
         })
--- a/shared/js/background.js
+++ b/shared/js/background.js
@@ -74,7 +74,9 @@ function Background() {
         ATB.onInstalled();
     }
 
-    // only show post install page on install
+    // only show post install page on install if:
+    // - the user wasn't already looking at the app install page
+    // - the user hasn't seen the page before
     if (details.reason.match(/install/)) {
         settings.ready().then( () => {
             chrome.tabs.query({currentWindow: true, active: true}, function(tabs) { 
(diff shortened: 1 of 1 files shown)


shared/js/background.js
```diff
@@ -77,10 +77,12 @@ function Background() {
     // only show post install page on install
     if (details.reason.match(/install/)) {
         settings.ready().then( () => {
-          if (!settings.getSetting('hasSeenPostInstall')) {
+          const domain = window.location ? window.location.href : ''
+          const postInstallUrl = 'https://www.duckduckgo.com/app'
```
andrey-p: I'm not sure if the `www` will always be there. E.g. if I install the extension from https://duckduckgo.com/app.

Also, if you use `match` the string gets converted to a regex ([MDN link](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/match)). It might be a good idea to escape the dots?

MariagraziaAlastra:
@andrey-p Looser regexp added, thanks! 👍

andrey-p:
Wait, sorry, I should've picked this up when I looked at this earlier. I don't think you can get the URL of the currently open tab by just checking against `window`.

Have you managed to get this to work without fudging the value of `domain`?

MariagraziaAlastra:
@andrey-p yeah it seems to pick up the right URL, I logged it in the console

andrey-p:
What steps did you do?

I tried the following:

1. Add `console.log("current window location: " + (window.location && window.location.href))` just where your changes are
2. On Chrome, open `chrome://extensions` and switch on Developer mode
3. Uninstall the extension if it's there, then install it again via "Load unpacked extension"
4. Click on the errors link, scroll down a bit, the following is in the log:
    
    ```
    current window location: chrome-extension://ogigmfedpbpnnbcpgjloacccaibkaoip/_generated_background_page.html
    ```

    (I'd expect to see `chrome://extensions`)

MariagraziaAlastra:
@andrey-p Ready again!
Just for the record, we talked about this and turns out window.location in background.js gets the extension's background instead of the current tab URL, which I didn't realize.
Now using the Chrome API instead to get the actual tab location!

andrey-p:
This is now good :+1:

Hope you don't mind, I jumped in and added a comment. That code block was getting a bit intense.

PR #83, opened 2018-01-24
Harden HTTPS redirect loop detection

andrey-p:
**Reviewer:** @jdorweiler

**CC:** @bsstoner @laurengarcia 

## Description:

This adds a couple of extra methods to deal with HTTPS redirect loops.

Specifically it deals with two redirect loops that we found, one on http://store.steampowered.com/ (which only happened in FF) and one at http://raisingchildren.net.au/articles/safe_water_temperature.html (both Chrome and FF).

See descriptions inline for more info.

## Steps to test this PR:

Visit both http://raisingchildren.net.au/articles/safe_water_temperature.html and http://store.steampowered.com/ on both Chrome and FF, there should be no HTTPS redirect loops.

Note, we removed the Steam Store rule from our HTTPS ruleset. To fudge it and pretend it's still there, you can add something like `if (reqUrl === 'http://store.steampowered.com/') { return 'https://store.steampowered.com/' }` right around [this line](https://github.com/duckduckgo/duckduckgo-privacy-extension/blob/develop/shared/js/https.js#L178).

## Automated tests:
- [ ] Unit tests
- [ ] Integration tests

###### Reviewer Checklist:

###### PR Author Checklist:

### DIFF
--- a/shared/js/background.js
+++ b/shared/js/background.js
@@ -242,9 +242,15 @@ chrome.webRequest.onBeforeRequest.addListener(
             return
         }
 
+        if (thisTab.downgradedUrls[requestData.url]) {
+            console.log('already tried https upgrades for this URL and failed, skip: ' + requestData.url)
+            return
+        }
+
         // Avoid redirect loops
         if (thisTab.httpsRedirects[requestData.requestId] >= 7) {
             console.log('HTTPS: cancel https upgrade. redirect limit exceeded for url: \n' + requestData.url)
+
             return {redirectUrl: thisTab.downgradeHttpsUpgradeRequest(requestData)}
         }
 
--- a/shared/js/background.js
+++ b/shared/js/background.js
@@ -274,7 +274,6 @@ chrome.webRequest.onBeforeRequest.addListener(
                 thisTab.upgradedHttps = true
                 thisTab.lastInProgressUrl = requestData.url
             }
-            thisTab.addHttpsUpgradeRequest(url)
             return {redirectUrl: url}
         } else {
           return
--- a/shared/js/background.js
+++ b/shared/js/background.js
@@ -147,7 +147,7 @@ chrome.webRequest.onBeforeRequest.addListener(
                 let newTab = tabManager.create(requestData)
 
                 // persist the last URL the tab was trying to upgrade to HTTPS
-                newTab.lastUpgrade = thisTab && thisTab.lastUpgrade
+                newTab.lastHttpsUpgrade = thisTab && thisTab.lastHttpsUpgrade
                 thisTab = newTab
             }
 
@@ -246,7 +246,7 @@ chrome.webRequest.onBeforeRequest.addListener(
             return
         }
 
-        if (thisTab.downgradedUrls[requestData.url]) {
+        if (thisTab.failedUpgradeUrls[requestData.url]) {
             console.log('already tried https upgrades for this URL and failed, skip:\n' + requestData.url)
             return
         }
@@ -262,9 +262,9 @@ chrome.webRequest.onBeforeRequest.addListener(
         const isMainFrame = requestData.type === 'main_frame' ? true : false
 
         if (isMainFrame &&
-                thisTab.lastUpgrade &&
-                thisTab.lastUpgrade.url === requestData.url &&
-                Date.now() - thisTab.lastUpgrade.time < 1000) {
+                thisTab.lastHttpsUpgrade &&
+                thisTab.lastHttpsUpgrade.url === requestData.url &&
+                Date.now() - thisTab.lastHttpsUpgrade.time < 1000) {
 
             console.log('already tried upgrading this url on this tab a few moments ago ' +
                 'and it didn\'t complete successfully, abort:\n' +
@@ -279,7 +279,7 @@ chrome.webRequest.onBeforeRequest.addListener(
             console.log('HTTPS: upgrade request url to ' + url)
             if (isMainFrame) {
                 thisTab.upgradedHttps = true
-                thisTab.lastUpgrade = {
+                thisTab.lastHttpsUpgrade = {
                     url: requestData.url,
                     time: Date.now()
                 }
--- a/shared/js/background.js
+++ b/shared/js/background.js
@@ -147,7 +147,7 @@ chrome.webRequest.onBeforeRequest.addListener(
                 let newTab = tabManager.create(requestData)
 
                 // persist the last URL the tab was trying to upgrade to HTTPS
-                newTab.lastUpgradeInProgressUrl = thisTab && thisTab.lastUpgradeInProgressUrl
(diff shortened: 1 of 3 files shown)


shared/js/tab.js
```diff
@@ -147,21 +145,6 @@ class Tab {
...
-    if (tab && header.statusCode < 400) {
-        tab.httpsRequests = tab.httpsRequests.filter((url) => {
-            // disregard diff between 'https://www.foo.com' and 'https://foo.com'
-            const _url = url.replace('https://www.', 'https://')
-            const _headerUrl = header.url.replace('https://www.', 'https://')
-            return _url !== _headerUrl
-        });
-    }
-
-}, {urls: ['<all_urls>']});
```
andrey-p: Removed this, it was unused code. Probably extra overhead for every single request.

shared/js/background.js
```diff
@@ -144,7 +144,11 @@ chrome.webRequest.onBeforeRequest.addListener(
         // don't have a tab instance for this tabId or this is a new requestId.
         if (requestData.type === "main_frame") {
             if (!thisTab || (thisTab.requestId !== requestData.requestId)) {
-                thisTab = tabManager.create(requestData);
+                let newTab = tabManager.create(requestData)
+
+                // persist the last URL the tab was trying to upgrade to HTTPS
+                newTab.lastHttpsUpgrade = thisTab && thisTab.lastHttpsUpgrade
```
andrey-p: This was necessary in the RaisingChildren case. The redirect wasn't done with a 30* HTTP status or something, it actually served a page which redirected you to the HTTP version:

```
$ curl https://raisingchildren.net.au/articles/safe_water_temperature.html
<html><head><title></title><!-- <script language="javascript">window.location.replace("http://raisingchildren.net.au/articles/safe_water_temperature.html");</script> --></head><body></body></html>
```

This meant the `requestId` for the tab was different, which meant `thisTab` was refreshed, and so we need to persist `lastHttpsUpgrade` between the two somehow.

shared/js/background.js
```diff
@@ -242,21 +246,44 @@ chrome.webRequest.onBeforeRequest.addListener(
             return
         }
 
+        if (thisTab.failedUpgradeUrls[requestData.url]) {
+            console.log('already tried https upgrades for this URL and failed, skip:\n' + requestData.url)
+            return
+        }
```
andrey-p: `failedUpgradeUrls` serves as a temporary whitelist that the tab builds up as it encounters failing URLs. Maybe we can share this between all tabs by storing it in `https.js`.

shared/js/background.js
```diff
@@ -242,21 +246,44 @@ chrome.webRequest.onBeforeRequest.addListener(
...
+
         // Avoid redirect loops
         if (thisTab.httpsRedirects[requestData.requestId] >= 7) {
             console.log('HTTPS: cancel https upgrade. redirect limit exceeded for url: \n' + requestData.url)
+
             return {redirectUrl: thisTab.downgradeHttpsUpgradeRequest(requestData)}
         }
```
andrey-p: The `main_frame` HTTPS upgrades are addressed by the logic below, so this now only handles redirect loops for page assets (e.g. images, etc).

While we could port the `main_frame` time-based logic to prevent repeated requests for these, it seems a bit overkill.

shared/js/background.js
```diff
@@ -242,21 +246,44 @@ chrome.webRequest.onBeforeRequest.addListener(
...
+                thisTab.lastHttpsUpgrade &&
+                thisTab.lastHttpsUpgrade.url === requestData.url &&
+                Date.now() - thisTab.lastHttpsUpgrade.time < 1000) {
+
+            console.log('already tried upgrading this url on this tab a few moments ago ' +
+                'and it didn\'t complete successfully, abort:\n' +
+                requestData.url)
+            thisTab.downgradeHttpsUpgradeRequest(requestData)
+            return
+        }
```
andrey-p: The time-based approach feels weird, but it's the only thing that worked for the RaisingChildren case.

One thing I tried was to cache the URL and clear it after successful page load like we do on the Safari branch. However, on Chrome the redirect page was still counted as a successful page load, so the URL was cleared prematurely.

bsstoner: yea, the timer doesn't seem ideal. Is there a "page successfully loaded and isn't going to be redirected anymore" event where we could clear `thisTab.lastHttpsUpgrade`?

bsstoner: What about clearing `lastHttpsUpgrade` here? https://github.com/duckduckgo/duckduckgo-privacy-extension/blob/master/shared/js/tabManager.js#L139

andrey-p: > Is there a "page successfully loaded and isn't going to be redirected anymore" event where we could clear `thisTab.lastHttpsUpgrade`

Not sure. For Chrome, the JS redirect page I [mentioned above](https://github.com/duckduckgo/duckduckgo-privacy-extension/pull/83#discussion_r163517171) *is* the successfully loaded page. We can't tell that the page isn't going to do any more redirects unless we:

- parse the content (this way madness lies)
- wait a bit (what we're doing now)

> What about clearing `lastHttpsUpgrade` here?

Can you elaborate? The headers will come before the redirect page content is parsed and acted upon, right?

bsstoner: I see, thanks, that makes more sense now. What about this `chrome.webNavigation` API? Does it do the redirect before onCompleted fires there?

bsstoner: https://developer.chrome.com/extensions/webNavigation

andrey-p: Nice find, but `onCompleted` seems to fire before the page loads is complete.

bsstoner: weird, really? It says in the docs it's after DOMCompleted.

andrey-p: Here's what I did. Testing on Chrome:

1. Add the following to [this line or thereabouts](https://github.com/duckduckgo/duckduckgo-privacy-extension/blob/eafef4fc7c47484f6be311a165d9863ef73ded34/shared/js/background.js#L70):
    
    ```
    chrome.webNavigation.onCompleted.addListener(function (details) {
       const tab = tabManager.get(details)
       console.log("navigation completed for: " + tab.url)
       delete tab.lastHttpsUpgrade
    })
    ```
2. Open http://raisingchildren.net.au/articles/safe_water_temperature.html - it'll enter a redirect loop because `tab.lastHttpsUpgrade` isn't there anymore
3. Open the debug log for the extension, here's what I see:

screenshot from 2018-01-24 17-46-56

That `tab.status` log line appears in the `tabs.onUpdated` event, [here](https://github.com/duckduckgo/duckduckgo-privacy-extension/blob/eafef4fc7c47484f6be311a165d9863ef73ded34/shared/js/tabManager.js#L112). I've tried removing `lastHttpsUpgrade` in that event, and it's still way too early.

jdorweiler:
Looks good to me

PR #121, opened 2018-02-06
Top trackers hero section

andrey-p:
**Reviewer:** @laurengarcia 

## Description:

Watch this space.

## Automated tests:
- [ ] Unit tests
- [ ] Integration tests

###### Reviewer Checklist:
- [ ] **Ensure the PR solves the problem**
- [ ] **Review every line of code**
- [ ] **Ensure the PR does no harm by testing the changes thoroughly**
- [ ] **Get help if you're uncomfortable with any of the above!**
- [ ] Determine if there are any quick wins that improve the implementation

###### PR Author Checklist:
- [ ] Get advice or leverage existing code
- [ ] Agree on technical approach with reviewer (if the changes are nuanced)
- [ ] Ensure that there is a testing strategy (and documented non-automated tests)
- [ ] Ensure there is a documented monitoring strategy (if necessary)
- [ ] Consider systems implications

PR #129, opened 2018-02-12
Tracker count updates

laurengarcia:
**Reviewer:** @laurengarcia @andrey-p 

## Description:
- Update the trackers count to show *all* the blocked tracker requests instead of counting each domain only once
- Consequently, change text from "X Tracker Networks Blocked/Found" to "X Trackers Blocked/Found"

## Steps to test this PR:
1. Build and reload extension
2. Navigate to sites with trackers and verify the count (console logging the trackersCount is also an easy option)
3. Optional: compare number of trackers blocked or found with other tracker blocker extensions - it should be mostly on par or better
 4. When 0 trackers are found, the text will say "0 Trackers Found"

## Automated tests:
- [ ] Unit tests
- [ ] Integration tests

###### Reviewer Checklist:

###### PR Author Checklist:

### DIFF
--- a/shared/js/ui/templates/shared/tracker-networks-text.es6.js
+++ b/shared/js/ui/templates/shared/tracker-networks-text.es6.js
@@ -7,7 +7,7 @@ module.exports = function (site, isMajorNetworksCount) {
   trackersText += (trackerNetworksCount === 1) ? ' Network ' : ' Networks '
   let finalText = trackerNetworksCount + trackersText + trackersBlockedOrFound(site, trackerNetworksCount)
 
-  const uniqueTrackersText = site.trackersCount === 1 ? ' Tracker In ' : ' Trackers In '
+  const uniqueTrackersText = site.trackersCount === 1 ? ' Tracker in ' : ' Trackers in '
   finalText = site.trackersCount + uniqueTrackersText + finalText
 
   return bel`${finalText}`
--- a/shared/js/ui/templates/shared/tracker-networks-text.es6.js
+++ b/shared/js/ui/templates/shared/tracker-networks-text.es6.js
@@ -1,16 +1,14 @@
 const bel = require('bel')
 
-module.exports = function (site, isMajorNetworksCount, includeUniqueTrackersCount) {
+module.exports = function (site, isMajorNetworksCount) {
   const trackerNetworksCount = isMajorNetworksCount ? site.majorTrackerNetworksCount : site.totalTrackerNetworksCount
 
   let trackersText = isMajorNetworksCount ? ' Major' : ''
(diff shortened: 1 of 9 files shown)


andrey-p:
Here's small tidbit that may be useful for this PR.

When I was looking into adding a Top Trackers hero section and making that trackers view scrollable, I ran into an issue where the internal part of the percentage bar was hardcoded with a pixel value rather than a percentage value. So every time you change the width of the outer bar, you need to change the hardcoded pixel value, which is painful.

With two very simple changes ([here](https://github.com/duckduckgo/duckduckgo-privacy-extension/pull/121/files#diff-9c2581b395500cfd1705c006f61977c5R10) and [here](https://github.com/duckduckgo/duckduckgo-privacy-extension/pull/121/files#diff-0f79666c93ee3a828a853a208da5e899R10)) we can save ourselves the effort of maintaining that hardcoded px value going forward.

shared/js/ui/templates/shared/tracker-networks-text.es6.js
```diff
@@ -1,16 +1,14 @@
...
-  let trackersText = isMajorNetworksCount ? ' Major Tracker' : ' Tracker'
+  let trackersText = isMajorNetworksCount ? ' Major' : ''
   trackersText += (trackerNetworksCount === 1) ? ' Network ' : ' Networks '
   let finalText = trackerNetworksCount + trackersText + trackersBlockedOrFound(site, trackerNetworksCount)
 
-  if (includeUniqueTrackersCount && trackerNetworksCount > 0) {
-    const uniqueTrackersText = site.trackersCount === 1 ? ' Unique Tracker In ' : ' Unique Trackers In '
-    finalText = site.trackersCount + uniqueTrackersText + finalText
-  }
+  const uniqueTrackersText = site.trackersCount === 1 ? ' Tracker In ' : ' Trackers In '
```
russellholt: lowercase 'in'

laurengarcia:
Overall looks good. Found a minor bug.

The counts appear to work as advertised and that is the meat of this PR. I added breakpoints into the `Tracker` class and watched things populate correctly. The one thing I had to look at twice is: some trackers make subsequent requests to other urls (that are also trackers), so if you've whitelisted a site the number blocked when its "on" vs. "off" will likely be different for many networks now now.

I might have broken this up into two separate PRs and spun out Andrey's suggestion on its own re: graph width, but its a good improvement.

The bug is just in the "major" networks count within details subview:

small-bug

MariagraziaAlastra:
@laurengarcia thanks for spotting that, it's fixed now:
screen shot 2018-02-15 at 11 40 20 am

self merging as you said in the task. 👍

PR #137, opened 2018-02-16
Add domains to whitelist from settings page

MariagraziaAlastra:
**Reviewer:** @andrey-p 
**CC:** @alohaas 

## Description:
Introduce the ability to add new domains to the whitelist from settings.
screen shot 2018-02-21 at 1 43 24 am
screen shot 2018-02-21 at 1 43 17 am
screen shot 2018-02-21 at 1 43 02 am

## Steps to test this PR:
1. Build and reload extension
2. Go to settings page
3. "Add site to whitelist" link-button should be visible, input and button should be hidden
4. Click on "Add site to whitelist". Page does not jump, and the link should then be hidden, while the input and button should be visible now
5. Input should be empty and button should be disabled and have the default cursor on hover
6. When typing in the input, the button should be enabled (blue with white font) and cursor on hover should be the pointer
7. Removing all the input should disable the button again
8. Either clicking on the button or hitting return inside the input should submit the domain
9. If the domain is not valid, an error popout should appear and the input box border should turn red. That all disappears when the user types again in the input box
10. If the domain is valid, it should be added to the whitelist, and in this case too the page will now look like at step 3
11. Clicking on any of the whitelisted domain's remove buttons should remove the respective domain from the list
12. Navigate to the whitelisted sites to make sure they're actually whitelisted

## Automated tests:
- [ ] Unit tests
- [ ] Integration tests

###### Reviewer Checklist:

###### PR Author Checklist:


shared/js/ui/models/whitelist.es6.js
```diff
@@ -21,6 +22,22 @@ Whitelist.prototype = window.$.extend({},
         value: false
       }
       })
+    },
+
+    addDomain: function (url) {
+      const parsedUrl = tldjs.tldExists(url) ? tldjs.parse(url) : ''
```
andrey-p: Not sure if `tldExists` is the best method to use. From the [readme](https://www.npmjs.com/package/tldjs#tldexists):

```js
tldExists('com');             // returns `true`
tldExists('uk');              // returns `true`
```

Might as well just parse the URL, we're checking it on L29 anyway.

MariagraziaAlastra: @andrey-p I should have left a comment here actually, but I'm using that method because otherwise random alphanumeric strings get accepted too - eg, without this check, strings like "asdasd" will be considered valid and added to the whitelist, while what we want is in the format "asdasd.com".

Also from the readme:
```js
tldExists('google.com');      // returns `true`
tldExists('google.local');    // returns `false` (not an explicit registered TLD)
```

andrey-p: Ohh got it, I see. I didn't expect `tldjs` to work this way.

Yeah, a code comment would be excellent here.

shared/js/ui/models/whitelist.es6.js
```diff
@@ -21,6 +22,25 @@ Whitelist.prototype = window.$.extend({},
         value: false
       }
       })
+    },
+
+    addDomain: function (url) {
+      // we only whitelist domains, not full URLs
+      // and getDomain will return null if the URL is invalid
+      const domain = tldjs.getDomain(url)
```
MariagraziaAlastra: So basically the old method I was using, tldExists(), was not ok because it would return true for strings like "com" and "uk", in addition to valid domains.
I also looked into isValid() but it has the same issue - accepts random strings without tld at the end, like 'asdasd'.
This method, getDomain(), returns the domain if available, otherwise null, which is exactly what we want for validation. See docs [here](https://www.npmjs.com/package/tldjs#getdomain)

shared/js/ui/views/whitelist.es6.js
```diff
@@ -15,19 +15,84 @@ function Whitelist (ops) {
 Whitelist.prototype = window.$.extend({},
   Parent.prototype,
   {
+    _isHiddenClass: 'is-hidden',
+    _isDisabledClass: 'is-disabled',
+    _isInvalidInputClass: 'is-invalid-input',
```
MariagraziaAlastra: Frankly I didn't want to declare a const for these classes in each function that uses them, and of course I don't want to just type the hardcoded string everytime - you know, typos...

I thought an external module would be a huge overkill and finally settled on this.

shared/js/ui/views/whitelist.es6.js
```diff
@@ -15,19 +15,84 @@ function Whitelist (ops) {
...
 
     _removeItem: function (e) {
-      var itemIndex = window.$(e.target).data('item')
+      const itemIndex = window.$(e.target).data('item')
+      const $listItem = window.$(e.target).parent()
       this.model.removeDomain(itemIndex)
-      this.setWhitelistFromSettings()
+
+      // No need to rerender the whole view
+      $listItem.remove()
```
MariagraziaAlastra: We used to rerender the whole view here. 
Not ok anymore because if the input is visible, by rerendering it we lose that state (and the input's content), which is a confusing behavior.

shared/js/ui/views/whitelist.es6.js
```diff
@@ -15,19 +15,84 @@ function Whitelist (ops) {
...
+        'remove',
+        'add',
+        'error',
+        'show-add',
+        'url'
+      ])
 
       this.bindEvents([
         [this.$remove, 'click', this._removeItem],
-        [this.store.subscribe, 'action:backgroundMessage', this.update]
```
MariagraziaAlastra: We don't really need this here - plus this would override my change a couple lines above and rerender the whole view when removing a whitelisted domain.

shared/scss/_mixins.scss
```diff
@@ -223,3 +223,21 @@
...
+    .is-browser--chrome & {
+        &::-webkit-input-placeholder {@content}
+    }
+    .is-browser--moz & {
+        &:-moz-placeholder {@content}
+        &::-moz-placeholder {@content}
+    }
+}
+
+@mixin arrow-left($size, $color) {
```
MariagraziaAlastra: I thought I'd leave the name as generic as possible even though it's meant mainly for popouts

shared/scss/base/base.scss
```diff
@@ -392,3 +407,40 @@ a {
         }
     }
 }
+
+/* Popout box with left arrow */
+.modal-box {
```
MariagraziaAlastra: I thought this is a generic element that we might want to maybe reuse at some point, and as such it belongs more in the base styles than in the ones specifically for the whitelist

MariagraziaAlastra:
@andrey-p back to you! I'm done with the UX for invalid URL too :)

shared/scss/base/base.scss
```diff
@@ -392,3 +407,40 @@ a {
...
+        text-align: center;
+        padding-top: 8px;
+        margin-left: 7px;
+        background-color: $white;
+        border: solid 1px #ddd;
+        color: $slate;
+        text-align: center;
+        z-index: 1;
+        float: right;
+    }
```
andrey-p: Could you make the centering of the popup a bit more consistent with the button? It looks a bit jarring when the popup appears.

screenshot from 2018-02-21 15-54-44

shared/scss/base/base.scss
```diff
@@ -315,6 +315,21 @@ a {
...
+        height: 24px;
+        vertical-align: bottom;
+
+        .is-browser--chrome & {
+            background-image: url($icon__error--chrome);
+        }
+
+        .is-browser--moz & {
+            background-image: url($icon__error--moz);
+        }
```
andrey-p: What do you think about using `@include cross_platform_bg_image("error.svg");` instead?

I've wanted to suggest changing all the places where background images are added using the `cross_platform_bg_image` mixin.

My thinking is:
- the mixin is one line in one file vs 8 lines across two files with the old method
- the compiled CSS looks exactly the same

shared/js/ui/models/whitelist.es6.js
andrey-p: The one problem is that this doesn't handle subdomains very well. E.g.:

1. Type in `andrey.duckduckgo.com` into the whitelist field
2. The entry that gets added is `duckduckgo.com`
(compare with going to `andrey.duckduckgo.com` and pressing the toggle to whitelist it, where the full `andrey.duckduckgo.com` is added)

shared/scss/_mixins.scss
```diff
@@ -223,3 +223,21 @@
...
+
+@mixin placeholder {
+    .is-browser--chrome & {
+        &::-webkit-input-placeholder {@content}
+    }
+    .is-browser--moz & {
+        &:-moz-placeholder {@content}
+        &::-moz-placeholder {@content}
+    }
+}
```
andrey-p: Oh cool, I didn't know about this `@content` thing! :+1:

shared/js/ui/models/whitelist.es6.js
MariagraziaAlastra: Good find! I can prefix it using getSubDomain - if there's no subdomain it just returns an empty string.

shared/js/ui/models/whitelist.es6.js
```diff
@@ -21,6 +22,31 @@ Whitelist.prototype = window.$.extend({},
...
       }
       })
+    },
+
+    addDomain: function (url) {
+      // We only whitelist domains, not full URLs:
+      // - use getDomain, it will return null if the URL is invalid
+      // - prefix with getSubDomain, which returns an empty string if none is found
+      // But first, strip the 'www.' part, otherwise getSubDomain will include it
+      // and whitelisting won't work for that site
```
MariagraziaAlastra: Welp, this is a long comment, but I'd rather explain what's going on than leaving it to people's best guess

MariagraziaAlastra:
@andrey-p Addressed all of your comments - once again, back to you!

shared/js/ui/views/whitelist.es6.js
```diff
@@ -17,17 +20,87 @@ Whitelist.prototype = window.$.extend({},
...
       this.model.removeDomain(itemIndex)
-      this.setWhitelistFromSettings()
+
+      // No need to rerender the whole view
+      // unless we need to show the "no sites in whitelist" message
+      if (this.$listitem && (this.$listitem.length > 1)) {
+        $clickedListItem.remove()
+      } else {
+        this.setWhitelistFromSettings()
+      }
```
andrey-p: Still seeing some issues, with this, sorry. To reproduce:

1. Go to options.html with an empty whitelist
2. Add two items
3. Remove them both
4. You won't see the "no sites in whitelist" message

I was going to say "just check the model, it should have the latest" but it somehow doesn't...

Turns out the `this.$listitem.length` doesn't update

MariagraziaAlastra: Had a chat with @andrey-p and agreed on merging this as-is and filing this issue as a known bug.

andrey-p:
Let's get this merged :+1:

PR #149, opened 2018-02-23
Implement browserify

andrey-p:
**Reviewer:** @dharb 

## Description:

**Note** this doesn't merge into develop

This actually switches over the background build system to use browserify. There's still a few warts here and there. I'm going to add loads upon loads of comments where I can.

This diff looks much nicer when you ignore whitespace: https://github.com/duckduckgo/duckduckgo-privacy-extension/compare/andrey/background-browserify...andrey/actually-browserify?w=1

## Steps to test this PR:

1. Build the extension, there should be no errors
2. Try using the extension normally, everything should look fine
3. There should be no console errors for the background page as you use the extension

Sorry I can't give you more detailed steps to test. I'm not sure if it's worth trying to catch out individual edge cases, since we're gonna integration test before `andrey/background-browserify` is merged into `develop` anyway.

## Automated tests:
- [ ] Unit tests
- [ ] Integration tests

###### Reviewer Checklist:

###### PR Author Checklist:


shared/js/abp-preprocessed.es6.js
```diff
@@ -125,11 +125,11 @@ function updateLists () {
 }
 
 // Make sure the list updater runs on start up
-settings.ready().then(() => abpLists.updateLists())
+settings.ready().then(() => updateLists())
 
 chrome.alarms.onAlarm.addListener(alarm => {
     if (alarm.name === 'updateLists') {
-        settings.ready().then(() => abpLists.updateLists())
+        settings.ready().then(() => updateLists())
```
andrey-p: Originally I wanted to consolidate all the `chrome` events before applying Browserify, hence why I added the `abpLists` thing.

I'm not doing that anymore, which is why I'm removing it...

shared/js/background.js
```diff
@@ -32,8 +35,6 @@ chrome.runtime.onMessage.addListener((req, sender, res) => {
 });
 
 function Background() {
-  $this = this;
```
andrey-p: For some reason, now that we're using Browserify, undefined variables cause an error where they didn't previously. (Maybe there's a `"use strict"` being now?)

`$this` is not defined nor used anywhere, so I removed it.

shared/js/companies.js
```diff
@@ -118,7 +121,7 @@ var Companies = (() => {
                 // uncomment for testing
                 //storageData.twitter = {count: 10, name: 'twitter', pagesSeenOn: 10}
                 storageData = Companies.sanitizeData(storageData)
-                for (company in storageData) {
+                for (let company in storageData) {
```
andrey-p: Another undefined variable that was causing problems.

There isn't a `company` (lower case) variable anywhere else in the background code, so I just scoped it to this `for` loop.

shared/js/settings.js
```diff
@@ -1,143 +1,143 @@
...
+    console.log("Settings are loaded")
+})
+
+const defaultSettings = require('../data/defaultSettings')
+
+// external settings defines a function that needs to run when a setting is updated
+let isExtensionEnabled
+var externalSettings = {
+    'httpsEverywhereEnabled': function(value){ isExtensionEnabled = value }
+};
```
andrey-p: If you look at the diff with `?w=1` in the URL you'll see that I added the line `let isExtensionEnabled`.

This is another one of those undefined global variables. `isExtensionEnabled` is not used anywhere else in the codebase, so I think this whole block may be good to remove. I've added a separate task for myself so I can come back to this later.

shared/js/tab.js
```diff
@@ -122,14 +129,4 @@ class Tab {
...
-    // count redirects
-    let tab = tabManager.get({'tabId': req.tabId})
-    if (!tab) return
-
-    if (tab.httpsRedirects[req.requestId]) {
-        tab.httpsRedirects[req.requestId] += 1
-    } else {
-        tab.httpsRedirects[req.requestId] = 1
-    }
-}, {urls: ["*://*/*"]})
```
andrey-p: Now, I know I said I was not going to move the `chrome` event listeners, but I was having trouble with this for some reason. `tabManager.get()` was erroring out as "not a function". It got fixed when I moved this chunk to `tabManager.js`.

Possibly something to do with the circular dependency between the two? I'm not sure.

shared/js/trackers.js
```diff
@@ -17,266 +17,263 @@ settings.ready().then(() => {
...
-        const whitelists = abpLists.getWhitelists()
+    }
+    return false
+}
 
-        if (whitelists.trackersWhitelist.isLoaded) {
-            match = checkABPParsedList(whitelists.trackersWhitelist.parsed, url, currLocation, request)
-        }
+function checkWhitelist(url, currLocation, request) {
+    let result = false
```
andrey-p: `result` was used further down but not defined. It didn't seem like a global, so I just defined here.

Gruntfile.js
```diff
@@ -36,39 +35,11 @@ module.exports = function(grunt) {
...
-    }
-
     /* watch any base files and browser specific files */
     let watch = {
-        sass: [['<%= dirs.src.scss %>/**/*.scss'], Object.values(fileMap[browser].sass)].join().split(','),
-        ui: [['<%= dirs.src.js %>/ui/**/*.es6.js'], Object.values(fileMap[browser].ui)].join().split(','),
-        background: [['<%= dirs.src.js %>/*.es6.js'], Object.values(fileMap[browser].background)].join().split(',')
+        sass: ['<%= dirs.src.scss %>/**/*.scss'],
+        ui: ['<%= dirs.src.js %>/ui/**/*.es6.js'],
+        background: ['<%= dirs.src.js %>/*.js']
```
andrey-p: Add data files here.

shared/js/tab.js
```diff
@@ -22,6 +22,13 @@ const scoreIconLocations = {
     "F": "img/toolbar-rating-f@2x.png"
 }
 
+const Site = require('./site')
+const Tracker = require('./tracker')
+const Score = require('./score')
+const utils = require('./utils')
+const Companies = require('./companies')
+const tabManager = require('./tabManager')
```
andrey-p: remove this line

shared/js/settings.js
```diff
@@ -1,143 +1,143 @@
-var load = require('load');
+var load = require('./load');
```
dharb: Any reason you're using `var load` here and in site.js, but `const load` elsewhere? Same for `var utils` in site.js and `var settings` in trackerLists.js among others?

shared/js/background.js
```diff
@@ -16,12 +16,15 @@
 
 
 var debugRequest = false
-var trackers = require('trackers')
-var utils = require('utils')
-var settings = require('settings')
-var stats = require('stats')
```
dharb: What's going on with stats here?

shared/js/settings.js
andrey-p: That's a good point. I was wondering whether to make them all `consts`s here or elsewhere. Lemme just change them to `const`.

shared/js/background.js
andrey-p: Oh, good question. I forgot I removed this.

It's not a module that exists anywhere, nor is it used in this file. It wasn't erroring out before, because with `require.scopes`, `stats` just ends up undefined. Browserify throws an error when you try and require a non-existent module.

andrey-p:
@dharb Updated all `require`s to use `const` :+1:

PR #153, opened 2018-02-27
Update model when removing items from whitelist

MariagraziaAlastra:
**Reviewer:** @andrey-p 
**CC:** @alohaas 

## Description:
Followup from #137, where we found out the whitelist model doesn't update its list when removing an item.
This prevents the "no whitelisted sites" message from showing when:
- Whitelist already contains more than 1 domain
- User removes all whitelisted domains at the same time

The solution in this PR does the following:
- Update the list in the model when removing an item
- Take out rendering the list items into another template
- Selectively render the above mentioned list items template when an item is removed

## Steps to test this PR:
See also screencasts:
- [No whitelisted sites, adding some](https://gfycat.com/SkinnyPerfumedIvorybackedwoodswallow)
- [Remove whitelisted sites with input hidden and visible](https://gfycat.com/CorruptBountifulBellfrog)

1. Build and reload extension
2. Go to options page (either from the hamburger menu or from "manage whitelist")
3. If no sites were added yet, the page should show the "no whitelisted sites" message
4. Click on "Add site to whitelist", type in the input, click on "Add to Whitelist" button - the site should now be added to the list
5. Try adding more sites
6. Remove sites while the input is visible- the list should update and the input should remain visible
7. Remove sites while the input is hidden - list should update and input should stay hidden
8. Remove all sites while input is visible - the "no whitelisted sites" message should show, and the input should remain visible too
9. Try reloading and mixing different steps (eg add a site, remove, reload or add a site, reload, remove) and make sure everything looks as expected
10. Also try navigating to a site, whitelisting via the toggle, then clicking on "manage whitelist" to see if the site was correctly added

## Automated tests:
- [ ] Unit tests
- [ ] Integration tests

###### Reviewer Checklist:

###### PR Author Checklist:

### DIFF
--- a/shared/js/ui/models/whitelist.es6.js
+++ b/shared/js/ui/models/whitelist.es6.js
@@ -4,7 +4,7 @@ const tldjs = require('tldjs')
 function Whitelist (attrs) {
   attrs.list = {}
   Parent.call(this, attrs)
-  
+ 
   this.setWhitelistFromSettings()
 }
 
--- a/shared/js/ui/models/whitelist.es6.js
+++ b/shared/js/ui/models/whitelist.es6.js
@@ -4,7 +4,7 @@ const tldjs = require('tldjs')
 function Whitelist (attrs) {
   attrs.list = {}
   Parent.call(this, attrs)
-
+  
   this.setWhitelistFromSettings()
 }
 
@@ -59,19 +59,19 @@ Whitelist.prototype = window.$.extend({},
 
     setWhitelistFromSettings: function () {
       let self = this
-      this.fetch({getSetting: {name: 'whitelisted'}}).then((list) => {
-        list = list || {}
-        let wlist = Object.keys(list)
+      this.fetch({getSetting: {name: 'whitelisted'}}).then((whitelist) => {
+        whitelist = whitelist || {}
+        let wlist = Object.keys(whitelist)
         wlist.sort()
-
-        // The list passed via fetch is sealed
+        
+        // Publish whitelist change notification via the store
+        // used to know when to rerender the view
+        self.set('list', wlist)
+        
+        // set() sealed both the model's list and wlist
         // unpack and wrap to allow adding/removing elements
         // in the model property
         self.list = [...wlist]
-
-        // Publish whitelist change notification via the store
-        // used to know when to rerender the view
-        self.set('whitelist', [...wlist])
       })
     }
   }
--- a/shared/js/ui/models/whitelist.es6.js
+++ b/shared/js/ui/models/whitelist.es6.js
@@ -22,6 +22,10 @@ Whitelist.prototype = window.$.extend({},
         value: false
       }
       })
+
+      // Update list
+      // use splice() so it reindexes the array
+      this.list.splice(itemIndex, 0)
     },
 
     addDomain: function (url) {
--- a/shared/js/ui/models/whitelist.es6.js
+++ b/shared/js/ui/models/whitelist.es6.js
@@ -4,7 +4,7 @@ const tldjs = require('tldjs')
 function Whitelist (attrs) {
   attrs.list = {}
   Parent.call(this, attrs)
- 
+
   this.setWhitelistFromSettings()
 }
 
--- a/shared/js/ui/models/whitelist.es6.js
+++ b/shared/js/ui/models/whitelist.es6.js
@@ -4,6 +4,8 @@ const tldjs = require('tldjs')
 function Whitelist (attrs) {
   attrs.list = {}
   Parent.call(this, attrs)
+  
+  this.setWhitelistFromSettings()
 }
 
 Whitelist.prototype = window.$.extend({},
@@ -48,9 +50,27 @@ Whitelist.prototype = window.$.extend({},
           value: true
         }
         })
+
+        this.setWhitelistFromSettings()
       }
 
       return domain
+    },
+
+    setWhitelistFromSettings: function () {
+      let self = this
+      this.fetch({getSetting: {name: 'whitelisted'}}).then((list) => {
+        list = list || {}
+        let wlist = Object.keys(list)
+        wlist.sort()
+
+        // Prevent sealing array elements
+        self.list = [...wlist]
+
+        // Publish whitelist change notification via the store
+        // used to know when to rerender the view
+        self.set('whitelist', wlist)
+      })
     }
   }
 )
--- a/shared/js/ui/models/whitelist.es6.js
+++ b/shared/js/ui/models/whitelist.es6.js
@@ -64,12 +64,14 @@ Whitelist.prototype = window.$.extend({},
         let wlist = Object.keys(list)
         wlist.sort()
 
-        // Prevent sealing array elements
+        // The list passed via fetch is sealed
+        // unpack and wrap to allow adding/removing elements
+        // in the model property
         self.list = [...wlist]
 
         // Publish whitelist change notification via the store
         // used to know when to rerender the view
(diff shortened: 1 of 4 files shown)


shared/js/ui/templates/whitelist-items.es6.js
```diff
@@ -0,0 +1,13 @@
+const bel = require('bel')
```
MariagraziaAlastra: It's easier to rerender this list selectively if I take it out in another template.

shared/js/ui/views/whitelist.es6.js
```diff
@@ -109,6 +106,12 @@ Whitelist.prototype = window.$.extend({},
       this.setup()
     },
 
+    _renderList: function () {
+      this.unbindEvents()
+      this.$container.html(whitelistItemsTemplate(this.model.list))
+      this.setup()
```
MariagraziaAlastra: Need to unbind and setup events again, even if it's just the list, otherwise after rerendering the "add to whitelist" button won't trigger an event anymore

shared/js/ui/models/whitelist.es6.js
```diff
@@ -22,6 +22,10 @@ Whitelist.prototype = window.$.extend({},
         value: false
       }
       })
+
+      // Update list
+      // use splice() so it reindexes the array
+      this.list.splice(itemIndex, 1)
```
MariagraziaAlastra: To expand on the comment, if I used 'delete list[itemIndex]' it would have been fine too, but it wouldn't have re-indexed the list.
Basically if the list was [google.com, msnbc.com]:
- After 'delete list[1]', it would be  [google.com, undefined]
- After 'this.list.splice(1, 1)', it's [google.com]

You can test the difference by console.logging the result here and then replacing splice() with delete

MariagraziaAlastra:
A note about unit tests: since we don't have a robust testing system in place for js/ui/, for now we agreed with @andrey-p to just list manual testing steps.

andrey-p:
Thanks @MariagraziaAlastra! This is definitely better but I think the way the model and view work together could be improved a lot more.

The `setWhitelistFromSettings()` method on the view gets data from the background and passes it over to the model. This really shouldn't be happening in the view. The model should do this automatically when it's initialized, and also when an item is added.

Also, the `update()` method on the whitelist view doesn't seem to be used, and can be removed.

shared/js/ui/views/whitelist.es6.js
```diff
@@ -99,7 +95,8 @@ Whitelist.prototype = window.$.extend({},
         [this.$remove, 'click', this._removeItem],
         [this.$add, 'click', this._addItem],
         [this.$showadd, 'click', this._showAddToWhitelistInput],
-        [this.$url, 'keyup', this._manageInputChange]
+        [this.$url, 'keyup', this._manageInputChange],
+        [this.store.subscribe, 'change:whitelist', this.rerender]
```
MariagraziaAlastra: Listening to the event published by the model - this will fire on load and when adding new elements

MariagraziaAlastra:
@andrey-p yeah you're right, now I moved that part to the model and I'm using the [store](https://github.com/duckduckgo/duckduckgo-privacy-extension/blob/e3b2c9be1c6a852b705a0dcbe482fb9de2dd552e/shared/js/ui/base/store.es6.js) to send notifications to the view so it knows when to rerender. Left some comments too.
Ready again!

shared/js/ui/models/whitelist.es6.js
```diff
@@ -44,9 +50,27 @@ Whitelist.prototype = window.$.extend({},
...
+
+    setWhitelistFromSettings: function () {
+      let self = this
+      this.fetch({getSetting: {name: 'whitelisted'}}).then((list) => {
+        list = list || {}
+        let wlist = Object.keys(list)
+        wlist.sort()
+
+        // Prevent sealing array elements
+        self.list = [...wlist]
```
andrey-p: I'm not sure what the benefit of `[...wlist]` is. `wlist` is an array of strings already, why unpack and wrap it again?

shared/js/ui/models/whitelist.es6.js
```diff
@@ -44,9 +50,27 @@ Whitelist.prototype = window.$.extend({},
...
+
+        // Publish whitelist change notification via the store
+        // used to know when to rerender the view
+        self.set('whitelist', wlist)
```
andrey-p: Having two properties seems superfluous. I think you can just do `self.set('list', wlist)` and listen for that property on the view?

shared/js/ui/models/whitelist.es6.js
MariagraziaAlastra: Because with regular assignment somehow self.list gets sealed and it throws an error when I try to splice() on it "TypeError: can't add/remove sealed array elements".
Unpacking and wrapping again seems like the only way around this.

andrey-p: Ohhh OK, that makes sense. I guess anything passed via `fecth` is sealed and so we can't add changes? We don't seem to be doing it on our side, at least.

Can you explain more in the comment?

MariagraziaAlastra: Yeah added a longer comment, hope it's helpful

shared/js/ui/models/whitelist.es6.js
MariagraziaAlastra: We talked about this but I'll just write it down here as well: sending the property that's actually being changed in the model doesn't actually update the view, for currently unknown reasons.
We agreed on not merging this and taking more time to understand why this happens and how to fix it, to avoid piling technical debt that will just make our lives harder later on.

shared/js/ui/views/whitelist.es6.js
```diff
@@ -99,7 +95,9 @@ Whitelist.prototype = window.$.extend({},
         [this.$remove, 'click', this._removeItem],
         [this.$add, 'click', this._addItem],
         [this.$showadd, 'click', this._showAddToWhitelistInput],
-        [this.$url, 'keyup', this._manageInputChange]
+        [this.$url, 'keyup', this._manageInputChange],
+        // listen to changes to the whitelist model
+        [this.store.subscribe, 'change:whitelist', this.rerender]
```
MariagraziaAlastra: I read better the comments in the [store](https://github.com/duckduckgo/duckduckgo-privacy-extension/blob/e3b2c9be1c6a852b705a0dcbe482fb9de2dd552e/shared/js/ui/base/store.es6.js) file, and when listening to changes, we listen to the whole model, not just to a property (unlike what we're used to) - so in this case it only works when listening to change:whitelist because whitelist is the name of the model!

shared/js/ui/models/whitelist.es6.js
```diff
@@ -44,9 +50,29 @@ Whitelist.prototype = window.$.extend({},
...
+        wlist.sort()
+        
+        // Publish whitelist change notification via the store
+        // used to know when to rerender the view
+        self.set('list', wlist)
+        
+        // set() sealed both the model's list and wlist
+        // unpack and wrap to allow adding/removing elements
+        // in the model property
+        self.list = [...wlist]
```
MariagraziaAlastra: It's not pretty but I need to redo the assignment here because, as explained in the comment, set() seals both model.list and the value, wlist.
So even if I pass [...wlist] to unpack it inside set (like this: self.set('list', [...wlist]) it's useless because it gets sealed again.

andrey-p:
Hope you don't mind, jumped in and removed that console.log. I want to discuss that freezing thing separately, but I think this is good to merge!

PR #161, opened 2018-03-05
Prevent model from accidentally freezing its own properties

andrey-p:

## Description:

In https://github.com/duckduckgo/duckduckgo-privacy-extension/pull/153 we noticed that the model's own properties were accidentally getting frozen (by [this line](https://github.com/duckduckgo/duckduckgo-privacy-extension/blob/7d0afb1dbdd0430e30d92f1228d654d28364c7d4/shared/js/ui/base/store.es6.js#L60)) because they were being passed by reference.

This PR makes sure anything passed to the store is deep cloned so that problem doesn't happen. It also removes a hack we had to add in the whitelist model.

## Steps to test this PR:

1. Try adding and removing some items in the whitelist - it should work without any errors
2. Other models should also work fine.

## Automated tests:
- [ ] Unit tests
- [ ] Integration tests

###### Reviewer Checklist:

###### PR Author Checklist:

### DIFF
--- a/shared/js/ui/base/model.es6.js
+++ b/shared/js/ui/base/model.es6.js
@@ -57,6 +57,9 @@ BaseModel.prototype = $.extend({},
       const lastValue = this[attr] || null
       this[attr] = val
 
+      // deep clone val in case it's something passed by reference
+      val = JSON.parse(JSON.stringify(val))
+
       this.store.publish({
         notifierName: this.modelName,
         change: {attribute: attr, value: val, lastValue: lastValue},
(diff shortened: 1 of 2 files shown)


shared/js/ui/base/model.es6.js
```diff
@@ -57,6 +57,9 @@ BaseModel.prototype = $.extend({},
       const lastValue = this[attr] || null
       this[attr] = val
 
+      // deep clone val in case it's something passed by reference
+      val = JSON.parse(JSON.stringify(val))
```
andrey-p: I was considering doing this to `lastValue` as well. Not sure if it's worth the extra processing time for it, through.

MariagraziaAlastra: I think it's fine, lastValue seems something we wouldn't modify after sending a notification

MariagraziaAlastra: @andrey-p would you mind checking whether JSON.stringify(val) is defined though? If val is null, that method returns undefined, and JSON.parse() will throw a SyntaxError.

andrey-p: Good find, lemme fix!

PR #171, opened 2018-03-15
HTTPS fixes

andrey-p:
**Reviewer:** @jdorweiler 

## Description:

This fixes a couple of issues with HTTPS redirection that make it slower for cases where we can't redirect. See inline for more info.

Redirects should still work as before.

## Steps to test this PR:

1. Visit e.g. http://theguardian.com
2. You should be taken to the HTTPS version
3. The popup should say "enforced encryption"

## Automated tests:
- [ ] Unit tests
- [ ] Integration tests

###### Reviewer Checklist:

###### PR Author Checklist:

### DIFF
--- a/shared/js/background/https.es6.js
+++ b/shared/js/background/https.es6.js
@@ -252,7 +252,9 @@ class HTTPS {
             // If it's already https, it should be the same:
             ['https://duckduckgo.com',                      'https://duckduckgo.com'],
             // If it's not in the list, it should stay http:
-            ['http://fdsakljfsa.fr',                        'http://fdsakljfsa.fr']
+            ['http://fdsakljfsa.fr',                        'http://fdsakljfsa.fr'],
+            // any other protocols should be left the same:
+            ['file:///home/foo/bar',                        'file:///home/foo/bar']
         ]
 
         let passed = true
--- a/shared/js/background/https.es6.js
+++ b/shared/js/background/https.es6.js
@@ -215,7 +215,6 @@ class HTTPS {
         const testHosts = [
             '1337x.to',
             'submit.pandora.com',
-            '*.api.roblox.com',
             'thump.vice.com',
             'yts.ag'
         ]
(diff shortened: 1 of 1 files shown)


shared/js/background/https.es6.js
```diff
@@ -178,7 +178,7 @@ class HTTPS {
 
         // Only deal with http calls
         const protocol = utils.getProtocol(reqUrl).toLowerCase()
-        if (!protocol.indexOf('http:') === 0) {
+        if (protocol !== 'http:') {
```
andrey-p: The problem here is that `!protocol.indexOf('http:')` evaluated before the equality, so the if statement ended up as either `if (true === 0)` or `if (false === 0)`.

This means we were running the linear search for e.g. HTTPS sites where it wasn't necessary.

shared/js/background/https.es6.js
```diff
@@ -196,20 +196,9 @@ class HTTPS {
...
-        }
-
-        // Check for upgrades
-        for (let i=0; i<hosts.length; i++) {
-            if (this.canUpgradeHost(hosts[i])) {
-                return reqUrl.replace(/^(http|https):\/\//i, 'https://')
-            }
+        if (this.canUpgradeHost(host)) {
+            return reqUrl.replace(/^(http|https):\/\//i, 'https://')
         }
```
andrey-p: Previously, we had wildcard rules like `*.foo.com`, which is why we had an extra check for the wildcard version.

We don't have wildcard rules at the moment, so we unnecessarily run an extra linear search for any URLs with subdomains.

shared/js/background/https.es6.js
```diff
@@ -226,7 +215,6 @@ class HTTPS {
         const testHosts = [
             '1337x.to',
             'submit.pandora.com',
-            '*.api.roblox.com',
```
andrey-p: I couldn't actually get the tests to work :man_shrugging: 

I'm hoping to fix them soon, so at least I wanted to keep the test cases a bit more up-to-date.

jdorweiler:
Looks good, thanks.

Issue #101, opened 2018-01-27
[Safari] Please add link to this Github-Repo


clemensbartz:
### Information about your system

- Extension Version: 2018.1.25
- Browser and Version: Safari 11.0.3
- Operating System: MacOS 10.13.3 High Sierra

### What does not work?

When I click on "more options" (Hamburger Menu), I cannot find a link to this repository.

### What is the expected behavior?

Please add a link to this repository in the "more options" menu and/or in the settings menu.

alohaas:
Hi @clemensbartz, thank you for sharing this feedback. We handle suggestions like this internally so I will be bringing your thoughtful request to the team.

Thanks again!

brindy:
Hi folks,

Things have moved on a bit since this was last updated.

We have a new extension for Safari that uses Tracker Radar and you can install it from the App Store:
https://apps.apple.com/us/app/duckduckgo-privacy-essentials/id1482920575?mt=12

You can find the code for it here:
https://github.com/duckduckgo/privacy-essentials-safari

Thanks!

PR #178, opened 2018-03-22
Add link back to source code

nickthetait:
I'm not terribly experienced writing HTML or javascript yet, so this is merely a "proposed fix" for #101


kdzwinel:
Thanks, this is a cool idea and we should consider it in the future. This PR is very outdated now so I'll close it.

PR #181, opened 2018-03-26
UX for listing unblocked first party requests + entity list changes

MariagraziaAlastra:
**Reviewer:** @andrey-p 

## Description:
MVP UX implementation of the first party unblocked requests list and updates to the entities list.
The unblocked domains paragraph lives in the tracker networks subview, at the bottom. 

## Steps to test this PR:
1. Build and reload extension
2. Go to to a major tracker network's website
3. Open popup
4. Click on the tracker networks line
5. The subview should contain a list of unblocked first party requests at the bottom, separated by a grey top border
6. Also check the unknown company trackers, if applicable: these should be positioned at the bottom, just before the unblocked list 
7. Visit sites like theguardian.com which have their own trackers and are normally listed under unknown - in this case the same domain unblocked trackers should show up in a separate unblocked section
8. Try whitelisting the sites. In that case we still want to show the same domain list, but without "(not blocked)" in the text
9. Run trackers unit tests. These are checking that isTracker returns the correct parent and reason data.

## Automated tests:
- [ ] Unit tests
- [ ] Integration tests

###### Reviewer Checklist:

###### PR Author Checklist:


scripts/buildEntityMap.js
```diff
@@ -1,23 +1,28 @@
 var request = require('request'),
     fs = require('fs');
 
-request('https://raw.githubusercontent.com/mozilla-services/shavar-prod-lists/master/disconnect-entitylist.json', function (err, res, body) {
+let requestData = {
+    method: 'get',
+    uri: 'https://jason.duckduckgo.com/contentblocking.js?l=entitylist2',
```
russellholt: flagging for final update

shared/data/tracker_lists/trackersWithParentCompany.json
```diff
@@ -1151,39 +1111,39 @@
             "u": "https://aloodo.com/"
         },
         "amazon-adsystem.com": {
-            "c": "Amazon",
+            "c": "Amazon.com",
```
russellholt: I thought we already updated Amazon.com. Was it just not pushed to the repo?

shared/js/ui/models/site.es6.js
```diff
@@ -185,7 +185,15 @@ Site.prototype = window.$.extend({},
     getUniqueTrackersBlockedCount: function () {
       // console.log('[model] getUniqueTrackersBlockedCount()')
       const count = Object.keys(this.tab.trackersBlocked).reduce((total, name) => {
-        return this.tab.trackersBlocked[name].urls.length + total
```
MariagraziaAlastra: Had to update getUniqueTrackersBlockedCount() and getUnknownTrackersCount() because of the new trackers structure. In both function there was a direct length check on the urls array, which is now an object.
Also added a bit more defensive checks just to make sure we don't error on Object.keys() and the like

shared/js/ui/templates/tracker-networks.es6.js
```diff
@@ -42,22 +42,31 @@ function renderHero (site) {
...
   if (companyListMap && companyListMap.length > 0) {
     return companyListMap.map((c, i) => {
-      if (c.name && c.name === 'unknown') c.name = '(Tracker network unknown)'
-      return bel`<li>
+      let borderClass = ''
+      if (c.name && c.name === 'unknown') {
+        c.name = '(Tracker network unknown)'
+      } else if (c.name && model.hasUnblockedTrackers(c)) {
+        const additionalText = ' associated domains'
+        const domain = model.site ? model.site.domain : c.name
```
MariagraziaAlastra: Showing the site domain instead of the company name for the unblocked paragraph, as requested in the task. The relationship between the site domain and the owner companies has yet to be explored for this scenario.
However, it's ok to show the company name as usual when we're blocking third party trackers.
Please feel free to reach out if this is unclear.

shared/data/tracker_lists/trackersWithParentCompany.json
andrey-p: Yeah, it wasn't pushed. I wonder if it's worth just adding this to `.gitignore` if it's regenerated every time you build the extension anyway.

shared/js/background/classes/tracker.es6.js
```diff
@@ -16,9 +17,9 @@ class Tracker {
      * to track you through many different entities.
      * We store a list of all unique urls here.
      */
-    addURL (url) {
-        if (this.urls.indexOf(url) === -1) {
-            this.urls.push(url)
+    update (t) {
+        if (!this.urls[t.url]) {
+            this.urls[t.url] = {isBlocked: t.block, reason: t.reason}
```
andrey-p: @MariagraziaAlastra  @jdorweiler There's multple places where we have to call `Object.keys` on this object. Why not store these both as an array and an object?

shared/data/tracker_lists/trackersWithParentCompany.json
jdorweiler: Yeah .gitignore works. The amazon change never made it to our backend service when we started using it for the mobile app so it's been `amazon.com` for awhile anyway.

shared/js/ui/models/site-company-list.es6.js
```diff
@@ -41,10 +41,11 @@ SiteCompanyList.prototype = window.$.extend({},
...
-          // don't sort "unknown" trackers since they will
+          // don't sort "unknown" nor unblocked trackers since they will
           // be listed individually at bottom of graph,
           // we don't want "unknown" tracker total as maxCount
-          if (name !== 'unknown') {
+          if (name !== 'unknown' || 
+            this.hasUnblockedTrackers(this.trackers[name])) {
             let compare = this.trackers[name].count
             if (compare > maxCount) maxCount = compare
           }
```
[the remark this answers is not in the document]
MariagraziaAlastra: @andrey-p I think you're right, I tested without and it doesn't seem to affect anything. Removed, thanks!

shared/js/background/classes/tracker.es6.js
jdorweiler: This update method runs in request blocking code so pushing that off to the UX is probably best for performance.

andrey-p: Hmm that's fair. Any chance we could just do it once on the UI code, then?

MariagraziaAlastra: @andrey-p how about I add the urls array to the company object in companyListMap

andrey-p: @MariagraziaAlastra Yeah that sounds good. Do it!

jdorweiler:
@andrey-p background code is ready to review again.

shared/js/background/classes/tracker.es6.js
MariagraziaAlastra: @andrey-p Ok so just writing here what we talked about via dm: I reduced the calls from companyListMap (which also allows us to not call Object.keys() in the tracker networks template) but unfortunately the site model still needs to do that, because we use trackersBlocked and trackers there, not the companyListMap.

I think we'll need to figure out an adequate structure with a followup tech design.

MariagraziaAlastra:
@andrey-p UI ready too!

andrey-p:
Thank you both for humouring me! Lemme test again.

shared/js/background/classes/tracker.es6.js
russellholt: @andrey-p - a better way to frame your question is: it is not convenient to iterate over a javascript plain object. I don't think the answer is to store the same information in two ways without a shared interface to it. You either a) punt, b) hide the Object.keys stuff behind another interface, or c) replace it with a better structure, like a Map.

I don't think right now is the time to change this. Object.keys might be slower or inconvenient, but this discussion sounds like it's digging the hole deeper.

andrey-p:
This is solid as far as I can tell :+1:

Two minor things:

1. The url to `entitylist2` should be updated [here](https://github.com/duckduckgo/duckduckgo-privacy-extension/pull/181/files#diff-e3627ded6c3c461e7ed62faf1e022464R6) and [here](https://github.com/duckduckgo/duckduckgo-privacy-extension/pull/181/files#diff-7399d3e94592ef48695985e91f915f10R6) (I'd do it now but the servers need to be updated)
2. I'm still seeing local uncommitted changes to `shared/data/tracker_lists/entityMap.json`. Let's wait until the servers are updated, then commit the latest version.

andrey-p:
Merging!

PR #182, opened 2018-03-27
Grade testing tool

andrey-p:
**Reviewer:** TBD

**CC:** @russellholt 

## Description:
Builds on https://github.com/duckduckgo/duckduckgo-privacy-extension/compare/r/gd2

## Steps to test this PR:

1. Prepare a list of newline-separated list of domains like:

```
theguardian.com
reddit.com
news.ycombinator.com
```

2. Rebuild stuff:

```
$ grunt browserify:backgroundTest --browser=chrome --type=dev
$ npm run dev-chrome
```

3. Grab site info, e.g.

```
$ ./selenium-test/ratings.js -f [list of files] -o foo -x
```

This will save JSON site info at `foo-sites/`

4. Run any post processing scripts, e.g.:

```
$ node selenium-test/post-process/csv.js foo
$ node selenium-test/post-process/tracker-entities.js foo
```

## Automated tests:
- [ ] Unit tests
- [ ] Integration tests

###### Reviewer Checklist:

###### PR Author Checklist:

### DIFF
--- a/selenium-test/lib/ratings.js
+++ b/selenium-test/lib/ratings.js
@@ -1,4 +1,5 @@
 const fs = require('fs');
+const childProcess = require('child_process');
 const {Builder, By, until, promise} = require('selenium-webdriver');
 const chrome = require('selenium-webdriver/chrome');
 const chromedriver = require('chromedriver');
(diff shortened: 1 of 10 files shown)


test/tests/gradedetails.js
```diff
@@ -0,0 +1,69 @@
...
+ * Navigate to a url
+ */
+function runTest(url) {
+    return new Promise((resolve) => {
+
+        let blockingOnStartTime = Date.now();
+
+        chrome.tabs.create({url}, (t) => {
+
+            getLoadedTabById(t.id, blockingOnStartTime, 20000, 9000).then((tab) => {
```
andrey-p: Had to add a crazy long wait here for extra accuracy (so this'll wait 20s for the site to timeout, and then an extra 9s after the tab is loaded).

shared/js/background/classes/score.es6.js
```diff
@@ -106,26 +120,69 @@ class Score {
...
 
-        if (this.inMajorTrackingNetwork) beforeIndex++
+        if (this.inMajorTrackingNetwork) {
+            beforeIndex++
+            this.addDecision({ change: 1, index: beforeIndex, why: `In major tracking network` })
+        } else {
+            this.addDecision({ change: 1, index: beforeIndex, why: `Not in major tracking network` })
+        }
+
+        this.hasHTTPS = this.hasHTTPS || https.canUpgradeHost(this.domain)
```
andrey-p: This is a temporary hack to guard against the HTTPS inconsistencies we were seeing. If we end up merging this into the develop branch, this is something we'll want to drop.

selenium-test/lib/ratings.js
```diff
@@ -166,25 +80,56 @@ exports.testUrl = function(path, opts) {
 
 exports.testUrls = async function(urlArray, opts) {
     return new Promise (async (resolve, reject) => {
+        // create output dir if it didn't exist already
+        let outputDir = `${process.cwd()}/${opts.output}-sites/`;
+        childProcess.execSync(`mkdir -p ${outputDir}`);
```
russellholt: why this and not `fs.mkdirSync` here?

shared/js/background/classes/score.es6.js
russellholt: The decision to do it this way could also be moved to a post processing step.

andrey-p: That's a good idea, but that can potentially affect which grade letter is picked, and we'd need to copy over that logic to the postprocessing step. Not sure if we care that much about the grade letter, though.

russellholt: True. So from a testing standpoint I think this should be the default behavior, but then perhaps in the future an option to choose which method is used might be useful.

PR #199, opened 2018-04-10
Allow localhost to be added to white list

dharb:

## Description: 
Update validation to allow localhost to be added to whitelist on settings page.

## Steps to test this PR:

1. On the options page, add localhost to whitelist. Now visit a localhost address and open the popup. Privacy protection should be turned off.
2. Visit options page again and remove localhost from the whitelist. Privacy protection should be back on for localhost.
3. Add other urls both valid and invalid to the whitelist. Check that the validation still works properly.
4. `npm run test`

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -34,7 +34,6 @@ module.exports = function(grunt) {
         },
         unitTest: {
             '<%= dirs.unitTest.build %>/background.js': ['<%= dirs.unitTest.background %>/**/*.js'],
-            // TODO uncomment this when we add some UI tests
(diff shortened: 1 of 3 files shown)


MariagraziaAlastra:
@dharb Not sure what's happening here, but after I try adding localhost to the whitelisted domains, I can't seem to add anything else.
Steps to reproduce:
- Go to options page and add localhost
- Now try adding another domain, eg amazon.com
- After clicking on the button, all domains disappear from the whitelist and it's empty again (??)
- Go to the second site you added - it should actually be whitelisted

I think something's up with rerendering the view in that case. 
I tried the above steps on develop and it looks fine

dharb:
I think somehow having that index line at the top of whitelist.es6.js was screwing up the view. Still not sure why, but removing it fixed the issue and the tests still pass.

PR #211, opened 2018-04-25
4 spaces, lint everywhere

andrey-p:
**Reviewer:** self

**CC:** @MariagraziaAlastra @jdorweiler @russellholt @dharb 

## Description:

Finally trying to standardize all the code across the entirety of the codebase. This is with `--fix`. There will be some manual changes as well, in a separate PR.

To see a diff without all the whitespace changes see: https://github.com/duckduckgo/duckduckgo-privacy-extension/compare/develop...andrey/4-spaces?w=1

## Automated tests:
- [ ] Unit tests
- [ ] Integration tests

###### Reviewer Checklist:
- [ ] **Ensure the PR solves the problem**
- [ ] **Review every line of code**
- [ ] **Ensure the PR does no harm by testing the changes thoroughly**
- [ ] **Get help if you're uncomfortable with any of the above!**
- [ ] Determine if there are any quick wins that improve the implementation

###### PR Author Checklist:
- [ ] Get advice or leverage existing code
- [ ] Agree on technical approach with reviewer (if the changes are nuanced)
- [ ] Ensure that there is a testing strategy (and documented non-automated tests)
- [ ] Ensure there is a documented monitoring strategy (if necessary)
- [ ] Consider systems implications

PR #212, opened 2018-04-25
Lint manual fixes

andrey-p:

**CC:** @dharb @jdorweiler @russellholt 

## Description:

Fixes a lot of small tweaks that Lint was complaining about. Some of them are nitpicky, but I figured we might as well do them. A couple are possible bugs.

## Steps to test this PR:

This is based on some of the slightly more hairy changes in the diff. Sorry about vagueness.

A lot of this is covered in the beta tests anyway, so we should be fine.

1. Lint should pass without errors
2. The `atb` query string data should be set correctly on searches when installing the extension (note: we'd only be able to test that properly when the Beta is out)
3. Grade should be set correctly in the popup
4. Fetching the HTTPS list should work correctly in FF/Chrome
5. Trackers should be successfully blocked in Safari, and surrogates should still run
6. Tab management should successfully work in Safari (e.g. click between tabs and the grade should uptade).

## Automated tests:
- [ ] Unit tests
- [ ] Integration tests

###### Reviewer Checklist:

###### PR Author Checklist:

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,10 +7,8 @@ npm-debug.log
 *.map
 public/js/*.*
 public/css/*.*
-shared/js/abp.js
 *.swp
 shared/data/tracker_lists/trackersWithParentCompany.json
-shared/js/tldjs.js
 build/firefox
 build/chrome
 build/safari
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,9 @@
 language: node_js
 node_js:
   - 8.11.1
+script: npm run lint && npm test
 before_script:
   - npm install -g grunt-cli
+cache:
+  directories:
+    - "node_modules"
--- a/shared/js/background/abp-lists.es6.js
+++ b/shared/js/background/abp-lists.es6.js
@@ -1,16 +1,11 @@
 /*
  * Load the abp-filter-parser node module and
  * pre-process the easylists.
- *
- * This will be browserifyed and turned into abp.js by running 'grunt'
  */
 const abp = require('abp-filter-parser')
-const deepFreeze = require('deep-freeze')
 const constants = require('../../data/constants')
-const defaultSettings = require('../../data/defaultSettings')
 const surrogates = require('./surrogates.es6')
 const settings = require('./settings.es6')
-const ATB = require('./atb.es6')
 const load = require('./load.es6')
 const browserWrapper = require('./$BROWSER-wrapper.es6')
 
@@ -71,7 +66,7 @@ function getWhitelists () {
  */
 function updateLists () {
     const atb = settings.getSetting('atb')
-    const set_atb = settings.getSetting('set_atb')
+    const setAtb = settings.getSetting('set_atb')
     const versionParam = getVersionParam()
 
     for (let listType in lists) {
@@ -87,7 +82,7 @@ function updateLists () {
             // only add url params to contentblocking.js duckduckgo urls
             if (url.match(/^https?:\/\/(.+)?duckduckgo.com\/contentblocking\.js/)) {
                 if (atb) url += '&atb=' + atb
-                if (set_atb) url += '&set_atb=' + set_atb
+                if (setAtb) url += '&set_atb=' + setAtb
                 if (versionParam) url += versionParam
             }
 
--- a/shared/js/background/atb.es6.js
+++ b/shared/js/background/atb.es6.js
@@ -10,8 +10,8 @@ var ATB = (() => {
     return {
         updateSetAtb: () => {
             return new Promise((resolve) => {
-                let atbSetting = settings.getSetting('atb'),
-                    setAtbSetting = settings.getSetting('set_atb')
+                let atbSetting = settings.getSetting('atb')
+                let setAtbSetting = settings.getSetting('set_atb')
 
                 if (!atbSetting || !setAtbSetting) { resolve(null) }
 
@@ -30,7 +30,7 @@ var ATB = (() => {
 
                 xhr.onreadystatechange = function () {
                     if (xhr.readyState === XMLHttpRequest.DONE) {
-                        if (xhr.status == 200) {
+                        if (xhr.status === 200) {
                             let curATB = JSON.parse(xhr.responseText)
                             resolve(curATB.version)
                         }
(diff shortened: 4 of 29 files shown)


.travis.yml
```diff
@@ -1,5 +1,9 @@
 language: node_js
 node_js:
   - 8.11.1
+script: npm run lint && npm test
```
andrey-p: This adds CI support.

.travis.yml
```diff
@@ -1,5 +1,9 @@
...
 before_script:
   - npm install -g grunt-cli
+cache:
+  directories:
+    - "node_modules"
```
andrey-p: Not really related to this PR, but it caches the `node_modules` directory between CI runs (updating it where necessary). It should speed things up by a lot.

shared/js/background/abp-lists.es6.js
```diff
@@ -1,16 +1,11 @@
...
- *
- * This will be browserifyed and turned into abp.js by running 'grunt'
  */
 const abp = require('abp-filter-parser')
-const deepFreeze = require('deep-freeze')
 const constants = require('../../data/constants')
-const defaultSettings = require('../../data/defaultSettings')
 const surrogates = require('./surrogates.es6')
 const settings = require('./settings.es6')
-const ATB = require('./atb.es6')
```
andrey-p: A lot of the variable declarations I've removed were marked as not used anywhere.

shared/js/background/abp-lists.es6.js
```diff
@@ -87,7 +82,7 @@ function updateLists () {
             // only add url params to contentblocking.js duckduckgo urls
             if (url.match(/^https?:\/\/(.+)?duckduckgo.com\/contentblocking\.js/)) {
                 if (atb) url += '&atb=' + atb
-                if (set_atb) url += '&set_atb=' + set_atb
+                if (setAtb) url += '&set_atb=' + setAtb
```
andrey-p: Casing complaint.

shared/js/background/atb.es6.js
```diff
@@ -30,7 +30,7 @@ var ATB = (() => {
 
                 xhr.onreadystatechange = function () {
                     if (xhr.readyState === XMLHttpRequest.DONE) {
-                        if (xhr.status == 200) {
+                        if (xhr.status === 200) {
```
andrey-p: Strict equality here is safe, `xhr.status` is `unsigned short` [according to MDN](https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/status).

shared/js/background/atb.es6.js
```diff
@@ -82,22 +82,23 @@ var ATB = (() => {
...
+            let localTime = localDate.getTime()
+            let utcTime = localTime + (localDate.getTimezoneOffset() * oneMinute)
+            let est = new Date(utcTime + (oneHour * -5))
+            let dstStartDay = 13 - ((est.getFullYear() - 2016) % 6)
+            let dstStopDay = 6 - ((est.getFullYear() - 2016) % 6)
+            let isDST = (est.getMonth() > 2 || (est.getMonth() === 2 && est.getDate() >= dstStartDay)) && (est.getMonth() < 10 || (est.getMonth() === 10 && est.getDate() < dstStopDay))
+            let epoch = isDST ? estEpoch - oneHour : estEpoch
+            let timeSinceEpoch = new Date().getTime() - epoch
+            let majorVersion = Math.ceil(timeSinceEpoch / oneWeek)
+            let minorVersion = Math.ceil(timeSinceEpoch % oneWeek / oneDay)
```
andrey-p: This felt a bit opinionated but ah well. :man_shrugging: Note I changed `isDST` to use strict equality for `getMonth`, too. This is safe, since the `getMonth` returns a number and not a string ([docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/getMonth))

shared/js/background/background.es6.js
```diff
@@ -18,21 +18,17 @@
...
+    if (window.navigator.userAgent.indexOf('Mac') !== -1) os = 'm'
+    if (window.navigator.userAgent.indexOf('Linux') !== -1) os = 'l'
 
     localStorage['os'] = os
 
     events.onStartup()
-}
-
-var background
-settings.ready().then(() => new Background())
```
andrey-p: Lint was complaining that `background` wasn't used. I didn't really see the point in having `background` as a class if all it did was call some things when settings are ready.

shared/js/background/classes/score.es6.js
```diff
@@ -62,6 +62,8 @@ class Score {
                     },
                     message: message
                 }
+
+                return result
```
andrey-p: Lint complains about assignments in `return` statements. See https://eslint.org/docs/rules/no-return-assign for the motivation behind it.

shared/js/background/https.es6.js
```diff
@@ -99,40 +99,19 @@ class HTTPS {
...
-                    data += localStorage['https-upgrade-list' + i]
-                }
-
-                try {
-                    let parsedList = JSON.parse(data)
-                    resolve(parsedList)
-                } catch (e) {
-                    console.log('HTTPS: getFromStorage() error parsing JSON from localStorage', e)
-                    reject()
-                }
```
andrey-p: This was a bit of a weird one.

`LOCAL_STORAGE_CHUNKS` wasn't defined anywhere. When this runs (if it runs?) the for statement would evaluate to `0 < undefined` which returns false. So we'd always end up with a rejection anyway.

We're not using HTTPS in Safari anyway, so I just removed this.

shared/js/background/safari-events.es6.js
```diff
@@ -158,7 +157,7 @@ let onBeforeRequest = (requestData) => {
 
     let redirectRequest = redirect.handleRequest(requestData)
     if (redirectRequest) {
-        redirectRequest
+        return redirectRequest
```
andrey-p: cc @jdorweiler - not sure if this was a bug or not. It seems like we would've noticed it if trackers weren't blocked properly on Safari.

shared/js/background/safari-wrapper.es6.js
```diff
@@ -1,3 +1,6 @@
+/* global safari:false */
+const tabManager = require('./tab-manager.es6')
+
```
andrey-p: cc @jdorweiler - `tabManager` is used elsewhere in the file, but it wasn't defined. Could that be behind all the weirdness you were seeing?

shared/js/background/tracker-lists.es6.js
```diff
@@ -6,10 +6,6 @@ function getLists () {
     return lists
 }
 
-function setList (name, data) {
-    lists[name] = data
-}
-
```
andrey-p: Not exported or used anywhere.

shared/js/background/safari-wrapper.es6.js
```diff
@@ -1,3 +1,6 @@
+/* global safari:false */
```
andrey-p: I ended up defining `safari` as a global in all the Safari-specific modules. Anywhere outside, I used `window.safari` instead.

My thinking was that it's guaranteed that Safari-specific modules will have `safari` defined as a global, but other files won't necessarily have that.

shared/js/content-scripts/content-script.js
```diff
@@ -49,15 +49,15 @@ function getLocation () {
     // content script only works from window.top and will throw
     // errors if it runs in iframes. Try to access hostname and
     // bail if there are any errors.
+    if (mainFrameURL) { return mainFrameURL }
+
     try {
-        window.top.location.hostname
+        var loc = window.top.location
+        mainFrameURL = loc.protocol + '//' + loc.hostname + loc.pathname
```
andrey-p: Lint was complaining about having a bare var reference without any operations inside the `try` statement. I decided it's better to just move the bits we need in the `try`.

shared/js/ui/templates/privacy-practices.es6.js
```diff
@@ -17,11 +17,11 @@ module.exports = function () {
...
-            status: tosdrStatus,
-            title: domain,
-            subtitle: `${tosdrMsg} Privacy Practices`,
-            showClose: true
-            })}
+        status: tosdrStatus,
+        title: domain,
+        subtitle: `${tosdrMsg} Privacy Practices`,
+        showClose: true
+    })}
```
andrey-p: Re: all the whitespace changes in the templates, I tweaked a few of them manually in #211. I then ran `--fix` here, and Eslint reverted them to look like this. Ah well.

shared/js/background/safari-events.es6.js
jdorweiler: Yeah still works but it should have a return.

shared/js/background/safari-wrapper.es6.js
jdorweiler: Defined as a global here for the UI to use: https://github.com/duckduckgo/duckduckgo-privacy-extension/blob/master/shared/js/background/safari-events.es6.js#L10

Doesn't hurt to have it required here just in case we ever fix the UI active tab problem.

MariagraziaAlastra:
@andrey-p this works for me! Let's remember to delete tldjs and abp.js locally, otherwise it will error 👍

.gitignore
```diff
@@ -7,10 +7,8 @@ npm-debug.log
 *.map
 public/js/*.*
 public/css/*.*
-shared/js/abp.js
 *.swp
 shared/data/tracker_lists/trackersWithParentCompany.json
-shared/js/tldjs.js
```
MariagraziaAlastra: 😃

PR #215, opened 2018-04-27
Tidy up globals in Safari

andrey-p:
**Reviewer:** @jdorweiler 

**CC:** @MariagraziaAlastra 

## Description:

This drops the use of globals on `contentWindow` for Safari. See inline comments for more info.

## Steps to test this PR:

1. Make Safari passes information correctly between the background tab and the popup:
    - number of trackers blocked updates
    - top trackers gets updated and reset correctly
    - whitelisting via the toggle works
    - grade is shown correctly
2. Make sure Safari passes information correctly between the background tab and the options page:
    - the embedded tweet blocking toggle should work
    - you can manually add / remove sites to the whitelist, and that gets updated in the popup
3. When you whitelist a site via the popup in Chrome/FF, the popup should close and the page should refresh.

## Automated tests:
- [ ] Unit tests
- [ ] Integration tests

###### Reviewer Checklist:

###### PR Author Checklist:

### DIFF
--- a/shared/js/background/safari-events.es6.js
+++ b/shared/js/background/safari-events.es6.js
@@ -128,7 +128,7 @@ safari.extension.globalPage.contentWindow.message = handleUIMessage
 let updateSetting = (e) => {
     let name = e.message.updateSetting.name
     let val = e.message.updateSetting.value
-    if (name && val) {
+    if (name) {
         settings.updateSetting(name, val)
     }
 }
--- a/shared/js/background/safari-events.es6.js
+++ b/shared/js/background/safari-events.es6.js
@@ -6,10 +6,6 @@ const settings = require('./settings.es6')
 const abpLists = require('./abp-lists.es6')
 const browserWrapper = require('./safari-wrapper.es6')
 
-// add these to contentWindow so the popup can access them
-safari.extension.globalPage.contentWindow.tabManager = tabManager
-safari.extension.globalPage.contentWindow.Companies = Companies
-
 let _getSafariTabIndex = (target) => {
     for (let i = 0; i < safari.application.activeBrowserWindow.tabs.length; i++) {
         if (target === safari.application.activeBrowserWindow.tabs[i]) {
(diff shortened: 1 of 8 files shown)


shared/js/background/safari-events.es6.js
```diff
@@ -97,10 +93,42 @@ let handleMessage = (e) => {
...
+    } else if (req.getSiteScore) {
+        let tab = tabManager.get({tabId: req.getSiteScore})
+        if (tab) res(tab.site.score.get())
+    } else if (req.getTopBlockedByPages) {
+        res(Companies.getTopBlockedByPages(req.getTopBlockedByPages))
+    } else if (req.resetTrackersData) {
+        Companies.resetData()
+        safari.self.hide()
+    }
+}
```
andrey-p: The idea here is, move the stuff that belongs in the background process to the background. That way you wouldn't need to expose `tabManager` and `companies` in any way.

shared/js/background/safari-events.es6.js
```diff
@@ -97,10 +93,42 @@ let handleMessage = (e) => {
...
+
+safari.extension.globalPage.contentWindow.message = handleUIMessage
+
 let updateSetting = (e) => {
     let name = e.message.updateSetting.name
     let val = e.message.updateSetting.value
-    if (name && val) {
+    if (name) {
```
andrey-p: This change is not necessarily related to this PR, but I noticed that you can't turn off the twitter embed toggle in the options page, since `val` is falsy. I should've spotted this in review.

shared/js/background/safari-wrapper.es6.js
```diff
@@ -1,6 +1,4 @@
-/* global safari:false tabManager:false */
-// const tabManager = require('./tab-manager.es6')
```
andrey-p: Turned out this was getting messed up because of a circular dependency. I'm sure we've got these elsewhere but ah well.

shared/js/ui/base/safari-ui-wrapper.es6.js
```diff
@@ -1,45 +1,58 @@
...
+        safari.extension.globalPage &&
+        safari.extension.globalPage.contentWindow) {
+    context = 'popup'
+} else if (safari &&
+        safari.self &&
+        safari.self.tab) {
+    context = 'options'
+} else {
+    throw new Error('safari-ui-wrapper couldn\'t figure out the context it\'s in')
+}
```
andrey-p: This wasn't completely necessary, but I figured we might as well keep all the insanity contained in the same place. Hence the changes to the UI models.

shared/js/ui/base/safari-ui-wrapper.es6.js
```diff
@@ -1,45 +1,55 @@
...
+        let id = Math.random()
+        message.id = id
+        safari.self.tab.dispatchMessage('getSetting', message)
+
+        safari.self.addEventListener('message', (e) => {
+            if (e.name === 'getSetting' && e.message.id === id) {
+                delete e.message.id
+                resolve(e.message)
+            }
+        }, true)
```
andrey-p: Note, I realized this repeatedly adds event listeners without removing any of them. Probably not a big issue since users will rarely change settings on the options page to the point where it would be a problem. Worth keeping in mind for the future though.

shared/js/background/safari-events.es6.js
jdorweiler: Nice, thanks.

shared/js/background/safari-wrapper.es6.js
jdorweiler: That makes sense now. Nice find.

jdorweiler:
👍

PR #227, opened 2018-05-28
Feedback browser version tweaks

andrey-p:

## Description:

Few tweaks to browser info we add to the feedback form:

- only include the first major browser version
- parse Safari version correctly (we were getting e.g. `601` before which is specifically the Webkit version)
- add some error proofing, just in case
- tests!
- standardize duplicate code across background and UI code

## Steps to test this PR:

1. To avoid spamming the feedback form, add some console.logs for the browser version and name here: https://github.com/duckduckgo/duckduckgo-privacy-extension/blob/0e66e93382de7abea56b9d7729bba52f53d59a39/shared/js/ui/models/feedback-form.es6.js#L13
2. Try opening the feedback form with various browsers - the version/browser name should be good
3. The browser info in the uninstall survey URL should look good too

## Automated tests:
- [ ] Unit tests
- [ ] Integration tests

###### Reviewer Checklist:

###### PR Author Checklist:


andrey-p:
I just noticed there's also UA parsing logic in the background, too. I'll need to bring that logic in line with this one.

shared/js/shared-utils/parse-user-agent-string.es6.js
```diff
@@ -0,0 +1,24 @@
+module.exports = (uaString) => {
+    if (!uaString) uaString = window.navigator.userAgent
+
+    let browser
+    let version
+
+    try {
+        const rgx = uaString.match(/(Firefox|Chrome|Safari)\/([0-9]+)/)
```
MariagraziaAlastra: @andrey-p what do you think about renaming rgx to browserVersionRgx or something like that?

MariagraziaAlastra:
LGTM, thanks 👍

PR #235, opened 2018-06-26
More https upgrades

jdorweiler:
**Reviewer:**

## Description:

Https upgrading from a bloom filter. 

## Steps to test this PR:
1. Open up the extension console and explicitly go to a popular site using http. You should see a line in the console that says `HTTPS: upgrade request url <your site>`.
2. Double check the URL was upgraded and that the popup says "Encrypted Connection".
3. You can test list updates by changing the alarm time and messing with the etags. https://github.com/duckduckgo/duckduckgo-privacy-extension/pull/235/files#diff-3cd94a6f60c9bfd33dffeca9dc0735f8R191
4. You can test corrupt data by clearing indexedDB or by forcing the checksum function to return false. In this case the HTTPSStorage.getLists() function should completely fail. 
https://github.com/duckduckgo/duckduckgo-privacy-extension/pull/235/files#diff-b3d1f859c76ee3ad6e55e5ccd6500bbaR91
5. Problems with HTTPS.setLists() should cause the isReady flag to set to false. You can test this by causing errors in setLists.  https://github.com/duckduckgo/duckduckgo-privacy-extension/pull/235/files#diff-8d433432227571fe75ece3d6c6f8b018R18

## Automated tests:
- [ ] Unit tests
- [ ] Integration tests

###### Reviewer Checklist:

###### PR Author Checklist:


shared/js/background/https.es6.js
```diff
@@ -1,151 +1,60 @@
...
-        return (httpsUpgradeList.indexOf(host) > -1)
-    }
+        if (this.whitelist.indexOf(host) !== -1) {
+            return false
+        }
 
-    getUpgradeList () {
-        return httpsUpgradeList
+        let canUpgrade = false
+        Object.keys(this.upgradeLists).some((listName) => {
```
jdorweiler: We only have one upgrade list for now but this should handle multiple lists when/if we get there.

shared/js/background/storage/https.es6.js
```diff
@@ -0,0 +1,71 @@
...
+                    } else {
+                        // No new data, look up old data from DB
+                        this.getDataFromLocalDB(list.name).then(storedData => {
+                            if (storedData.data) {
+                                list.data = storedData.data
+                            }
+                            resolve(list)
+                        })
+                    }
+                })
```
kdzwinel: `this.getDataXHR` will never resolve if:
- request is unsuccessful or 
- request takes forever.

This means `getLists` never resolves, which means that `HTTPS.setLists` is never called leaving user (unconsciously) unprotected.

https://github.com/duckduckgo/duckduckgo-privacy-extension/blob/efc3ed53117fbb2d8886038be6978ff5b19cce10/shared/js/background/https.es6.js#L55-L56

This is potential vector of the attack - all attacker has to do is to block (or postpone indefinitely) request to `duckduckgo.com/https-bloom.json` or `duckduckgo.com/whitelist.json`.

We should do couple of things:
- load local list as soon as possible, don't wait for update (update only once we get new, valid data)
- have a timeout on https://github.com/duckduckgo/duckduckgo-privacy-extension/blob/efc3ed53117fbb2d8886038be6978ff5b19cce10/shared/js/background/load.es6.js#L31
- don't resolve only on success - https://github.com/duckduckgo/duckduckgo-privacy-extension/blob/efc3ed53117fbb2d8886038be6978ff5b19cce10/shared/js/background/load.es6.js#L50-L51
- add missing `.catch`es
- potentially inform user if we are unable to load local filter/list or download updated list for a long time
- I also think that HTTPS object should not be responsible for updating lists, there should be a separate entity for that. HTTPS should only load local files while other entity regularly updates lists and saves them locally.

andrey-p: I generally agree with all those points. Here's a few notes:

> don't resolve only on success

FWIW I've added some error handling for the `load` module here #238 so we'd be able to tackle different error states after that PR is merged.

> have a timeout

Shouldn't the XHR fail eventually on timeout?

> This is potential vector of the attack - all attacker has to do is to block (or postpone indefinitely) request to `duckduckgo.com/https-bloom.json` or `duckduckgo.com/whitelist.json`.

Can you elaborate on the attack vector piece? We've got HSTS on duckduckgo.com and it's really unlikely that a user will install the extension without having visited DDG immediately before. Once the user downloads the HTTPS files for the first time, they're covered.

> I also think that HTTPS object should not be responsible for updating lists, there should be a separate entity for that. HTTPS should only load local files while other entity regularly updates lists and saves them locally.

This *is* the separate entity, we're separating storage from business logic in this PR :)

kdzwinel: > FWIW I've added some error handling for the load module here #238 so we'd be able to tackle different error states after that PR is merged.

Awesome 👏 That almost fixes the `request is unsuccessful` issue I mentioned, we just need to add missing `.catch`es to this file.

> Shouldn't the XHR fail eventually on timeout?

Unfortunately, not. By default there is no timeout. https://stackoverflow.com/questions/9502410/ajax-xmlhttprequest-timeout-length-by-browser#9503384

> Once the user downloads the HTTPS files for the first time, they're covered.

Blocking `duckduckgo.com/https-bloom.json` will cause HTTPS object to never enter the `isReady` state (as far as I can see), which completely disables its functionality. Even though the (older versions) of those lists are available locally.

andrey-p: > Unfortunately, not. By default there is no timeout.

Huh, TIL. Yeah, let's add a timeout.

> Blocking `duckduckgo.com/https-bloom.json` will cause HTTPS object to never enter the `isReady` state (as far as I can see), which completely disables its functionality. Even though the (older versions) of those lists are available locally.

Got it. We should handle this here, yeah.

@jdorweiler this actually works pretty well with the thing I was unsure about in the other PR, where [the load promise gets rejected if the server returns 304](https://github.com/duckduckgo/duckduckgo-privacy-extension/pull/238/files#r199819195). In case of any kind of rejected promise from the `load` module, we can fall back to the DB.

shared/js/background/https.es6.js
```diff
@@ -1,151 +1,60 @@
...
+            canUpgrade = this.upgradeLists[listName].checkEntry(host)
+            return canUpgrade === true
+        })
+
+        return canUpgrade
     }
 
     getUpgradedUrl (reqUrl, tab, isMainFrame) {
+        if (!this.isReady) return reqUrl
```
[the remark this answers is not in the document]
jdorweiler: I'll add a warn. This state isn't too uncommon though. For example, opening the browser with saved tabs could make some requests before https is ready.

shared/js/background/https.es6.js
```diff
@@ -1,151 +1,92 @@
...
+                    this.createBloomFilter(list.data).then(bloom => { 
+                        resolve(this.upgradeLists.set(list.name, bloom))
+                    }).catch((e) => {
+                        reject(e)
+                    })
+                } else if (list.type === 'whitelist') {
+                    resolve(this.whitelist = list.data)
+                }
+            })
+        })).then(() => {
```
jdorweiler: Only set https.isReady if all the promises resolve

shared/js/background/https.es6.js
```diff
@@ -1,151 +1,92 @@
...
+            this.isReady = true
+            console.log('HTTPS: is ready')
+        }).catch((e) => {
```
jdorweiler: If any promises fail then set https.isReady to false.

shared/js/background/storage/https.es6.js
```diff
@@ -0,0 +1,61 @@
...
+            })
+        }))
+    }
+
+    getDataXHR (url) {
+        return new Promise((resolve, reject) => {
+            load.JSONfromExternalFile(url, resolve)
+            // if load fails, resolve and try to load from db instead
+            setTimeout(() => resolve(), 30000)
```
jdorweiler: Wait 30sec in cases where an xhr fails. The promise resolves without any data and falls back to the database here https://github.com/duckduckgo/duckduckgo-privacy-extension/pull/235/files#diff-b3d1f859c76ee3ad6e55e5ccd6500bbaR28

shared/js/background/abp-lists.es6.js
```diff
@@ -75,16 +75,18 @@ function updateLists () {
             if (Object.keys(list.parsed).length === 0) etag = ''
 
             load.loadExtensionFile({url: url, source: 'external', etag: etag}).then((response) => {
-                const listData = response.data
-                const newEtag = response.getResponseHeader('etag') || ''
-                console.log('Updating list: ', name)
+                if (response && response.status === 200) {
```
jdorweiler: I updated the load function to return the response for 200 and 304.

shared/js/background/load.es6.js
```diff
@@ -80,7 +80,8 @@ function loadExtensionFile (params) {
                     xhr.data = returnResponse(xhr, params.returnType)
                     resolve(xhr)
                 } else if (xhr.status === 304) {
-                    reject(new Error(`${url} returned 304, resource not changed`))
+                    console.log(`${url} returned 304, resource not changed`)
```
jdorweiler: I updated this to a log instead of error and I'm returning an empty object with just a status code. I think it makes more sense to handle the 304 case explicitly.

shared/js/background/https.es6.js
```diff
@@ -1,152 +1,72 @@
...
-        })
     }
 
-    getFromStorage (fn) {
-        return new Promise((resolve, reject) => {
-            // For Chrome/Firefox:
-            if (window.chrome) {
-                console.log('HTTPS: getFromStorage() using chrome.storage.local (Chrome/FF)')
-
-                chrome.storage.local.get('https-upgrade-list', (results) => {
```
andrey-p: I've been thinking about the possible upgrade path for users. I wonder whether we should delete any previous data that users on the current extension version might have.

Maybe we can add a check in `chrome-events`/`upgrade` reason: https://github.com/duckduckgo/duckduckgo-privacy-extension/blob/b4856e22c809cf2dcad583f32cfa3438b5d835e8/shared/js/background/chrome-events.es6.js#L9-L16

... check if `https-upgrade-list` is in local storage, and delete it if yes.

Not sure if this is a big deal though.

andrey-p:
All right, LGTM. Let's merge!

PR #238, opened 2018-06-29
Tighten ATB, add unit tests

andrey-p:
**Reviewer:** @jdorweiler 
**CC:** @jkv 

## Description:
- Tightens up ATB pipeline steps
- Adds `dev=1` to all requests made from dev builds of the extension
- Adds the ability to handle errors from the load module
- ... and adds unit test coverage as for the ATB module as much as possible.

See inline comments for more info.

## Steps to test this PR:

Test the following two scenarios on Chrome, FF and Safari. Note, Safari may need to be put on DDH6 for us to test the entire flow.

**Case 1: no post-install success page (e.g. if the user installed via the Chrome web store)**

1. Install the extension with no DDG tabs open
2. The following requests should be made:
    - `atb.js`
    - `/exti?atb=v124-2`
3. Run a search. The following request should be made:
    - `atb.js?atb=v124-2&set_atb=v124-2`

**Case 2: post-install success page**

1. Open a browser tab for https://duckduckgo.com/?exti=2
2. Continue with all the steps in case 2.
3. The same URLs should be hit, only the variants should be added (e.g. `exti?atb=v124-2__` or `exti?atb=v124-2ab`

## Automated tests:
- [ ] Unit tests
- [ ] Integration tests

###### Reviewer Checklist:

###### PR Author Checklist:

### DIFF
--- a/shared/js/background/atb.es6.js
+++ b/shared/js/background/atb.es6.js
@@ -1,6 +1,7 @@
 const settings = require('./settings.es6')
 const parseUserAgentString = require('../shared-utils/parse-user-agent-string.es6')
 const load = require('./load.es6')
+const browserWrapper = require('./$BROWSER-wrapper.es6')
 
 var ATB = (() => {
     // regex to match ddg urls to add atb params to.
@@ -76,22 +77,15 @@ var ATB = (() => {
         },
 
         inject: () => {
-            // skip this for non webextension browsers
-            if (!window.chrome) return
-
-            chrome.tabs.query({ url: 'https://*.duckduckgo.com/*' }, function (tabs) {
+            browserWrapper.getTabsByURL('https://*.duckduckgo.com/*', (tabs) => {
                 let i = tabs.length
                 let tab
+
                 while (i--) {
                     tab = tabs[i]
 
-                    chrome.tabs.executeScript(tab.id, {
-                        file: '/public/js/content-scripts/on-install.js'
-                    })
-
-                    chrome.tabs.insertCSS(tab.id, {
-                        file: '/public/css/noatb.css'
-                    })
+                    browserWrapper.executeScript(tab.id, '/public/js/content-scripts/on-install.js')
+                    browserWrapper.insertCSS(tab.id, '/public/css/noatb.css')
                 }
             })
         },
@@ -157,8 +151,7 @@ var ATB = (() => {
             let browserInfo = parseUserAgentString()
             let browserName = browserInfo.browser
             let browserVersion = browserInfo.version
-            let extensionVersion = window.chrome.runtime.getManifest().version
-
+            let extensionVersion = browserWrapper.getExtensionVersion()
             if (browserName) url += `&browser=${browserName}`
             if (browserVersion) url += `&bv=${browserVersion}`
             if (extensionVersion) url += `&v=${extensionVersion}`
@@ -172,10 +165,8 @@ settings.ready().then(() => {
     // migrate over any localStorage values from the old extension
     ATB.migrate()
 
-    // set initial uninstall url. webextension only
-    if (window.chrome) {
-        chrome.runtime.setUninstallURL(ATB.getSurveyURL())
-    }
+    // set initial uninstall url
+    browserWrapper.setUninstallURL(ATB.getSurveyURL())
 })
 
 module.exports = ATB
--- a/shared/js/background/atb.es6.js
+++ b/shared/js/background/atb.es6.js
@@ -70,9 +70,9 @@ var ATB = (() => {
             settings.updateSetting('atb', atb)
             settings.updateSetting('set_atb', atb)
 
-            let xhr = new XMLHttpRequest()
-            xhr.open('GET', 'https://duckduckgo.com/exti/?atb=' + atb, true)
-            xhr.send()
+            load.JSONfromExternalFile(`https://duckduckgo.com/exti/?atb=${atb}`, () => {
+                // no-op, we only care that the request was made
+            })
         },
 
         inject: () => {
--- a/shared/js/background/atb.es6.js
+++ b/shared/js/background/atb.es6.js
@@ -88,7 +88,7 @@ var ATB = (() => {
 
             settings.updateSetting('set_atb', atb)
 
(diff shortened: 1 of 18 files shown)


unit-test/background/atb.es6.js
```diff
@@ -1,33 +1,316 @@
...
+
+        setTimeout(() => {
+            validateExtiWasHit('v112-2')
+            expect(settings.getSetting('atb')).toEqual('v112-2')
+            expect(settings.getSetting('set_atb')).toEqual('v112-2')
+
+            done()
+        }, 600)
+    })
+})
```
andrey-p: I'm not too happy about these timeout-based unit tests. I tried to use a way to mock these (I've used `timekeeper` in other tests) but I couldn't figure it out.

shared/js/background/atb.es6.js
```diff
@@ -60,87 +46,84 @@ var ATB = (() => {
...
             // wait until settings is ready to try and get atb from the page
-            settings.ready().then(() => {
-                ATB.inject()
-                ATB.migrate()
-                ATB.setInitialVersions()
-            })
+            return settings.ready()
+                .then(ATB.setInitialVersions)
+                .then(() => {
+                    ATB.migrate()
```
andrey-p: delete this

shared/js/background/load.es6.js
```diff
@@ -1,20 +1,31 @@
...
-        loadExtensionFile({url: url, returnType: 'json', source: 'external'}, (res, xhr) => cb(JSON.parse(res), xhr))
-    } catch (e) {
-        console.log(e)
-        return {}
-    }
+function JSONfromExternalFile (url) {
+    return loadExtensionFile({url: url, returnType: 'json', source: 'external'})
+}
+
+function url (url, cb) {
```
kdzwinel: `cb` is not called 🤔

shared/js/background/load.es6.js
```diff
@@ -27,35 +38,60 @@ function returnResponse (xhr, returnType) {
...
-        xhr.open('GET', params.url)
+        if (dev) {
+            if (url.indexOf('?') > -1) {
+                url += '&'
+            } else {
+                url += '?'
+            }
+
+            url += 'dev=1'
+        }
```
kdzwinel: Since we are using promises and started doing search parameters manipulation, IMO it would simplify things quite a bit to switch to `fetch` and `URLSearchParams` from `XMLHttpRequest`.

shared/js/background/load.es6.js
andrey-p: A previous iteration of this still used callbacks. Lemme remove.

shared/js/background/load.es6.js
```diff
@@ -27,35 +38,60 @@ function returnResponse (xhr, returnType) {
...
+            let done = XMLHttpRequest.DONE ? XMLHttpRequest.DONE : 4
+            if (xhr.readyState === done) {
+                if (xhr.status === 200 || (xhr.type && xhr.type === 'internal')) {
+                    xhr.data = returnResponse(xhr, params.returnType)
+                    resolve(xhr)
+                } else if (xhr.status === 304) {
+                    reject(new Error(`server returned 304, resource not changed`))
+                } else {
+                    reject(new Error(`couldn't reach ${url}, got status: ${xhr.status}`))
+                }
```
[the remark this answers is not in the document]
jdorweiler: 304 is not really an error but we also don't want to do anything with it so rejecting is fine. Could you add `xhr.responseURL` to the message though.

shared/js/background/atb.es6.js
```diff
@@ -60,78 +39,77 @@ var ATB = (() => {
...
+            if (!atb || settings.getSetting('extiSent')) return
+
+            settings.updateSetting('extiSent', true)
             settings.updateSetting('set_atb', atb)
 
-            let xhr = new XMLHttpRequest()
-            xhr.open('GET', 'https://duckduckgo.com/exti/?atb=' + atb, true)
-            xhr.send()
+            // just a GET request, we only care that the request was made
+            load.url(`https://duckduckgo.com/exti/?atb=${atb}`)
```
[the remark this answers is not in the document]
jdorweiler: Is it possible for this to run multiple times if the user clears their extension storage?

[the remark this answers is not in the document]
andrey-p: Actually, yes - this can possibly run on `update` as well as `install`: https://github.com/duckduckgo/duckduckgo-privacy-extension/blob/cc6c045925ee4bd962b3040e0e7e0dabe94cae90/shared/js/background/chrome-events.es6.js#L11-L13

Do you know the original reason for having this run on `update` too? Do we want to change it just runs on `install`?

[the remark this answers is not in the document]
jdorweiler: `update` was to handle users upgrading from legacy extensions to webextensions. Can probably get rid of that now.

shared/js/background/load.es6.js
```diff
@@ -27,35 +38,65 @@ function returnResponse (xhr, returnType) {
...
 
-    xhr.onreadystatechange = function () {
-        let done = XMLHttpRequest.DONE ? XMLHttpRequest.DONE : 4
-        if (xhr.readyState === done) {
-            if (xhr.status === 200 || (xhr.type && xhr.type === 'internal')) {
-                cb(returnResponse(xhr, params.returnType), xhr)
+    return new Promise((resolve, reject) => {
+        xhr.ontimeout = () => {
+            reject(new Error(`${url} timed out`))
+        }
```
andrey-p: @jdorweiler @kdzwinel I've added the timeout protection here, since ATB can possibly also be affected. I've set a 20s timeout limit, which seems reasonable (e.g. we don't want to wait too long before loading HTTPS rules from the DB)

That said, it seems from some of the answers in https://stackoverflow.com/questions/9502410/ajax-xmlhttprequest-timeout-length-by-browser#9503384 that browsers vary a bit in how they handle XHR timeouts, so hopefully this'll work.

jdorweiler: Could we do a default of 20 but also pass a value to increase it if we need to?

jdorweiler:
👍

PR #243, opened 2018-07-24
HTTPS: don't strip out www from domain

MariagraziaAlastra:
**Reviewer:** @jdorweiler 

**CC:** @zachthompson 

## Description:
We currently strip out the "www" from domains when we check if they're in the list or not. 
Since some domains in the list contain the "www" part, we should not do that.

## Steps to test this PR:

1. Build and reload extension
2. Try loading some of the  domains with "www" in the list (eg "mozdev.org", "nowmentalhealth.com", "bosch-home.com.au", "banche.roma.it") 
3. Open the popup. Expected: the https line should say "Encrypted Connection" with a green lock icon
4. Also try some sites without the "www"
5. Repeat step 3, with same expected result

I've also added some unit tests for utils.es6.js

## Automated tests:
- [x] Unit tests
- [ ] Integration tests

###### Reviewer Checklist:

###### PR Author Checklist:


shared/js/background/https.es6.js
```diff
@@ -85,8 +85,8 @@ class HTTPS {
             return reqUrl
         }
 
-        // Determine host
-        const host = utils.extractHostFromURL(reqUrl)
+        // Determine host without stripping 'www',
+        const host = utils.extractHostFromURL(reqUrl, true) || ''
 
         if (this.canUpgradeHost(host)) {
```
jdorweiler: Can you check that `host` exists here? Just in case line 89 sets the host to '' we get a bad bloom filter that decides '' is upgradable.

MariagraziaAlastra: @jdorweiler done, thanks

jdorweiler:
👍

PR #244, opened 2018-07-25
Move whitelist over to trackers file

jdorweiler:
**Reviewer:**

## Description:

A script that generates a regex entry from an adblock formatted filter.

## Automated tests:
- [ ] Unit tests
- [ ] Integration tests

###### Reviewer Checklist:

###### PR Author Checklist:

### DIFF
--- a/shared/data/tracker_lists/trackersWithParentCompany.json
+++ b/shared/data/tracker_lists/trackersWithParentCompany.json
@@ -698,6 +698,17 @@
                     }
                 },
                 {
+                    "rule": "ad4game\\.com[?/].*\\/images\\/",
+                    "options": {
+                        "types": [
+                            "image"
+                        ],
+                        "domains": [
+                            "kissanime.com"
+                        ]
+                    }
+                },
+                {
                     "rule": "ad4game\\.com\\/js\\/",
                     "options": {
                         "types": [
@@ -10914,7 +10925,7 @@
             "u": "http://www.ibm.com/",
             "whitelist": [
                 {
-                    "rule": "coremetrics\\.com.*\\/eluminate\\.js"
+                    "rule": "coremetrics\\.com[?/].*\\/eluminate\\.js"
                 }
             ]
         },
@@ -12081,7 +12092,7 @@
                     "rule": "pixel\\.facebook\\.com($|[?/])"
                 },
                 {
-                    "rule": "facebook\\.com.*\\/impression\\.php"
+                    "rule": "facebook\\.com\\/.*\\/impression\\.php"
                 },
                 {
                     "rule": "graph\\.facebook\\.com\\/\\?ids=.*&callback=.*",


shared/js/background/trackers.es6.js
```diff
@@ -220,7 +195,20 @@ function checkTrackersWithParentCompany (url, siteDomain, request) {
...
     if (toBlock) {
-        return toBlock
+        // we have a blocked tracker but need to see if it should be whitelisted
+        if (toBlock.trackerData && toBlock.trackerData.whitelist) {
+            // check whitelist
+            toBlock.trackerData.whitelist.some(ruleObj => {
+                if (requestMatchesRule(request, ruleObj.rule) && matchRuleOptions(ruleObj, request, siteDomain)) {
+                    toBlock = null
+                }
+            })
```
kdzwinel: this looks suspicious 🤔 we are modifying `toBlock` inside of the loop which is going over data from `toBlock`?

BTW this will go through the whole list every time (even if whitelisted entry is first). we can avoid that by doing:

```js
const foundOnWhitelist = toBlock.trackerData.whitelist.some(ruleObj => {
    if (requestMatchesRule(request, ruleObj.rule) && matchRuleOptions(ruleObj, request, siteDomain)) {
        return true;
    }
    return false;
})
```

shared/js/background/trackers.es6.js
```diff
@@ -198,6 +170,8 @@ function checkTrackersWithParentCompany (url, siteDomain, request) {
             tracker.rules.some(ruleObj => {
                 if (requestMatchesRule(request, ruleObj.rule) && matchRuleOptions(ruleObj, request, siteDomain)) {
                     toBlock.rule = ruleObj
+                    toBlock.trackerData = tracker
```
kdzwinel: Can we immediately check the whitelist here and bail out if request is whitelisted? Wouldn't it save us some work?

test/tests/trackers.js
```diff
@@ -197,4 +173,18 @@
       });
   });
 
+  QUnit.test("Whitelists rules", (assert) => {
+      bkg.settings.updateSetting('trackerBlockingEnabled', true);
+      
+      let testTab = Object.assign({}, JSON.parse(JSON.stringify(fakeTab)))
```
kdzwinel: why do we need `Object.assign` here? `JSON.parse(JSON.stringify` should be enough to create a deep copy of an object

shared/js/background/trackers.es6.js
jdorweiler: 👍 good idea.

shared/js/background/trackers.es6.js
jdorweiler: Me too. I spent some time reorganizing the whole algorithm today. Still needs some more work but it's looking better.

shared/js/background/surrogates.es6.js
```diff
@@ -66,7 +66,11 @@ class Surrogates {
         filename = filename.split('?')[0]
         // concat with domain to match the original rule:
         let ruleToMatch = parsedUrl.domain + '/' + filename
-        return surrogateList[ruleToMatch]
+
+        let surrogate = surrogateList[ruleToMatch]
+
+        if (surrogate) console.log(`Serving surrogate for ${url}`)
+        return surrogate
```
jdorweiler: I moved this log message out of trackers.js

shared/js/background/trackers.es6.js
```diff
@@ -271,41 +245,7 @@ function getCommonParentEntity (currLocation, urlToCheck) {
...
-    const urlToCheckParsed = tldjs.parse(urlToCheck)
-    const parentEntity = entityMap[urlToCheckParsed.domain]
-    if (parentEntity) {
-        return parentEntity
-    } else {
-        return 'unknown'
-    }
-}
-
-function getTrackerDetails (trackerUrl, listName) {
```
jdorweiler: Replaced with `getReturnTrackerObj`

shared/js/background/trackers.es6.js
```diff
@@ -271,41 +245,7 @@ function getCommonParentEntity (currLocation, urlToCheck) {
...
-            domain: siteDomain,
-            elementTypeMask: abp.elementTypes[request.type.toUpperCase()]
-        })
-    return match
-}
-
 module.exports = {
     isTracker: isTracker,
-    loadLists: loadLists,
-    getParentEntity: getParentEntity
```
jdorweiler: Moved this to `utils`

shared/js/background/trackers.es6.js
```diff
@@ -230,8 +157,55 @@ function checkTrackersWithParentCompany (url, siteDomain, request) {
...
+}
+
+/* Check the matched rule  options against the request data
+ * return: true (all options matched)
+ */
+function matchRuleOptions (rule, request, siteDomain) {
+    if (!rule.options) return true
+
+    if (rule.options.types) {
+        let matchesType = rule.options.types.findIndex(t => { return t === request.type })
```
kdzwinel: `return rule.options.types.includes(request.type);`

shared/js/background/trackers.es6.js
```diff
@@ -230,8 +157,55 @@ function checkTrackersWithParentCompany (url, siteDomain, request) {
...
+        if (matchesType === -1) {
+            return false
+        }
+    }
+
+    if (rule.options.domains) {
+        let matchesDomain = rule.options.domains.findIndex(d => { return d === siteDomain })
```
kdzwinel: ditto

scripts/filterToRule.js
```diff
@@ -0,0 +1,246 @@
...
+
+
+(() => {
+    let filters = fs.readFileSync(program.file).toString().split('\n')
+
+    // process filters, translate to rules, merge duplicates
+    filters.map(f => {
+        if (!f) return 
+
+        let rule = parseFilter(f.toLowerCase())
```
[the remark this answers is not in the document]
jdorweiler: I think it will be okay. Other tracker blockers ignore case also.

scripts/filterToRule.js
```diff
@@ -0,0 +1,246 @@
...
+    let options = {}
+
+    // split option string on commas
+    let optionList = optionStr.split(',')
+
+
+    optionList.map(o => {
+        // skip third party option. All of our rules
+        // are by default third party
+        if (!o || o.match(/third-party|first-party/)) {
```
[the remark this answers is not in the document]
jdorweiler: I should update this a little. 
1. We never block first-party so those can be completely skipped.
2. Anything we do block is implied third-party. we can just ignore the option but still process the rule.

shared/js/background/trackers.es6.js
```diff
@@ -18,83 +15,71 @@ function loadLists () {
...
-        if (surrogateTracker) {
-            let commonParent = getCommonParentEntity(currLocation, urlToCheck)
-            if (commonParent) {
-                return addCommonParent(surrogateTracker, commonParent)
-            }
-            return surrogateTracker
+    if (parsedUrl && parsedUrl.hostname) {
+        hostname = parsedUrl.hostname
+    } else {
+        // fail gracefully if tldjs chokes on the URL e.g. it doesn't parse
```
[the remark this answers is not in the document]
jdorweiler: 👍  I'd be happy to see it go. I'll make it a follow up to look into it.

shared/js/background/trackers.es6.js
```diff
@@ -179,48 +113,41 @@ function checkTrackersWithParentCompany (url, siteDomain, request) {
...
-            parentCompany: tracker.c,
-            url: utils.extractHostFromURL(request.url),
-            type: trackerType,
-            block: true,
-            rule: '',
-            reason: 'trackersWithParentCompany'
-        }
-
         // Check to see if this request matches any of the blocking rules for this tracker
         if (tracker.rules) {
```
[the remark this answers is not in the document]
jdorweiler: Shouldn't happen but I'll check for a length just in case.

scripts/filterToRule.js
```diff
@@ -0,0 +1,246 @@
...
+    })
+
+    const rulesByHost = groupRulesByHost(rules)
+
+    writeRules(rules)
+
+    combineWithTrackers(rulesByHost) 
+
+    // run tests on known input-output
+    Object.keys(tests).map(f => {
```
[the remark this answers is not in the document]
jdorweiler: I moved them to a separate function and run by passing a `--test` option to the script.

jdorweiler:
@kdzwinel thanks for the review. It's ready to go again.

scripts/filterToRule.js
```diff
@@ -0,0 +1,246 @@
...
+ * node filterToRule.js -f <filterlist> -o <outputfile> -c <path to tracker file> -t <rule type: 'rule' or 'whitelist'>
+ */
+
+const program = require('commander')
+const fs = require('fs')
+const merge = require('deepmerge')
+const _ = require('underscore')
+const assert = require('assert')
+const utils = require('../shared/js/background/utils.es6')
+const RandExp = require('randexp')
```
kdzwinel: After `randexp` was plugged in `filterToRule` script started failing randomly because some of the generated URLs are invalid. Examples:

- http://coremetrics.com+cO|6;!B-'LM(z(u7Zq=1=piDNa=r_(=J$/eluminate.js
- http://ad4game.comZ1N*@+S/vS|LnnK~f/:W?=Z)fo&|E]h>s/images/

Both of those work when we add a `/` after `.com`. Also both of those end up in `unMatchedRules.json`. This might be suggesting some underlying issue with the generated regexes (or the input data?).

scripts/filterToRule.js
```diff
@@ -0,0 +1,246 @@
...
+
+    // *^ pattern to [?/].*
+    filter = filter.replace(/\*\^/g, '[?/].*')
+    
+    // ^* pattern to [?/].*
+    filter = filter.replace(/\^\*/g, '[?/].*')
+
+    // single wild card * to .*
+    // don't match on [?/].* pattern from above
+    filter = filter.replace(/(?<!\[\?\/\]\.)\*/g, '.*')
```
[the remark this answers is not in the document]
kdzwinel: Oh, this is a "lookbehind assertion". Fancy! If we are OK with only supporting node 10 then disregard my previous comment.

scripts/filterToRule.js
jdorweiler: Nice catch. Looks like the input filters are bad. ` ||ad4game.com*/images/` `||coremetrics.com*/eluminate.js`. I'll fix them and rerun the script.

jdorweiler:
@kdzwinel ready to go again.

PR #245, opened 2018-07-25
Refactor element hiding content script to be an es6 class

dharb:
**Reviewer:** @kdzwinel 

## Description:
No functionality change here, just refactoring to make the element hiding content script an ES6 class.

## Steps to test this PR:
1. Make sure element hiding works exactly as it did before. Note that not all of the ads on the site we've been testing this on (yahoo.com) will be collapsed since they don't all contain blocked tracking requests.

### DIFF
--- a/shared/js/content-scripts/element-hiding.js
+++ b/shared/js/content-scripts/element-hiding.js
@@ -15,48 +15,41 @@
             this.foundScripts = []
             this.foundFrames = []
             this.frameListener = this.frameListener.bind(this)
+            this.messageListener = this.messageListener.bind(this)
+
             // Initialize event listener for interframe messaging
             window.addEventListener('message', this.frameListener)
+            // Initialize event listener for background page messaging
+            chrome.runtime.onMessage.addListener(this.messageListener)
+        }
 
-            /**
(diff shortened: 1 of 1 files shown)


shared/js/content-scripts/element-hiding.js
```diff
@@ -7,124 +7,131 @@
...
+                this.hideFrame(frame.parentNode)
             }
-        })
-    } else {
-        chrome.runtime.sendMessage({hideElements: true, url: document.location.href, frame: contentScript.frameType})
+        }
     }
+
+    // Instantiate content script
+    var contentScript = new ContentScript()
```
dharb: eslint is throwing a no-unused-var error here - https://travis-ci.org/duckduckgo/duckduckgo-privacy-extension/builds/408297396.

andrey-p:
Just a note, I don't know if content scripts are ever transpiled. Is that something we're worried about?

andrey-p:
@kdzwinel Cool, thanks. Just don't want to get caught off guard by something on some slightly older browser that people are somehow still using.

dharb:
Thanks @kdzwinel! That makes perfect sense to me, good call with those changes.

PR #282, opened 2018-08-23
Add anonymous engagement metrics

MariagraziaAlastra:
**Reviewer:** @andrey-p 

## Description:
Setting up anonymous engagement metrics for the extension.
Learn more here: https://duck.co/help/privacy/atb

## Steps to test this PR:
1. Build and reload extension
2. Navigate through sites, open the popup, look at the subviews etc
3. Inspect the Network requests in the background, make sure the right metrics are sent as per specs
4. There are also new unit tests

## Automated tests:
- [ ] Unit tests
- [ ] Integration tests

###### Reviewer Checklist:

###### PR Author Checklist:

### DIFF
--- a/shared/js/background/pixel.es6.js
+++ b/shared/js/background/pixel.es6.js
@@ -98,9 +98,7 @@ function concatParams (args) {
         }
     })
 
-    if (paramString.length) {
-        resultString = `${paramString}?${randomNum}${objParamString}`
-    }
+    resultString = `${paramString}?${randomNum}${objParamString}`
 
     return resultString
 }
--- a/shared/js/background/pixel.es6.js
+++ b/shared/js/background/pixel.es6.js
@@ -32,6 +32,7 @@ function fire () {
     if (!url) return
 
     args = args.slice(1)
+    args.push(getAdditionalParams())
     const paramString = concatParams(args) || ''
 
     // Send the request
@@ -77,8 +78,6 @@ function getAdditionalParams () {
 function concatParams (args) {
     args = args || []
 
-    args.push(getAdditionalParams())
-
     let paramString = ''
 
     args.forEach((arg) => {
@@ -100,6 +99,5 @@ function concatParams (args) {
 module.exports = {
     fire: fire,
     getURL: getURL,
-    getAdditionalParams: getAdditionalParams,
     concatParams: concatParams
 }
--- a/shared/js/background/pixel.es6.js
+++ b/shared/js/background/pixel.es6.js
@@ -21,12 +21,16 @@ const parseUserAgentString = require('../shared-utils/parse-user-agent-string.es
  */
 
 function fire (pixelName) {
+    if (!arguments.length) return
+
     let args = Array.prototype.slice.call(arguments)
+    const pixelName = args[0]
+    const paramString = concatParams(args)
 }
 
 /**
  *
- * returns URL for the pixel request
+ * Return URL for the pixel request
  *
  */
 function getURL () {
@@ -36,6 +40,11 @@ function getURL () {
     return domain + path
 }
 
+/**
+ *
+ * Return additional params for the pixel request
+ *
+ */
 function getAdditionalParams () {
     const browserInfo = parseUserAgentString()
 
@@ -44,6 +53,17 @@ function getAdditionalParams () {
         extensionVersion: browserWrapper.getExtensionVersion(),
         atb: settings.getSetting('atb')
     }
+}
+
+/**
+ *
+ * @param {array} args - data we need to append
+ *
+ */
+function concatParams (args) {
+    args = args ? args : []
+    
+    args.push(getAdditionalParams())
 }
 
 module.exports = {
--- a/shared/js/background/pixel.es6.js
+++ b/shared/js/background/pixel.es6.js
@@ -19,25 +19,20 @@ const parseUserAgentString = require('../shared-utils/parse-user-agent-string.es
  *
  */
 function fire () {
-    console.log('firing...')
-    console.log(arguments)
-    console.log(typeof arguments)
     if (!arguments.length) return
 
     let args = Array.prototype.slice.call(arguments)
     const pixelName = args[0]
-    console.log('firing ' + pixelName) 
 
     if (typeof pixelName !== 'string') return
 
     const url = getURL(pixelName)
-    console.log('firing url: ' + url) 
 
     if (!url) return
 
-    args = Array.prototype.slice.call(args, 1)
+    args = args.slice(1)
     args.push(getAdditionalParams())
-    const paramString = concatParams(args) || ''
+    const paramString = concatParams(args)
 
     // Send the request
     load.url(url + paramString)
@@ -87,7 +82,7 @@ function concatParams (args) {
     args.forEach((arg) => {
         // append keys if object
         if (typeof arg === 'object') {
-            paramString = Object.keys(arg).reduce((params, key) => {
+            paramString += Object.keys(arg).reduce((params, key) => {
                 const val = arg[key]
                 if (val) return params + '&' + key + '=' + val
             }, '')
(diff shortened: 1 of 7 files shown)


shared/js/background/pixel.es6.js
```diff
@@ -0,0 +1,102 @@
...
+        // append keys if object
+        if (typeof arg === 'object') {
+            paramString += Object.keys(arg).reduce((params, key) => {
+                const val = arg[key]
+                if (val) return params + '&' + key + '=' + val
+            }, '')
+        } else if (arg) {
+            // otherwise just add args separated by _
+            paramString += '_' + arg
+        }
```
andrey-p: There's a couple of things I spotted here:

1. This produces malformed URLs, e.g. `/t/ep&foo=bar&bar=baz`. The querystring param needs to be separated with a `?`, so something like: `/t/ep?foo=bar&bar=baz`.

2. This is arguably an edge case, but if you pass an object _before_ a string argument, like this:

```js
pixel.fire('ep', { foo: 'bar' }, 'baz')
```

you'll end up with something like:

```
/t/ep_&foo=bar_baz
```

whereas our internal pixel class would produce something like:

```
/t/ep_baz?foo=bar
```

unit-test/background/pixel.es6.js
```diff
@@ -0,0 +1,60 @@
...
+            'param1'
+        ],
+        'partialResult': '_param1'
+    },
+    {
+        'params': [
+            'param1',
+            'param2'
+        ],
+        'partialResult': '_param1_param2'
```
andrey-p: Can we add more test cases here? I'd like to see test cases for the two things from my comment above, as well as a test case for multiple object params like:

```js
[{ foo: 'bar' }, { bar: 'baz' }]
```

shared/js/background/pixel.es6.js
```diff
@@ -0,0 +1,102 @@
...
+/**
+ *
+ * Fire a pixel
+ *
+ * @param {string} pixelName
+ * @param {...*} args - any number of extra data
+ *
+ */
+function fire () {
+    if (!arguments.length) return
```
andrey-p: We're making the assumption that `settings` is ready by the time this function is called. It's extremely unlikely that we'd call it _immediately_ on extension start, though, so it should be fine.

I don't think there's anything actionable here, just wanted to point it out.

unit-test/background/pixel.es6.js
```diff
@@ -0,0 +1,60 @@
...
+})
+
+describe('pixel.concatParams()', () => {
+    concatParamsTestCases.forEach((test) => {
+        it(`should return a string containing ${test.partialResult} and for params: ${test.params}`, () => {
+            let result = pixel.concatParams(test.params)
+            expect(result).toMatch(test.partialResult)
+        })
+    })
+})
```
andrey-p: Can we add a specific test that covers the entirety of `fire`? You'd need to set up spies for `load.url()` and `settings.getSetting()`. I made a handy reusable spy for `settings` - ping me when you're about if you want me to show you it!

shared/js/background/pixel.es6.js
```diff
@@ -0,0 +1,102 @@
...
+ *
+ */
+function getAdditionalParams () {
+    const browserInfo = parseUserAgentString()
+    const browser = browserInfo.browser
+    const extensionVersion = browserWrapper.getExtensionVersion()
+    const atb = settings.getSetting('atb')
+    const result = {}
+
+    if (browser) result.browser = browser
```
andrey-p: Can we add browser as a positional argument rather than a query string one? So something like:

```
/t/ep_chrome?foo=bar
/t/ept_off_firefox?foo=bar
```

shared/js/background/pixel.es6.js
andrey-p: Also, I really don't trust browser caching, can we add a `Math.random()` same way we do in ATB?

https://github.com/duckduckgo/duckduckgo-privacy-extension/blob/fa8b37dde86812c7343e92256a961b732ff8f129/shared/js/background/atb.es6.js#L37-L38

unit-test/background/pixel.es6.js
andrey-p: Note, if you add a `Math.random()` as per my earlier comment above you'll probably need to add something like:

```js
spyOn(Math, 'random').and.returnValue(1)
```

so you don't need to worry about having to match the random part with a regex or something.

unit-test/background/pixel.es6.js
MariagraziaAlastra: Hey @andrey-p, I added other test cases for the existing unit tests. I don't think it's worth adding tests for fire()? All the data used to send the URL is returned from other functions that we're unit testing already. I mean, besides actually loading the URL, what do you think should be tested here?
I wouldn't hold the PR for it, but let me know if you disagree :)

unit-test/background/pixel.es6.js
```diff
@@ -0,0 +1,93 @@
...
+            },
+            'param1',
+            {
+                'p5': 'param5',
+                'p6': 'param6'
+            },
+            'param2'
+        ],
+        'result': '_param1_param2?1000000&p3=param3&p4=param4&p5=param5&p6=param6'
+    }
```
andrey-p: This is awesome, thank you!

unit-test/background/pixel.es6.js
```diff
@@ -0,0 +1,93 @@
...
+    {
+        'pixelName': undefined,
+        'result': undefined
+    }
+]
+const concatParamsTestCases = [
+    {
+        'params': [],
+        'result': ''
+    },
```
andrey-p: Is this the behaviour we want? I'd have expected no params to just produce `?100000`

shared/js/ui/models/site.es6.js
```diff
@@ -66,7 +66,7 @@ Site.prototype = window.$.extend({},
 
         setSiteProperties: function () {
             if (!this.tab) {
-                this.domain = 'new tab' // tab can be null for firefox new tabs
+                this.domain = 'new tab' // tab can be null for firePixelfox new tabs
```
andrey-p: I think find+replace went a bit too far here :)

shared/js/ui/views/top-blocked-truncated.es6.js
```diff
@@ -22,6 +22,7 @@ TruncatedTopBlocked.prototype = window.$.extend({},
     {
 
         _seeAllClick: function () {
+            this.model.fetch({ fire: 'eps' })
```
andrey-p: `firePixel`?

unit-test/background/pixel.es6.js
MariagraziaAlastra: oh, welp that's exactly what it did before and it looked wrong so I changed it. Sorry, of course we want the random number to still be attached to the URL -__- my bad

MariagraziaAlastra: Basically I forgot the pixelName is added separately, welp

andrey-p:
All good now, thank you!

shared/js/background/pixel.es6.js
russellholt: I wonder if you could trigger that by clicking on the extension menu bar item as soon as you can on first install, or when restoring a browsing session

MariagraziaAlastra: - Clicking asap on first install: tried, no problems
- Restoring a browsing session: didn't try, good one thanks

andrey-p: @russellholt `settings` only needs to retrieve some stuff from extension local storage before it's ready, which is likely going to be somewhere in the millisecond range - much faster than a person can reasonably click the popup button.

The only way we can run into this problem is programmatically calling `fire()` somewhere from the background code immediately when the browser starts up, or when the extension is installed. I don't think we're likely to be doing that soon, though.

PR #290, opened 2018-09-05
New privacy grade mvp

andrey-p:

**CC:** @russellholt 

## Description:

Plugs in the new privacy grade class.

## Steps to test this PR:

Run `npm install`, then:

1. Test specific sites. Pick several sites. Here's how to test the grades are being set correctly:

- Open the console on the background page. Set a console filter for "Grade".
- Open the site in question
- Verify grade
    - Check the Grade object in the background page console log - open "scores" and note what grades are listed there
    - Check that the grade is the same across the toolbar icon, the rating hero, and the grade details subview (at the bottom)
- Verify trackers
    - Check the "entitiesBlocked" and the "entitiesNotBlocked" entries in the background tab console.log
    - Each of the entries in "entitiesBlocked" should be listed as a network in the Trackers subview
    - If there's more than one entry in "entitiesNotBlocked", it would be in the Trackers subview with a (not blocked) on it. (If there's one entity there, it might refer to the site's own parent entity, so it might not be listed in the Trackers subview)
    - In the grade details subview the number of major tracker networks should include all "entitiesBlocked" that have a prevalence of over 7% (the number may be higher than current prod)
    - If the site's own parent tracker prevalence is over 7%, the grade details subview should say "site is a major tracker network"
- Verify HTTPS
    - If the site supports HTTPS, the https portion of the site and enhanced scores should say 0
    - Otherwise it should say 10
- Verify privacy practices
    - If we have TOSDR info on the site, it should be shown in the privacy practices subview as it currently is in production
    - If we list the privacy practices for the site as "bad", the privacy grade should have a privacy score of 7 or 10
    - If we list the privacy practices for the site as "mixed" (e.g. for Github) the privacy score should be 1
    - If we list the privacy practices as "good" the privacy score should be 0
    - If we list the privacy practices as "unknown" the privacy score should be 2

2. Test that special rating states ("calculating", special pages, etc) still work correctly
3. Test that element hiding still works on Oath sites (just checking it still works should be good)

## Automated tests:
- [ ] Unit tests
- [ ] Integration tests

###### Reviewer Checklist:

###### PR Author Checklist:


shared/js/background/classes/site.es6.js
```diff
@@ -7,21 +7,34 @@
...
-        this.score = new Score(this.specialDomain(), this.domain)
+        this.score = new Grade()
         this.whitelisted = false // user-whitelisted sites; applies to all privacy features
         this.setWhitelistStatusFromGlobal(domain)
         this.isBroken = this.checkBrokenSites(domain) // broken sites reported to github repo
         this.didIncrementCompaniesData = false
 
+        this.tosdr = privacyPolicy.getTosdr(domain)
+
+        this.parentEntity = utils.findParent(domain) || ''
```
andrey-p: Note, the usage for `findParent` changed in https://github.com/duckduckgo/duckduckgo-privacy-extension/pull/244, which got reverted.

I've used it as it was as of #244. So, if we merge this PR before #244 is brought back, **we need to change this line**.

shared/js/background/classes/site.es6.js
```diff
@@ -7,21 +7,34 @@
...
+const Grade = require('privacy-grade').Grade
+const trackerPrevalence = require('../../../data/tracker_lists/prevalence')
 
 class Site {
     constructor (domain) {
         if (domain) domain = domain.toLowerCase()
         this.domain = domain
         this.trackerUrls = []
-        this.score = new Score(this.specialDomain(), this.domain)
+        this.score = new Grade()
```
andrey-p: I should probably rename this to `this.grade` for consistency.

shared/js/background/trackers.es6.js
```diff
@@ -193,6 +194,7 @@ function getReturnTrackerObj (tracker, request, reason) {
     let fullURL = request.url ? request.url : request
     return {
         parentCompany: tracker.data.c,
+        prevalence: prevalence[tracker.data.c],
```
andrey-p: Default to `0` if there's nothing.

shared/js/background/classes/tab.es6.js
```diff
@@ -16,9 +16,13 @@
  */
 const scoreIconLocations = {
     'A': 'img/toolbar-rating-a@2x.png',
+    'B+': 'img/toolbar-rating-b-plus@2x.png',
     'B': 'img/toolbar-rating-b@2x.png',
+    'C+': 'img/toolbar-rating-c-plus@2x.png',
     'C': 'img/toolbar-rating-c@2x.png',
     'D': 'img/toolbar-rating-d@2x.png',
+    // we don't currently show the D- grade
+    'D-': 'img/toolbar-rating-d@2x.png',
```
andrey-p: Ask Russell if we should just drop it from the privacy-grade repo

shared/js/background/privacy-practices.es6.js
```diff
@@ -0,0 +1,115 @@
...
+    getTosdrScore (hostname) {
+        const domain = tldjs.getDomain(hostname)
+        const parent = utils.findParent(hostname)
+
+        // grab the first available val
+        // starting with most general first
+        const score = [
+            tosdrScores[parent],
+            tosdrScores[domain],
+            tosdrScores[hostname]
```
andrey-p: Add comment about possible edge case: what if `foo.bar.com` and `bar.com` are both in TOSDR and have different privacy scores?

shared/js/ui/models/site.es6.js
```diff
@@ -40,8 +47,8 @@ Site.prototype = window.$.extend({},
                         this.set('tab', tab)
                         this.domain = tab.site.domain
                         this.fetchSiteRating()
-                        this.set('tosdr', tab.site.score.tosdr)
-                        this.set('isaMajorTrackingNetwork', tab.site.score.isaMajorTrackingNetwork)
+                        this.set('tosdr', tab.site.tosdr)
+                        this.set('isaMajorTrackingNetwork', tab.site.parentPrevalence >= 7)
```
andrey-p: Use a constant for `7`.

shared/js/ui/templates/shared/grade-scorecard-grades.es6.js
```diff
@@ -13,8 +13,8 @@ function getGrades (rating, isWhitelisted) {
 
     // transform site ratings into grades
     // that the template can display more easily
-    const before = rating.before
-    const after = rating.after
+    const before = rating.cssBefore
+    const after = rating.cssAfter
```
andrey-p: These are now called `cssBefore` and `cssAfter`.

package.json
```diff
@@ -61,6 +61,7 @@
     "timekeeper": "^2.1.1"
   },
   "dependencies": {
+    "privacy-grade": "https://github.com/duckduckgo/privacy-grade#master",
```
andrey-p: https://github.com/duckduckgo/privacy-grade/pull/16, pending response from Russell about D-.

shared/js/background/trackers.es6.js
```diff
@@ -309,7 +313,7 @@ function checkABPParsedList (list, url, siteDomain, request) {
  *   * to content scripts to start the process of hiding blocked ads
  *    */
 function tryElementHide (requestData, tab) {
-    if (tab.parentEntity === 'Oath') {
```
andrey-p: This is a very small change. I noticed we were already setting the parentEntity on the Tab class because of element hiding. I've changed it to use on the Site class, to avoid getting the parent multiple times unnecessarily. (See also: https://github.com/duckduckgo/duckduckgo-privacy-extension/pull/290/files#diff-50231d3e2a7216164812f45897e7ec77L42)

cc @dharb

integration-test/background/grades.js
```diff
@@ -0,0 +1,56 @@
+/* global dbg:false */
+const harness = require('../helpers/harness')
+
+const tests = [
+    { url: 'duckduckgo.com', siteGrade: 'A', enhancedGrade: 'A' },
+    { url: 'theguardian.com', siteGrade: 'D-', enhancedGrade: 'B+' },
+    { url: 'google.com', siteGrade: 'D', enhancedGrade: 'D' }
```
andrey-p: TODO add more.

integration-test/background/grades.js
```diff
@@ -0,0 +1,82 @@
+/* global dbg:false */
+const harness = require('../helpers/harness')
+
+const tests = [
+    { url: 'duckduckgo.com', siteGrade: 'A', enhancedGrade: 'A' },
+    { url: 'theguardian.com', siteGrade: 'D-', enhancedGrade: 'B+' },
+    { url: 'google.com', siteGrade: 'D', enhancedGrade: 'D' },
+    { url: 'amazon.com', siteGrade: 'D-', enhancedGrade: 'C+' },
+    { url: 'facebook.com', siteGrade: ['D', 'C+'], enhancedGrade: 'C+' },
```
andrey-p: Was seeing failures with this one - it switched between having a single blocked tracker and none at all. I figured I might as well allow a bit of a leeway when determining the site grade.

MariagraziaAlastra:
@andrey-p I also noticed, when loading a site, sometimes the initial original and enhanced grade is incorrect, but then it gets fixed when loading is complete. (eg on roam.co at first the original grade is C and enhanced C+, but after loading the original is D and the enhanced is B).

I'm not sure if you can do anything about that, but even in production, when it takes a while to load a site we at least show the correct original grade, and then the enhanced grade - so eg in the previous example, production would show D as grade at first and then enhance it as B, after it finishes calculating.

Other than these two things, everything else looks good and tests pass!

shared/js/ui/views/tracker-networks.es6.js
```diff
@@ -77,9 +77,10 @@ TrackerNetworks.prototype = window.$.extend({},
 
         _rerender: function (e) {
             if (e && e.change) {
-                if (e.change.attribute === 'isPartOfMajorTrackingNetwork' ||
-            e.change.attribute === 'isWhitelisted' ||
-            e.change.attribute === 'totalTrackerNetworksCount') {
+                if (e.change.attribute === 'isaMajorTrackingNetwork' ||
+                    e.change.attribute === 'isWhitelisted' ||
+                    e.change.attribute === 'totalTrackerNetworksCount' ||
+                    e.change.attribute === 'siteRating') {
```
andrey-p: @MariagraziaAlastra this is the line that resolves the green tracker networks icon issue. There was a race condition which meant the `siteRating` was updated after the view had already rendered, which meant the site became D and didn't get reflected in the tracker networks icon.

I'm not sure why we only just saw it in this PR. Possibly because the threshold for a site being "D" is slightly different.

shared/js/background/classes/site.es6.js
```diff
@@ -1,29 +1,49 @@
...
+
+        if (this.parentEntity && this.parentPrevalence) {
+            this.grade.setParentEntity(this.parentEntity, this.parentPrevalence)
+        }
+
+        this.grade.setPrivacyScore(privacyPractices.getTosdrScore(domain))
+
+        if (this.url.match(/^https:\/\//)) {
+            this.grade.setHttps(true, true)
+        }
```
andrey-p: @MariagraziaAlastra this resolves the delayed score bug. Basically, HTTPS on the grade didn't get updated until the tab had finished loading, so we considered HTTPS to be false until then.

Which was kind of silly because we'd already know the URL contains "https".

I've added an initial check to the `Site` class, and also left the final check when the tab finishes loading, just in case we ended up downgrading the tab.

(This also made this check on startup redundant: https://github.com/duckduckgo/duckduckgo-privacy-extension/pull/290/files#diff-3cd94a6f60c9bfd33dffeca9dc0735f8L199)

MariagraziaAlastra:
Hey @andrey-p, thanks for the fixes! However, now I'm noticing the original grade with and without privacy protection is different. Eg for nytimes.com, it's C with privacy protection on and D when it's disabled:
screen shot 2018-09-19 at 12 50 49 pm
screen shot 2018-09-19 at 12 50 35 pm

MariagraziaAlastra:
Attaching scores:
screen shot 2018-09-19 at 12 53 01 pm
screen shot 2018-09-19 at 12 52 55 pm

shared/js/background/classes/site.es6.js
MariagraziaAlastra: Hey @andrey-p, actually I see a slightly different behavior here now. Again on roam.co, now initially the original grade vs enhanced are B and B+. Then it gets updated to D and B+.
I wonder if there are other parts of the score that get loaded later on?

shared/js/ui/views/tracker-networks.es6.js
MariagraziaAlastra: Cool, looks fixed now, thanks!

andrey-p:
> Hey @andrey-p, thanks for the fixes! However, now I'm noticing the original grade with and without privacy protection is different. Eg for nytimes.com, it's C with privacy protection on and D when it's disabled

This is kinda expected, and happens in production (e.g. roam.co's `before` changes from C to D when you whitelist it). If we block a tracker, we may be preventing a whole slew of other trackers from being loaded. There isn't really an easy way of addressing this at the moment, annoyingly.

shared/js/background/classes/site.es6.js
andrey-p: I'll look at it. This may be due to more trackers being loaded as the page does its thing though.

MariagraziaAlastra:
LGTM, thanks @andrey-p!

PR #295, opened 2018-09-10
return early if e.data is undefined

### LINKED ISSUE #294: cannot read property 'type' of null
screen shot 2018-09-10 at 6 27 08 pm

Think it's happening here: https://github.com/duckduckgo/duckduckgo-privacy-extension/blob/develop/shared/js/content-scripts/element-hiding.js#L69

Was a Mailchimp email list subscription confirmation page with the Google reCAPATCHA that caused it.

bsstoner:
Should fix #294, but I'm not sure if there's any negative side effects.

**Reviewer:**

## Automated tests:
- [ ] Unit tests
- [ ] Integration tests

###### Reviewer Checklist:

###### PR Author Checklist:

### DIFF
--- a/shared/js/content-scripts/element-hiding.js
+++ b/shared/js/content-scripts/element-hiding.js
@@ -65,7 +65,7 @@
          * 3. iframes send messages to main frame when they contain blocked elements (hideFrame)
          */
         frameListener (e) {
-            if (this.disabled) return
+            if (this.disabled || !e.data) return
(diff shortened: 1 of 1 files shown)


dharb:
Thanks @bsstoner for catching this! It looks like that recaptcha is passing a null data object via window.postMessage, which is tripping us up here. I actually didn't realize that you could even do that, I assumed it would be an empty object, not null. This looks good to me.

PR #307, opened 2018-09-20
Fixed mixed content issues by appending 'upgrade-insecure-request' header

dharb:
**Reviewer:** @andrey-p 
cc: @jdorweiler @zachthompson @russellholt 

## Description:
When a site is loaded over https, all active mixed content is blocked at the browser level before we have a chance to upgrade the requests to https. Mixed passive content is often blocked for several reasons (block-mixed-content header/meta tag, user agent). This PR fixes many sites broken due to improperly set up https websites that contain http resource links by appending the 'upgrade-insecure-requests' response header to the document.

## Steps to test this PR:
1. npm run test
2. visit site with mixed content, such as https://www.protrails.com/trail/744/aspen-snowmass-4-pass-loop. The mixed content should now load over https.

## Automated tests:
- [x] Unit tests
- [ ] Integration tests

###### Reviewer Checklist:

###### PR Author Checklist:

### DIFF
--- a/shared/js/background/chrome-events.es6.js
+++ b/shared/js/background/chrome-events.es6.js
@@ -21,6 +21,7 @@ const constants = require('../../data/constants')
 const redirect = require('./redirect.es6')
 const tabManager = require('./tab-manager.es6')
 const pixel = require('./pixel.es6')
+const https = require('./https.es6')
 
 chrome.webRequest.onBeforeRequest.addListener(
     redirect.handleRequest,
@@ -45,6 +46,14 @@ chrome.webRequest.onHeadersReceived.addListener(
     {
         urls: ['<all_urls>']
     }
+)
+
+chrome.webRequest.onHeadersReceived.addListener(
+    https.setUpgradeInsecureRequestHeader,
+    {
+        urls: ['<all_urls>']
+    },
+    ['blocking', 'responseHeaders']
 )
 
 /**
@@ -166,7 +175,6 @@ chrome.runtime.onMessage.addListener((req, sender, res) => {
  */
 
 const abpLists = require('./abp-lists.es6')
-const https = require('./https.es6')
 const httpsStorage = require('./storage/https.es6')
 
 // recheck adblock plus and https lists every 30 minutes
--- a/shared/js/background/chrome-events.es6.js
+++ b/shared/js/background/chrome-events.es6.js
@@ -49,7 +49,16 @@ chrome.webRequest.onHeadersReceived.addListener(
 )
 
 chrome.webRequest.onHeadersReceived.addListener(
-    https.setUpgradeInsecureRequestHeader,
+    (request) => {
+        // ignore background requests and requests without urls
+        if (request.tabId === -1 || !request.url) return {}
+
+        const tab = tabManager.get(request)
+
+        if (tab && tab.site && !tab.site.whitelisted) {
+            return https.setUpgradeInsecureRequestHeader(request)
+        }
+    },
     {
         urls: ['<all_urls>']
     },
--- a/shared/js/background/https.es6.js
+++ b/shared/js/background/https.es6.js
@@ -95,6 +95,52 @@ class HTTPS {
         // If it falls to here, default to reqUrl
         return reqUrl
     }
+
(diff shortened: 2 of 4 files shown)


shared/js/background/https.es6.js
```diff
@@ -95,6 +95,67 @@ class HTTPS {
...
+     *    to allow resources to be loaded on the https version of site.
+     */
+    setUpgradeInsecureRequestHeader (request) {
+        // skip requests to background tabs
+        if (request.tabId === -1) return {}
+
+        // Don't alter headers if site is whitelisted or http
+        if (request.url && request.url.indexOf('https://') === 0) {
+            let headersChanged = false
+            if (request.type === 'main_frame') {
```
dharb: We only need to add the 'upgrade-insecure-requests' CSP to the main document, as it propagates to all subrequests on the page.

shared/js/background/https.es6.js
```diff
@@ -95,6 +95,67 @@ class HTTPS {
...
+                // If no CSP header, add one
+                if (!cspHeaderExists) {
+                    const upgradeInsecureRequests = {
+                        name: 'Content-Security-Policy',
+                        value: 'upgrade-insecure-requests'
+                    }
+                    request.responseHeaders.push(upgradeInsecureRequests)
+                    headersChanged = true
+                }
+            } else {
```
dharb: Only need to edit access-control-allow-origin header on subrequests, since it doesn't really matter if the main document has this header.

unit-test/background/https.es6.js
```diff
@@ -52,4 +53,17 @@ describe('Https upgrades', () => {
...
     })
+
+    describe('https headers', () => {
+        beforeAll(() => {
+            https.isReady = true
+        })
+
+        httpsRequests.forEach((test) => {
+            it(`should ${test.testCase}`, () => {
+                console.log(test.expectedResult)
```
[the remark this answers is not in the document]
dharb: oooh yeah forgot I put that in there, will remove.

shared/js/background/chrome-events.es6.js
```diff
@@ -47,6 +48,14 @@ chrome.webRequest.onHeadersReceived.addListener(
     }
 )
 
+chrome.webRequest.onHeadersReceived.addListener(
+    https.setUpgradeInsecureRequestHeader,
```
andrey-p: Just posting a comment so this doesn't get lost - as we discussed over chat, this needs to not happen if the site is whitelisted.

shared/js/background/https.es6.js
dharb: I can't think of any instances where this would cause breakage.. access-control-allow-origin should either be *, null or an url, and this will only kick in when it's an url and the main frame is https. In this case, this header needs to contain an https url in order for the resource to load.

It's possible that there are sites that we force upgrade to https that include the upgrade-insecure-requests header but haven't updated their resource request headers to reflect the new access-control-allow-origin value. This could be the case if a site has an https version of the site available but they primarily use the http version. Probably quite rare.

andrey-p: Cool, let's leave it like that :+1:

shared/js/background/https.es6.js
```diff
@@ -95,6 +95,65 @@ class HTTPS {
...
+            let cspHeaderExists = false
+
+            for (const header in request.responseHeaders) {
+                // If CSP header exists and doesn't include upgrade-insecure-request
+                // directive, append it.
+                if (request.responseHeaders[header].name.match(/Content-Security-Policy/i)) {
+                    cspHeaderExists = true
+                    const cspValue = request.responseHeaders[header].value
+                    // exit loop if upgrade-insecure-requests directive present
+                    if (cspValue.match(/upgrade-insecure-requests/i)) break
```
andrey-p: I thought a bit about whether we should exit the loop early elsewhere once we've found the header we're interested in, but it seems some headers allow multiple entries (https://stackoverflow.com/a/4371395/). So this is probably good to leave as-is.

andrey-p:
Right, I think this is good. Let's merge!

PR #311, opened 2018-09-25
Block requests with type "beacon" or "ping"

MariagraziaAlastra:
**Reviewer:** @dharb 

## Description:
Beacon and ping type requests are slipping through currently, because they're not included in the list of resource types we recognize and block.
However, adding "beacon" to the list errors on Chrome, while "ping" doesn't seem to block anything in Firefox.
With this change we'll add either beacon or ping on extension install/update depending on the browser.
Also add "beacon" as a reason for blocking these requests.

## Steps to test this PR:
1. build and reload extension
2. check both firefox and chrome.
3. visit a site that uses beacons (google, youtube, amazon etc) and disable privacy protection
4. in the network tab you should be able to see the beacons loading (usually you can filter by "analytics" to make them show up)
5. enable privacy protection; now the beacon requests should be blocked
6. also run unit tests

## Automated tests:
- [x] Unit tests
- [ ] Integration tests

###### Reviewer Checklist:

###### PR Author Checklist:

### DIFF
--- a/shared/js/background/trackers.es6.js
+++ b/shared/js/background/trackers.es6.js
@@ -88,6 +88,7 @@ function isTracker (urlToCheck, thisTab, request) {
         // same as a first party requrest and return
         let trackerByParentCompany = checkTrackersWithParentCompany(urlSplit, siteDomain, request)
         if (trackerByParentCompany) {
+            trackerByParentCompany = checkReason(trackerByParentCompany)
             let commonParent = getCommonParentEntity(currLocation, urlToCheck)
             if (commonParent) {
                 return addCommonParent(trackerByParentCompany, commonParent)
@@ -96,6 +97,15 @@ function isTracker (urlToCheck, thisTab, request) {
         }
     }
     return false
+}
+
+// Check for reason to block the tracker
+// And add it
+// For now it just adds "beacon" if that's the tracker type
+function checkReason (trackerObj) {
+    if (trackerObj.type === utils.getBeaconName()) {
+        trackerObj.reason = 'beacon'
+    }
 }
 
 // add common parent info to the final tracker object returned by isTracker
--- a/shared/js/background/trackers.es6.js
+++ b/shared/js/background/trackers.es6.js
@@ -88,7 +88,7 @@ function isTracker (urlToCheck, thisTab, request) {
         // same as a first party requrest and return
         let trackerByParentCompany = checkTrackersWithParentCompany(urlSplit, siteDomain, request)
         if (trackerByParentCompany) {
-            trackerByParentCompany = checkReason(trackerByParentCompany)
+            trackerByParentCompany = addBeaconReason(trackerByParentCompany)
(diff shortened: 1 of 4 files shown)


shared/js/background/trackers.es6.js
```diff
@@ -98,6 +99,15 @@ function isTracker (urlToCheck, thisTab, request) {
     return false
 }
 
+// Check for reason to block the tracker
+// And add it
+// For now it just adds "beacon" if that's the tracker type
+function checkReason (trackerObj) {
```
russellholt: Since other nearby functions may also assign a reason may I suggest that this name is too generic. In this case I would probably just write this short bit of code inline above on line 91. You've already factored out `getBeaconName()` so it's pretty short. I think it is not likely to expand to cover more cases here since the other functions pretty clearly need to be refactored.

shared/js/background/utils.es6.js
```diff
@@ -91,6 +92,29 @@ function getUpgradeToSecureSupport () {
...
+
+// Add beacon or ping to the blocked types of requests
+// Depending on browser
+function updateRequestListenerTypes () {
+    if (!constants.requestListenerTypes ||
+        constants.requestListenerTypes.find(beaconName)) {
+        return
+    }
+
+    constants.requestListenerTypes.push(beaconName)
```
andrey-p: Sorry, I don't wanna end up in a "too many reviewers" situation but it seems really odd for `constants` to be edited on-the-fly. It goes against the expectation the name `constants` implies.

What do you think about just swapping this line over with a locally defined array which contains both `constants.requestListenerTypes` and either `ping` or `beacon`? https://github.com/duckduckgo/duckduckgo-privacy-extension/blob/cdc7c2ea72fc9d133a88befd2e2205cf3ce3d5cf/shared/js/background/chrome-events.es6.js#L32

I'm not sure this needs to live in Utils, either.

shared/js/background/trackers.es6.js
MariagraziaAlastra: I mostly created a function so it'd be easier to unit test - which is also why it needs to return something. I can change this though.

shared/js/background/utils.es6.js
MariagraziaAlastra: Thanks @andrey-p, this is a really good point. I'm now making a shallow copy in chrome events and adding the beacon/ping type there.
However, I left getBeaconName() in utils since it's used elsewhere and it looks like a good, intuitive place where to find it.
Thanks!

shared/js/background/trackers.es6.js
russellholt: Ok, then I would just make the name more specific. But this is so small a function, and the reason property is set in multiple places, maybe unit testing `getBeaconName()` is sufficient?

MariagraziaAlastra: Yeah, that's what I ended up doing. Thanks!

PR #313, opened 2018-09-26
Set hostname to empty string if it's null so extractHostFromUrl returns something

dharb:
**Reviewer:** @andrey-p 

## Description:
For whatever reason it appears that tldjs.parse(url) in extractHostFromUrl is returning an urlObj without a hostname. I would assume that this is due to tldjs choking on some kind of url, but I haven't been able to reproduce it so I'm not 100% sure. I just saw this error cropping up in the new chrome extension error page:
chrome_2018-09-25_22-51-18

This just makes sure that hostname is defined so no errors are thrown.

### DIFF
--- a/shared/js/background/utils.es6.js
+++ b/shared/js/background/utils.es6.js
@@ -5,7 +5,7 @@ function extractHostFromURL (url, shouldKeepWWW) {
     if (!url) return
 
     let urlObj = tldjs.parse(url)
-    let hostname = urlObj.hostname
+    let hostname = urlObj.hostname || ''
 
     if (!shouldKeepWWW) {
         hostname = hostname.replace(/^www\./, '')


andrey-p:
I wonder whether we can go one step further and drop `tldjs` for this specific method.

The problem with `tldjs` is that it's extremely opinionated. If it encounters a URL that is not valid according to some specific RFC, it bails. For example, I've seen this bit cause problems:

screenshot from 2018-09-26 14-17-15

We need it for certain cases that need something smarter, like getting the domain name without subdomains (`bar.co.uk` from `foo.bar.co.uk`). However, in this case we need the whole hostname so `tldjs` can be replaced with something like https://www.npmjs.com/package/url-parse.

What do you think?

andrey-p:
Thanks for doing this BTW. We definitely need to make this stuff a less error prone.

andrey-p:
Merging this, it's fixing errors elsewhere. We can look at the tldjs thing at a later date. :+1:

dharb:
@andrey-p I realized I never followed up on your above comment. I've run into a slew of other issues stemming from tldjs as well (private TLDs are a big one). I found a fork of tldjs[1] that not only deals with the issue you mention above, but also allows us to specify that we don't want to include private TLDs in our url parsing. I think we should switch to it posthaste.

[1] https://github.com/remusao/tldts

remusao:
Hi! Maintainer of `tldts` here. I would be more than happy to answer any question or discuss changes you would need to satisfy your use-case in parsing hostnames/domains.

PR #317, opened 2018-09-27
Make sure redirect protection doesn't trigger accidentally

andrey-p:
**Reviewer:** @dharb 
**CC:** @nilnilnil @jdorweiler 

## Description:
It's currently possible for redirect protection to be triggered manually, e.g. if a user presses enter multiple times on the URL bar while it's loading.

This PR adds a limit of 7 hits per main frame URL within 3s. This should be pretty easy for a redirect loop to hit, not so much for a human being.

## Steps to test this PR:

1. Open a site which is currently vulnerable to HTTPS redirect loops (some NSFW Tumblr ones apparently?) and verify the redirect protection still does its thing
2. Try opening the same HTTP link manually, end verify that redirect protection doesn't trigger (e.g. no `ehd` pixel with error code `13`.)

## Automated tests:
- [ ] Unit tests
- [ ] Integration tests

###### Reviewer Checklist:

###### PR Author Checklist:

### DIFF
--- a/shared/js/background/classes/https-redirects.es6.js
+++ b/shared/js/background/classes/https-redirects.es6.js
@@ -9,7 +9,7 @@ const REQUEST_REDIRECT_LIMIT = 7
  * This class protects users from accidentally being sent into a redirect loop
  * if a site we've included into our HTTPS list redirects them back to HTTP.
  *
- * Every redirect we perform on a tab gets registered against this class.
+ * Every redirect we perform on a tab gets registered against an instance of this class.
  * If we hit too many redirects for a request, we block it via canRedirect().
  */
 
@@ -101,6 +101,12 @@ class HttpsRedirects {
         return canRedirect
     }
 
+    /**
+     * We regenerate tab objects every time a new main_frame request is made.
+     *
+     * persistMainFrameRedirect() is used whenever a tab object is regenerated,
+     * so we can maintain redirect loop protection across multiple main_frame requests
+     */
     persistMainFrameRedirect (redirectData) {
         if (!redirectData) { return }
 
--- a/shared/js/background/classes/https-redirects.es6.js
+++ b/shared/js/background/classes/https-redirects.es6.js
@@ -5,6 +5,14 @@ const constants = require('../../../data/constants')
 const MAINFRAME_RESET_MS = 3000
 const REQUEST_REDIRECT_LIMIT = 7
 
+/**
+ * This class protects users from accidentally being sent into a redirect loop
+ * if a site we've included into our HTTPS list redirects them back to HTTP.
+ *
+ * Every redirect we perform on a tab gets registered against this class.
+ * If we hit too many redirects for a request, we block it via canRedirect().
+ */
+
 class HttpsRedirects {
     constructor () {
         this.failedUpgradeHosts = {}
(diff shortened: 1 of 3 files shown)


shared/js/background/classes/https-redirects.es6.js
```diff
@@ -5,6 +5,14 @@ const constants = require('../../../data/constants')
 const MAINFRAME_RESET_MS = 3000
 const REQUEST_REDIRECT_LIMIT = 7
 
+/**
+ * This class protects users from accidentally being sent into a redirect loop
+ * if a site we've included into our HTTPS list redirects them back to HTTP.
+ *
+ * Every redirect we perform on a tab gets registered against an instance of this class.
+ * If we hit too many redirects for a request, we block it via canRedirect().
+ */
```
andrey-p: I realized this class doesn't have a high level description, so I did a commenting pass here and a few other places.

unit-test/background/classes/https-redirects.es6.js
```diff
@@ -3,110 +3,143 @@ const tk = require('timekeeper')
...
+
+const getImageRequest = () => {
+    id += 1
+
+    return {
+        requestId: id,
+        type: 'image',
+        url: 'http://example.com/cat.gif'
+    }
+}
```
andrey-p: Took a moment to simplify the tests by adding helpers. We were creating these objects all over the place unnecessarily. Hopefully the tests are much more readable now.

shared/js/background/classes/https-redirects.es6.js
```diff
@@ -44,17 +59,23 @@ class HttpsRedirects {
          * Redirect loop detection is different when the request is for the main frame vs
          * any other request on the page.
          *
-         * For main frames, the redirect loop could happen like this:
-         * 1. We upgrade a page to HTTPS
-         * 2. The server gives 200 but prints out some HTML that redirects the page to HTTP
-         * 3. We try to upgrade again
+         * For main frames, the redirect loop could happen as part of several distinct hits to the same URL
+         * (e.g. we saw a case where a site returned 200 and the redirected to HTTP via Javascript)
```
andrey-p: The site that was previously causing this problem (http://raisingchildren.net.au/articles/safe_water_temperature.html) has fixed it since, so we have no way to test this case anymore. Good for them, not good for us.

I'd assume the redirect loop we encountered on their site would still be able to hit the same URL 7 times in 3 seconds.

andrey-p:
I was also considering dropping the timeout to 2s. Not sure whether that would risk some redirect loops going through. Seems like a balance we'd need to strike.

unit-test/background/classes/https-redirects.es6.js
dharb: Thanks for doing this! Good idea.

dharb:
Thanks @andrey-p, looks good to me! Let's stick a pin in this and revisit the multiple tab bug we were seeing after next week.

dharb:
Hey @andrey-p I think I found a site that breaks when we attempt to redirect this many times in a row due to a `429 Too Many Requests` http status:
image

It's actually broken with or without this logic, just in different ways (without this logic, the site doesn't seem to pass username and password to the popup window at all, it says 'no user credentials provided').

You can check it out yourself here: https://www.thelearningodyssey.com. Just enter random username and password and hit log in.

PR #318, opened 2018-09-27
Handle special domains better

andrey-p:
**CC:** @dharb 

## Description:

Just tightening up how we determine what's a special page and what isn't.

(Note, there's an outstanding bug where the popup might not update if you go from a non-special page to a special page. This PR doesn't fix that one.)

## Steps to test this PR:

Open a bunch of special pages on Chrome, Firefox and Vivaldi.
None of them should be showing a grade, and the description should be sane.

## Automated tests:
- [ ] Unit tests
- [ ] Integration tests

###### Reviewer Checklist:

###### PR Author Checklist:

### DIFF
--- a/shared/js/background/chrome-wrapper.es6.js
+++ b/shared/js/background/chrome-wrapper.es6.js
@@ -21,6 +21,10 @@ let getFromStorage = (key, cb) => {
     })
 }
 
+let getExtensionId = () => {
+    return chrome.runtime.id
+}
+
 let notifyPopup = (message) => {
     chrome.runtime.sendMessage(message)
 }
@@ -63,5 +67,6 @@ module.exports = {
     normalizeTabData: normalizeTabData,
     mergeSavedSettings: mergeSavedSettings,
     getDDGTabUrls: getDDGTabUrls,
-    setUninstallURL: setUninstallURL
+    setUninstallURL: setUninstallURL,
+    getExtensionId: getExtensionId
 }
--- a/shared/js/background/classes/site.es6.js
+++ b/shared/js/background/classes/site.es6.js
@@ -134,7 +134,7 @@ class Site {
         // FF-style about: pages don't get their domains parsed properly
         // so just extract the bit after about:
         if (url.match(/^about:/)) {
-            domain = url.match(/^about:([a-z\-]+)/)[1]
+            domain = url.match(/^about:([a-z-]+)/)[1]
             return domain
         }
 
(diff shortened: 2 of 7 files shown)


shared/js/background/chrome-wrapper.es6.js
```diff
@@ -21,6 +21,10 @@ let getFromStorage = (key, cb) => {
     })
 }
 
+let getExtensionId = () => {
+    return chrome.runtime.id
+}
```
andrey-p: Moved it to chrome-wrapper so I can stub it more easily.

shared/js/background/classes/site.es6.js
```diff
@@ -43,8 +44,8 @@ class Site {
             this.grade.setHttps(true, true)
         }
 
-        // set isSpecialDomain when the site is created
-        this.isSpecialDomain = this.specialDomain()
+        // set specialDomainName when the site is created
+        this.specialDomainName = this.getSpecialDomain()
```
andrey-p: Sorry, couldn't resist renaming these. Variables that pretend to be booleans and method names that pretend to not be method names are a pet peeve of mine.

shared/js/background/classes/site.es6.js
```diff
@@ -103,24 +104,56 @@ class Site {
...
-        if (this.domain === 'extensions') { return 'extensions' }
+    getSpecialDomain () {
+        const extensionId = browserWrapper.getExtensionId()
+        const url = this.url
+        let domain = this.domain
 
-        if (window.chrome && this.domain === chrome.runtime.id) { return 'options' }
+        if (url === '') {
+            return 'new tab'
+        }
```
andrey-p: In the previous version, we only did that if the browser is FF. I can't imagine a case in which the URL will be empty but we wouldn't want to consider it a new tab.

shared/js/background/utils.es6.js
```diff
@@ -2,7 +2,7 @@ const tldjs = require('tldjs')
 const entityMap = require('../../data/tracker_lists/entityMap')
 
 function extractHostFromURL (url, shouldKeepWWW) {
-    if (!url) return
+    if (!url) return ''
```
andrey-p: This was erroring out in the empty URL case. Not sure if it was ever hit in production, but I think it's worth maintaining consistency between different outputs.

(E.g. if a caller expects a function to return a string, always try and return a string.)

MariagraziaAlastra:
@andrey-p this looks good, only thing is on Vivaldi the message for the new tab is "extension page" as opposed to "new tab". I think it's pretty minor though, and it improves functionality, so LGTM 👍

Issue #323, opened 2018-10-03
Requests to local 127.0.0.1 on MusicBrainz.org are changed to HTTPS


phw:
On MusicBrainz.org when opening the URL with the tport=8000 parameter it shows green "tagger" links. Those links point to the local IP address 127.0.0.1 with the port specified with the parameter (e.g. 8000). This is used to communicate with the locally running Picard tagger [1].

Example: https://musicbrainz.org/recording/3795d1c1-c6ef-4f92-9493-7fd1e73bade2?tport=8000

This site contains http://127.0.0.1:8000/opennat?id=3795d1c1-c6ef-4f92-9493-7fd1e73bade2&t=1538558388

DuckDuckGo Privacy Essentials is changing those requests to communicate via HTTPS, which is not supported by Picard, which is acting as a local server. Ideally DuckDuckGo Privacy Essentials should never do HTTPS requests to local 127.0.0.1 addresses, but definitely not on musicbrainz.org for those specific links.

Please see the discussion on https://tickets.metabrainz.org/browse/PICARD-1359 , your feedback is welcomed.

dharb:
Thanks for the report! I'm looking into the issue. In the meantime, I would just toggle 'Site Privacy Protection' off in the extension pop up on musicbrainz.org. This will just turn off any sort of upgrading the extension does on that one site, and should resolve the problem for you.

phw:
Thanks a lot for the quick response.

> In the meantime, I would just toggle 'Site Privacy Protection' off in the extension pop up on musicbrainz.org.

Yes, that's what we suggesting users right now and this works well as a workaround.

PR #325, opened 2018-10-05
disable csp headers

### LINKED ISSUE #322: extension modifies request origin schema from HTTP to HTTPS
Extension enforces in a implicit way the server to switch CORS schema from HTTP to HTTP in the `access-control-allow-origin` header resulting in behavior below.

```
The 'Access-Control-Allow-Origin' header has a value 'https://localhost:3000' that is not equal to the supplied origin. Origin 'http://localhost:3000' is therefore not allowed access.
```
chrome: chrome/69.0.3497.100
ext. version: 2018.9.28

In devtools network tab there are origin and referer displayed correctly with HTTP - see screenshot (RIGHT is extension **enabled**, LEFT extension **disabled**)
screenshot from 2018-09-29 18-06-25

russellholt:
## Description:

Disabling adding/modifying CSP headers for now - will need much deeper testing before it is re-enabled.

Fixes #322, #323

### DIFF
--- a/shared/js/background/chrome-events.es6.js
+++ b/shared/js/background/chrome-events.es6.js
@@ -54,22 +54,27 @@ chrome.webRequest.onHeadersReceived.addListener(
     }
 )
 
-chrome.webRequest.onHeadersReceived.addListener(
-    (request) => {
-        // ignore background requests and requests without urls
-        if (request.tabId === -1 || !request.url) return {}
+// Temporarily disabling temporarily due to CSP causing problems.
+// Also, since the new HTTPS upgrade list does not allow mixed content
+// this will be less of an issue.
+// https://github.com/duckduckgo/duckduckgo-privacy-extension/issues/322
+//
+// chrome.webRequest.onHeadersReceived.addListener(
+//     (request) => {
+//         // ignore background requests and requests without urls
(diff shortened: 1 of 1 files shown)


dharb:
@azuroff - we’ve run into many instances of sites that appear to work properly over https when we crawl them, but then break when we upgrade them to https due to either 
A) http requests for mixed content being blocked at the browser level before they can be upgraded to https, or
B) requests containing an Access-Control-Allow-Origin header that specifies the http version of the site, which we’ve upgraded to https.

This change was an attempt to mitigate this breakage, but unfortunately had some unintentional side effects.

azuroff:
(A) makes perfect sense, as browsers throw all kinds of hissy fits if you request insecure content while on a secure page.

With (B), though, are you seeing requests with that particular header?  I've never seen it as a request header.  I suppose there could be servers (that aren't following the CORS spec) that look at that header, but I know that if I saw that header on the server side, I would just ignore it - it's the **Origin** request header I care about.  If there are servers using looking at that header in the request, it would made sense to modify it if you've upgraded the server to https, but don't modify the response header - that breaks things.  :)

You could always "encourage" adherence to the spec by deleting the Access-Control-Allow-Origin header if you see it in a request.  ;)

PR #330, opened 2018-10-10
Fix adding 'beacon' as blockable request type on FF

MariagraziaAlastra:
**Reviewer:** @dharb 

## Description:
#311 Added the beacon or ping type to the list of blockable requests, but Firefox sometimes gets mistaken as Chrome and we add "ping" (which FF claims is supported but doesn't do anything) instead of "beacon" to the list.
This happens because Chrome is the default browser name we use, in case getBrowserInfo() is not available. However, sometimes it takes a while for it to be available on FF and we set the request types for the onBeforeRequest handler too early.

This PR fixes that by double checking the UA string before returning the browser name.

## Steps to test this PR:
1. Build and reload extension
2. Go on sites with beacons like github.com
3. Check on the network tab and in the background that beacon requests (in github's case, it's google analytics) are blocked
4. Open the popup: all popup UI assets are properly loaded and the beacon is included in the blocked trackers count
5. Make sure to test both Chrome and Firefox

## Automated tests:
- [ ] Unit tests
- [ ] Integration tests

###### Reviewer Checklist:

###### PR Author Checklist:

### DIFF
--- a/shared/js/background/utils.es6.js
+++ b/shared/js/background/utils.es6.js
@@ -1,6 +1,8 @@
 const tldjs = require('tldjs')
 const entityMap = require('../../data/tracker_lists/entityMap')
 const constants = require('../../data/constants')
+const parseUserAgentString = require('../shared-utils/parse-user-agent-string.es6')
+const browserInfo = parseUserAgentString()
 
 function extractHostFromURL (url, shouldKeepWWW) {
     if (!url) return ''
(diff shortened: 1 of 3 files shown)


shared/js/background/utils.es6.js
```diff
@@ -74,7 +75,6 @@ try {
     chrome.runtime.getBrowserInfo((info) => {
```
MariagraziaAlastra: @dharb @andrey-p Do you know of any specific reason why we use getBrowserInfo() instead of checking the user agent? I added a check further down using the UA, further down, since the timing here seems pretty unreliable.

I wonder if it's worth refactoring/cleaning up a bit more, it's starting to look confusing to me

MariagraziaAlastra:
screen shot 2018-10-09 at 8 01 23 pm
screen shot 2018-10-09 at 8 01 11 pm

shared/js/background/utils.es6.js
andrey-p: Using both methods seems like it opens us up for potential confusion. I think I'm on the side of just using UA.

Another option would be to get the manifest synchronously and check for FF-specific properties, e.g.:

```js
const manifest = chrome.runtime.getManifest()

if (manifest.applications && manifest.applications.gecko) {
    browser = 'moz'
}
```

Might be worth moving that to `chrome-wrapper` so we don't have to account for this in e.g. unit tests.

MariagraziaAlastra: @andrey-p do you mean moving specifically the browser name stuff to chrome-wrapper? Or also the beacon vs ping functionality?
I'm also thinking, we use the User Agent string in the UI already, eg [here](https://github.com/duckduckgo/duckduckgo-privacy-extension/blob/3075db77fc06da08caaaa7cf44b75572448e4293/shared/js/shared-utils/parse-user-agent-string.es6.js)

MariagraziaAlastra:
@andrey-p @dharb I decided to use [parseUserAgentString()](https://github.com/duckduckgo/duckduckgo-privacy-extension/blob/develop/shared/js/shared-utils/parse-user-agent-string.es6.js) as a jumpstart for the browser/version detection, since it's already in place and we're using it elsewhere.

shared/js/background/utils.es6.js
```diff
@@ -62,45 +65,45 @@ function getCurrentTab (callback) {
...
-            var browserVersion = info.version.match(/^(\d+)/)[1]
-            if (browserVersion >= 59) {
-                upgradeToSecureSupport = true
-            }
-        }
-    })
-} catch (e) {}
-
+// Get correct name for fetching UI assets
 function getBrowserName () {
```
MariagraziaAlastra: Keeping getBrowserName() because we need the browser name to be either 'chrome' (lowercase) or 'moz' in order to fetch the right UI assets

PR #337, opened 2018-10-22
Make sure Site class is kept up-to-date

andrey-p:
**Reviewer:** @dharb 

## Description:

Adds the webNavigation.onCommitted event as another signal that we should update the site URL.

This fixes two problems:

- not updating the popup around special sites (e.g. chrome://extensions)
- not updating the popup when we get 200 from a service worker, rather than the server (see below)

## Steps to test this PR:

To test the first problem:

1. Open a new tab and open e.g. `chrome://extensions` or `about:addons` - the popup should say "extensions"
2. Open e.g. duckduckgo.com in the same tab - the popup should say duckduckgo.com
3. Open the extension page again in the same tab - the popup should say "extensions" again
4. Pressing back and opening the popup should always show the correct site name

And the second (requires a Gmail account):

1. Log into your Gmail account
2. Go to googlemail.com
3. The popup should now say `mail.google.com`.

And because of a small tweak I made to add tests to the Tab class, another check that needs to happen is:

1. Clean top trackers data if it's present
2. Visit a site 30 times to top trackers data is restored from scratch
3. Restart the browser
4. The top trackers data should be intact

## Automated tests:
- [ ] Unit tests
- [ ] Integration tests

###### Reviewer Checklist:

###### PR Author Checklist:

### DIFF
--- a/shared/js/background/chrome-events.es6.js
+++ b/shared/js/background/chrome-events.es6.js
@@ -63,7 +63,7 @@ chrome.webNavigation.onCommitted.addListener(
         // ignore navigation on iframes
         if (details.frameId !== 0) return
 
-        let tab = tabManager.get({tabId: e.tabId})
+        const tab = tabManager.get({tabId: details.tabId})
 
         if (!tab) return
 
--- a/shared/js/background/chrome-events.es6.js
+++ b/shared/js/background/chrome-events.es6.js
@@ -52,6 +52,25 @@ chrome.webRequest.onHeadersReceived.addListener(
     }
 )
 
+// keep track of URLs that the browser navigates to.
+//
+// this is currently meant to supplement tabManager.updateTabUrl() above:
+// tabManager.updateTabUrl only fires when a tab has finished loading with a 200,
+// which misses a couple of edge cases like browser special pages
+// and Gmail's weird redirect which returns a 200 via a service worker
+chrome.webNavigation.onCommitted.addListener(
+    (details) => {
+        // ignore navigation on iframes
+        if (details.frameId !== 0) return
+
+        let tab = tabManager.get({tabId: e.tabId})
+
+        if (!tab) return
+
+        tab.updateSite(details.url)
+    }
+)
+
 /**
  * TABS
  */
--- a/shared/js/background/chrome-events.es6.js
+++ b/shared/js/background/chrome-events.es6.js
@@ -241,6 +241,8 @@ let onStartup = () => {
             .catch(e => console.log(e))
 
         https.sendHttpsUpgradeTotals()
+
+        Companies.buildFromStorage()
     })
 }
 
--- a/shared/js/background/classes/site.es6.js
+++ b/shared/js/background/classes/site.es6.js
@@ -131,6 +131,14 @@ class Site {
             return domain
         }
 
+        // for some reason chrome passes this back from webNavigation events
+        // for new tabs instead of chrome://newtab
+        //
+        // "local-ntp" -> "local new tab page"
+        if (url.match(/^chrome-search:\/\/local-ntp/)) {
+            return 'new tab'
+        }
+
         // for special pages with a protocol, just return whatever
         // word comes after the protocol
         // e.g. 'chrome://extensions' -> 'extensions'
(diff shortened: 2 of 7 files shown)


shared/js/background/chrome-events.es6.js
```diff
@@ -52,6 +52,25 @@ chrome.webRequest.onHeadersReceived.addListener(
...
+        // ignore navigation on iframes
+        if (details.frameId !== 0) return
+
+        const tab = tabManager.get({tabId: details.tabId})
+
+        if (!tab) return
+
+        tab.updateSite(details.url)
+    }
+)
```
andrey-p: I was wondering whether, now that we have this, keeping the URL updated via `onHeadersReceived` was still necessary. We do still need to get the status code from that listener though.

shared/js/background/classes/site.es6.js
```diff
@@ -131,6 +131,14 @@ class Site {
             return domain
         }
 
+        // for some reason chrome passes this back from webNavigation events
+        // for new tabs instead of chrome://newtab
+        //
+        // "local-ntp" -> "local new tab page"
+        if (url.match(/^chrome-search:\/\/local-ntp/)) {
+            return 'new tab'
+        }
```
andrey-p: I'm not sure why Chrome returns this weird URL in the webNavigation event. Seems we can handle it pretty easily though.

shared/js/background/tab-manager.es6.js
```diff
@@ -118,8 +118,7 @@ class TabManager {
         if (tab) {
             tab.statusCode = request.statusCode
             if (tab.statusCode === 200) {
-                tab.url = request.url
-                tab.updateSite()
```
andrey-p: Moved setting the URL within `updateSite` so I don't have to do it in multiple places.

shared/js/background/classes/tab.es6.js
```diff
@@ -81,8 +81,11 @@ class Tab {
         }
     }
 
-    updateSite () {
-        this.site = new Site(this.url)
+    updateSite (url) {
+        if (this.site.url === url) return
```
andrey-p: `webNavigation.onCommitted` apparently happens after the server has started sending us data (https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/webNavigation/onCommitted) so `onHeadersReceived` will have already happened.

For the vast majority of sites, this guard will return for the `onCommitted` call because the URL would be exactly the same.

shared/js/background/chrome-events.es6.js
dharb: Really good call using onCommitted! I hadn't heard of it before, but looking at the docs it seems to be perfect for this case.

I suspect we can probably remove updating the url from `onHeadersReceived`, although I do wonder if that might expose some underlying race conditions since `onCommitted` occurs so much later in the cycle. I need to test this without the url updating in `onHeadersReceived` a bit more extensively, but as long as I don't run into any oddities there I would be pro removing it. One less check that has to be made on every request makes me happy.

shared/js/background/classes/site.es6.js
dharb: Thanks for including this! It had been bugging me for a while.

dharb:
Hey @andrey-p, this is really good! Sorry for dropping the ball on reviewing this sooner, totally slipped my mind. I've done quite a bit of manual testing and haven't been able to break anything, but what would you say to writing a quick unit test or two for this change before we merge?

shared/js/background/chrome-events.es6.js
```diff
@@ -222,6 +241,8 @@ let onStartup = () => {
             .catch(e => console.log(e))
 
         https.sendHttpsUpgradeTotals()
+
+        Companies.buildFromStorage()
```
andrey-p: This forces the companies module to build its contents from browser memory when the extension starts, rather than when the module gets loaded. It brings the companies module slightly more in line with other modules that depend on storage.

This was necessary so the Tab class can be used in an extension environment.

To test:

1. Clean top trackers data
2. Visit a site 30 times to top trackers data is restored from scratch
3. Restart the browser
4. The top trackers data should be intact

dharb:
Awesome, looks good! 🎆

PR #349, opened 2018-11-09
Memoize HTTPS subrequest checks

dharb:
**Reviewer:**

## Description:
When a webpage loads, we check every http subrequest made against our upgrade list to see if it can be upgraded to https. The trouble with this is, most sites make a lot of requests to the same domains, so this leads to us needlessly querying the bloom filter for domains that we have already checked. Querying the bloom filter is fast, but this still leads to unnecessary overhead.

Here's an example of 100+ bloom filter queries being done for the same half dozen domains on cafehulu.com:
image

Here's the same page, with memoized https upgrade checks:
virtualbox_2018-11-08_15-01-26

This PR simply memoizes the canUpgradeHost function so that we only need to check the bloom filter once per host per site.

## Steps to test this PR:
1. Add `console.log('cache hit')` here: https://github.com/duckduckgo/duckduckgo-privacy-extension/blob/dharb/memoize-bloom/shared/js/background/https.es6.js#L64 
2. Add `console.log('bloom filter query')` here: https://github.com/duckduckgo/duckduckgo-privacy-extension/blob/dharb/memoize-bloom/shared/js/background/https.es6.js#L66
2. Navigate to a site like cafehulu.com that makes a lot of http subrequests and open the background page console. You'll see that we've gone from 100+ bloom filter queries to 5-10. This is especially noticeable on sites that make a lot of requests to the same domain.

## Automated tests:
- [x] Unit tests
- [ ] Integration tests

###### Reviewer Checklist:

###### PR Author Checklist:

### DIFF
--- a/shared/js/background/classes/tab.es6.js
+++ b/shared/js/background/classes/tab.es6.js
@@ -44,6 +44,7 @@ class Tab {
         this.requestId = tabData.requestId
         this.status = tabData.status
         this.site = new Site(this.url)
+        this.httpsUpgrades = {}
         this.httpsRedirects = new HttpsRedirects()
         this.statusCode = null // statusCode is set when headers are recieved in tabManager.js
         this.stopwatch = {
--- a/unit-test/background/https.es6.js
+++ b/unit-test/background/https.es6.js
@@ -4,6 +4,7 @@ const httpsStorage = require('../../shared/js/background/storage/https.es6')
 const httpsBloom = require('./../data/httpsBloom.json')
 const httpsWhitelist = require('./../data/httpsWhitelist.json')
 const load = require('./../helpers/https.es6')
+const newTab = require('./../data/newTab.json')
 
 describe('Https upgrades', () => {
     beforeAll(() => {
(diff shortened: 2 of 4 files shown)


shared/js/background/classes/tab.es6.js
```diff
@@ -44,6 +44,7 @@ class Tab {
         this.requestId = tabData.requestId
         this.status = tabData.status
         this.site = new Site(this.url)
+        this.httpsUpgrades = {}
```
dharb: I ended up storing this here on the tab object, although I probably could have also done it on the site object. @andrey-p curious where you think the best place to store a short-lived cache like this is.

unit-test/background/https.es6.js
```diff
@@ -4,6 +4,7 @@ const httpsStorage = require('../../shared/js/background/storage/https.es6')
 const httpsBloom = require('./../data/httpsBloom.json')
 const httpsWhitelist = require('./../data/httpsWhitelist.json')
 const load = require('./../helpers/https.es6')
+const newTab = require('./../data/newTab.json')
```
dharb: I'm not fully sure why, but tests error out when I try to include the tab class. I get the error message `Uncaught ReferenceError: chrome is not defined`.

The newTab.json object I'm using here is just a fake tab object.

unit-test/background/https.es6.js
```diff
@@ -26,13 +27,17 @@ describe('Https upgrades', () => {
 
         it('should upgrade known upgradable domains', () => {
             testDomains.shouldUpgrade.forEach(domain => {
-                expect(https.canUpgradeHost(domain)).toEqual(true)
+                let tab = JSON.parse(JSON.stringify(newTab))
```
dharb: This is just a shallow copy so that the newTab object isn't modified.

shared/js/background/classes/tab.es6.js
andrey-p: Yeah this is probably the place for it.

unit-test/background/https.es6.js
andrey-p: It's because of the call to `browserWrapper.setBadgeIcon()`, which assumes an extension context. You'd need to stub that out. See the Site unit tests for an example: https://github.com/duckduckgo/duckduckgo-privacy-extension/blob/7c1bf9472ab5d55114586ea2c346afa8364de8a8/unit-test/background/classes/site.es6.js#L7-L9

andrey-p: Fixed in #337, see https://github.com/duckduckgo/duckduckgo-privacy-extension/pull/337/files#r232644157.

PR #350, opened 2018-11-10
Memoize Tracker Checks

dharb:
**Reviewer:**

## Description:
Many sites use trackers that continuously send requests to the same url as you browse a website. These often take the form of pixels, but they can also be beacons, scripts or even xhr requests. There are also tracking scripts that repeatedly attempt to phone home when they realize that their requests aren't completing successfully. 
Here's an example of comscore making repeated requests on yahoo.com:
image

We currently handle each request individually, and do a full lookup to identify it as a tracker. This ends up being a lot of unnecessary overhead, since we often already know the result of the look up. This PR simply memoizes the isTracker function so that only one full lookup is required per request url.

TODO: add tests

## Automated tests:
- [ ] Unit tests
- [ ] Integration tests

###### Reviewer Checklist:

###### PR Author Checklist:

### DIFF
--- a/shared/js/background/trackers.es6.js
+++ b/shared/js/background/trackers.es6.js
@@ -31,6 +31,7 @@ function loadLists () {
 function isTracker (urlToCheck, thisTab, request) {
     let currLocation = thisTab.url || ''
     let siteDomain = thisTab.site ? thisTab.site.domain : ''
+    let abbrUrl = urlToCheck.slice(0, urlToCheck.indexOf('?')).substring(0, 75)
     if (!siteDomain) return
 
     // DEMO embedded tweet option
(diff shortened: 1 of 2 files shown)


shared/js/background/trackers.es6.js
```diff
@@ -31,6 +31,7 @@ function loadLists () {
 function isTracker (urlToCheck, thisTab, request) {
     let currLocation = thisTab.url || ''
     let siteDomain = thisTab.site ? thisTab.site.domain : ''
+    let abbrUrl = urlToCheck.slice(0, urlToCheck.indexOf('?')).substring(0, 75)
```
dharb: Some tracking request urls are ridiculously long and contain very slightly different params, so I'm abbreviating them here by dropping everything after the ?, then limiting to 75 chars. This should help to reduce the total number of calls made for the similar tracking requests (eg pixels of the form domain.com/pixel.gif?tracking-params).

sammacbeth:
Closing as this seems to be based off old code. Feel free to re-open if that's not the case @dharb.

dharb:
Yep this logic moved to a different repo https://github.com/duckduckgo/privacy-grade

PR #370, opened 2019-04-03
Check for and parse additional success page params

moollaza:
**Reviewer:** @andrey-p 

**Depends on:**  Internal Changes

## Description:
Adds ability to parse additional url params from success page

## Steps to test this PR:
1. Visit https://duckduckgo.com?natb=v123-1zz&cp=atbzz
2. Install extension
3. Ajax call to /exti/ should include `cp=atbzz` (See Kibana)

## Automated tests:
- [x] Unit tests
- [x] Integration tests

###### Reviewer Checklist:

###### PR Author Checklist:

### DIFF
--- a/shared/js/background/atb.es6.js
+++ b/shared/js/background/atb.es6.js
@@ -10,6 +10,9 @@ const browserWrapper = require('./$BROWSER-wrapper.es6')
 
 const ATB_ERROR_COHORT = 'v1-1'
 
+// list of accepted params in ATB url
+const ACCEPTED_URL_PARAMS = ['cp']
+
 let dev = false
 
 const ATB = (() => {
@@ -102,7 +105,7 @@ const ATB = (() => {
             })
         },
 
-        finalizeATB: () => {
+        finalizeATB: (params) => {
             let atb = settings.getSetting('atb')
 
             // make this request only once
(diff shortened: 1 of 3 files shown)


shared/js/background/atb.es6.js
```diff
@@ -126,24 +129,42 @@ const ATB = (() => {
             return atb
         },
 
+        // iterate over a list of accepted params, and retrieve them from a URL
+        // builds a new query string containing only accepted params
+        getAcceptedParamsFromURL: (url) => {
+            let validParams = new URLSearchParams()
```
andrey-p: Nit, this should be `const` since we don't reassign to it.

shared/js/background/atb.es6.js
```diff
@@ -126,24 +129,42 @@ const ATB = (() => {
...
                         return !!atb
                     })
 
                     if (atb) {
                         settings.updateSetting('atb', atb)
                     }
 
-                    ATB.finalizeATB()
+                    // only pass params if we have an ATB
+                    ATB.finalizeATB(atb ? params : '')
```
andrey-p: This works perfectly, and very much :+1: on using the URL API in `getAcceptedParamsFromURL`.

I was wondering if we can generalize it further. E.g. we're reading the URL twice (in `getNewATBFromURL` and `getAcceptedParamsFromURL`) and sorta serializing it twice (in `getAcceptedParamsFromURL` and `finalizeATB`).

Maybe we could grab both `natb` and `cp` in `getAcceptedParamsFromURL`, do any validation within (e.g. correct ATB format, renaming from `natb` to `atb`) and return an object. Then maybe we can pass URL params wholesale to `finalizeATB`.

What do you think?

moollaza: I had almost implemented something like this, but opted to not change the current ATB logic to keep things simple. I totally agree tho that we're duplicating efforts with parsing the URL twice. I'll go ahead with your suggested approach :+1:

andrey-p:
:+1: !

PR #374, opened 2019-04-15
Improved broken site reporting

dharb:

**CC:** @jdorweiler 

## Description:
This PR streamlines the process for broken site reporting, and makes the reports more useful to us. It's important to note that this is all strictly opt-in, and it is totally up to the user if they would like to share any data at all about broken sites. The main changes are:
- The form is now totally contained within the popup, and triggered upon disabling privacy protection.
- The free form text field has been replaced with a categorization select menu, so that we may better classify breakage at scale. The free form text field produced a lot of noise that didn't help us to diagnose issues and fix them.
- If a user opts in to sharing extra data about a broken site with us, a few additional parameters will be added to the form. These include the domains of trackers that were blocked on the page, whether an https upgrade was performed on the site, and whether any surrogates were active on the page. The url of the site in question will also be attached to the form, with params and fragments removed to protect privacy.

## Steps to test this PR:
1. Open background page network tab
2. Visit any normal domain with this installed
3. Open popup and click whitelist toggle. An ept.off pixel should fire and a small form should appear under the toggle asking if the page is broken.
4. Click no in that form. It should thank you for your feedback and reload the page.
5. Remove site from whitelist by toggling privacy protection in the pop up. Check the network tab, an ept.on pixel should have fired that doesn't include the site url.
6. Repeat 2-4, this time hit yes.
7. Breakage categorization form should appear, with the submit button disabled.
8. Select a category from the dropdown. Submit button should enable.
9. Click submit. Thank you message should fade in and page should automatically reload in a couple seconds.
10. Check the network tab of the background page. An epbf pixel should have fired with the site url (minus fragments and params), trackers blocked, https upgrade status, category of breakage, surrogates, and atb. Trackers blocked and surrogates are comma separated values, and the site url should be URI encoded properly.
11. On the same site, toggle privacy protection back on. Check the network tab again, an ept.on pixel should have fired with the site domain and path attached.
12. Remove site from whitelist, open the popup and click report broken site in either the hamburger menu or the main area. Breakage form should appear. Fill it out and hit submit. This time you should see the thank you message, then the form disappear but no page reload or popup close.
13. Get to breakage form again, and X out of it. If you got to breakage form from toggling privacy protection, the form should close and the page should reload. Otherwise, the form should close but the page shouldn't reload.
14. Visit special site such as new tab page. Open the popup and click the hamburger menu. The report broken site option should be hidden. We limited this form to only submitting reports on the current page, so it doesn't make sense to allow users to submit reports about local pages.

TODO: finish unit tests, internal pr to register pixel.

## Automated tests:
- [ ] Unit tests
- [ ] Integration tests

###### Reviewer Checklist:

###### PR Author Checklist:


shared/js/ui/models/site.es6.js
```diff
@@ -248,7 +248,20 @@ Site.prototype = window.$.extend({},
...
                 const whitelistOnOrOff = this.isWhitelisted ? 'off' : 'on'
-                this.fetch({ firePixel: ['ept', whitelistOnOrOff] })
+
+                // fire ept.on pixel if just turned privacy protection on,
+                // fire ept.off pixel if just turned privacy protection off.
+                if (whitelistOnOrOff === 'on') {
+                    // If user reported broken site and opted to share data on site,
+                    // attach domain to ept.on pixel if they turn privacy protection back on.
+                    if (this.whitelistOptIn) {
+                        this.fetch({ firePixel: ['ept', 'on', this.tab.site.domain] })
```
MariagraziaAlastra: Could this whole block be simplified? it seems like there are really 2 cases here: is whitelist on + did the user opt in? ---> Then send additional data.
Otherwise, whether it's on or off ----> just send whitelistOnOrOff

shared/js/ui/models/site.es6.js
```diff
@@ -248,7 +248,20 @@ Site.prototype = window.$.extend({},
...
+                    } else {
+                        this.fetch({ firePixel: ['ept', 'on'] })
+                    }
+                } else {
+                    this.fetch({ firePixel: ['ept', 'off'] })
```
MariagraziaAlastra: super minor nit, but I wonder if it's worth having 'ept' in a variable, since it's repeated multiple times, so if the name changes you'd only have to replace it on one line :)

shared/js/ui/models/site.es6.js
dharb: Totally. I should've mentioned that I need to revisit this block anyway, as that whitelistOptIn actually isn't hooked into anything at the moment.

shared/js/ui/models/site.es6.js
```diff
@@ -258,6 +271,36 @@ Site.prototype = window.$.extend({},
                     }
                 })
             }
+        },
+
+        generateBreakagePixel: function (category) {
```
MariagraziaAlastra: I think this function could be tested with some mock data for the tab, what do you think?
The cases to test would be no tab, blocked and nonblocked trackers, and then surrogates vs everything else.

.gitignore
```diff
@@ -8,6 +8,7 @@ npm-debug.log
 public/js/*.*
 public/css/*.*
 *.swp
+*.swo
```
dharb: This was annoying me.. vim on linux leaves these.

shared/scss/views/_breakage-form.scss
```diff
@@ -0,0 +1,179 @@
...
+            padding: 0 28px 0 1em;
+            position: relative;
+            display: block;
+            font-size: 1em;
+            line-height: inherit;
+            min-width: 10em;
+            width: 100% !important;
+            height: 2.5em !important;
+            outline: none !important;
+            border: none !important
```
dharb: This is all a hack to make a custom select menu show up. Otherwise it looks really bad on some platforms, like linux.

shared/js/ui/models/whitelist.es6.js
```diff
@@ -25,6 +25,14 @@ Whitelist.prototype = window.$.extend({},
                     value: false
                 }
             })
+            // Remove domain whitelist opt-in status, if present
+            this.fetch({
```
dharb: We're storing opt-in settings per domain. For instance, say I'm on `example.com/index.html` and I whitelist the site and fill out the form, then I navigate to `example.com/help.html` and decide the site isn't actually broken and turn privacy protection back on. The domain and path will be attached to the ept.on pixel, because I previously opted in to sharing breakage data on this domain.

shared/js/ui/views/hamburger-menu.es6.js
```diff
@@ -53,9 +54,8 @@ HamburgerMenu.prototype = window.$.extend({},
 
         _handleBrokenSiteClick: function (e) {
             e.preventDefault()
-
-            let url = encodeURIComponent(this.model.tabUrl)
-            browserUIWrapper.openExtensionPage(`/html/feedback.html?broken=1&url=${url}`)
+            this.$el.addClass('is-hidden')
+            this.pageView.views.site.showBreakageForm('reportBrokenSite')
```
dharb: Passing a param to showBreakageForm so that the page doesn't automatically reload when the form is submitted, unless the user came to the form from toggling privacy protection (in which case we want to keep behavior consistent with what it was previously -- toggle privacy protection and site reloads with new setting).

shared/js/ui/views/site.es6.js
```diff
@@ -39,26 +41,29 @@ Site.prototype = window.$.extend({},
     {
         _onWhitelistClick: function (e) {
             if (this.$body.hasClass('is-disabled')) return
+
             this.model.toggleWhitelist()
-            this._showWhitelistedStatusMessage()
+            const whitelisted = this.model.isWhitelisted
+            this._showWhitelistedStatusMessage(!whitelisted)
+
+            if (whitelisted) {
```
dharb: Show breakage confirmation if user just whitelisted site

shared/js/ui/views/site.es6.js
```diff
@@ -39,26 +41,29 @@ Site.prototype = window.$.extend({},
     {
         _onWhitelistClick: function (e) {
             if (this.$body.hasClass('is-disabled')) return
+
             this.model.toggleWhitelist()
-            this._showWhitelistedStatusMessage()
+            const whitelisted = this.model.isWhitelisted
+            this._showWhitelistedStatusMessage(!whitelisted)
```
dharb: Automatically reload page if user just unwhitelisted a site

shared/scss/base/base.scss
```diff
@@ -298,7 +298,7 @@ a {
     &.icon__check {
         width: 22px;
         height: 22px;
-        margin-top: -4px;
+        margin-top: -2px;
```
dharb: design asked that I move this up a hair

shared/js/background/classes/site.es6.js
```diff
@@ -69,7 +70,7 @@ class Site {
      * and set the new site whitelist statuses
      */
     setWhitelistStatusFromGlobal () {
-        let globalwhitelists = ['whitelisted']
+        let globalwhitelists = ['whitelisted', 'whitelistOptIn']
```
dharb: We're creating a new list that mirrors the whitelist for white list opt ins. This will save which sites users have added to the whitelist and opted into sharing breakage data on. We are doing this so that we can attach the domain to the ept.on pixel when a user unwhitelists a site, which we think is a pretty good sign that the site isn't broken.

shared/js/ui/templates/breakage-form.es6.js
```diff
@@ -0,0 +1,55 @@
...
+            <div class="form__icon"></div>
+        </div>
+        <div class="breakage-form__element js-breakage-form-element">
+            <h2 class="breakage-form__title">Something Broken?</h2>
+            <div class="breakage-form__explanation">Submitting an anonymous broken site report helps us debug these issues and improve the extension.</div>
+            <div class="form__label__select">Describe What Happened</div>
+            <div class="form__select breakage-form__input--dropdown">
+                <select class="js-breakage-form-dropdown">
+                    <option value=''>Pick your issue from the list...</option>
+                    ${shuffle(categories).map(function (item) { return bel`<option value=${item.value}>${item.category}</option>` })}
```
dharb: Randomize the categories so people don't just pick the first one

shared/js/background/redirect.es6.js
```diff
@@ -66,10 +66,6 @@ function handleRequest (requestData) {
          * Tracker blocking
          * If request is a tracker, cancel the request
          */
-        if (window.chrome) {
-            // this can send an error message when the popup is not open. check lastError to hide it
-            chrome.runtime.sendMessage({'updateTabData': true}, () => chrome.runtime.lastError)
-        }
```
dharb: This was an old remnant that doesn't actually need to be there. It was causing every single request to rerender the popup, including those from background tabs that aren't trackers.

shared/js/ui/views/site.es6.js
```diff
@@ -121,8 +137,41 @@ Site.prototype = window.$.extend({},
...
+        _onConfirmNotBrokenClick: function () {
+            const isTransparentClass = 'is-transparent'
+            this.$confirmbreakagemessage.removeClass(isTransparentClass)
+            this.$confirmbreakage.addClass(isTransparentClass)
+            this.$body.removeClass('confirmation-active')
+            this.closePopupAndReload(1500)
+        },
+
+        _showBreakageConfirmation: function () {
+            this.$body.addClass('confirmation-active')
```
dharb: Adding a class here so that we can prevent rerenders when the confirmation form is active.

dharb:
Thanks for the review @MariagraziaAlastra!

PR #399, opened 2019-07-12
Full HTTPS coverage

kdzwinel:
**Reviewer:** @jdorweiler 

## Description:
We are adding support for remote Smarter Encryption service to cover even more websites.

## Steps to test this PR:

⚠️ Note that FF and Chrome implementations are slightly different (Chrome has to handle requests synchronously), so be sure to test in both browsers ✌️ 

1. Test https websites by going directly to their https versions:
- https://duckduckgo.com
- https://www.cnn.com/
- https://wikipedia.org

Those shouldn't be affected or even mentioned in the background page console.

2. Test https websites by going to their http versions:
- http://duckduckgo.com
- http://www.cnn.com/
- http://wikipedia.org

Those should be upgraded by our bloom filters ("HTTPS: Bloom filter - host is upgradable" in the bg page console)

3. Test http only websites that are not in our bloom filter e.g.
- http://burymewithmymoney.com/
- http://tinytuba.com/
- http://xinhuanet.com/
- http://soso.com/

Those are not in our bloom filters and will require checking with the SE service ("HTTPS Service: Requesting information for reverso.net "). You may see website being upgraded to https for a split of a second and then downgraded to http once response from the service returns. Subsequent loads should happen immediately, w/o us trying to upgrade because response should be cached ("HTTPS: Service cache - host is not upgradable").

4. Reload extension (to clear the cache), block access to the service (127.0.0.1 duckduckgo.com) and visit websites from #3 

Those should be broken because they are not in our bloom filters and when service is unavailable we default to HTTPS.

5. Test sites from https://staticcdn.duckduckgo.com/https/https-whitelist.json - those should NOT be upgraded ("HTTPS: Safelist - host is not upgradable") - note that those hosts can upgrade to HTTPS via redirect/HSTS themselves sometimes

6. Test sites from https://ddg-sandbox.s3.amazonaws.com/negative-whitelist.json - those should be upgraded ("HTTPS: Safelist - host is upgradable")

7. Make sure that turning off site privacy protection works and host is not updated

## Automated tests:
- [x] Unit tests
- [ ] Integration tests

###### Reviewer Checklist:

###### PR Author Checklist:


package.json
```diff
@@ -73,12 +73,13 @@
     "file-url": "2.0.2",
     "is-plain-object": "2.0.1",
     "jquery": "3.4.1",
-    "jsbloom": "git://github.com/duckduckgo/jsbloom.git#1.0",
+    "jsbloom": "git://github.com/duckduckgo/jsbloom.git#1.0.2",
```
kdzwinel: I had to fix a bug in jsbloom because it didn't allow to create two bloom filters in the same app: https://github.com/duckduckgo/jsbloom/releases/tag/1.0.2

package.json
```diff
@@ -73,12 +73,13 @@
...
     "normalize.scss": "0.1.0",
     "object.values": "1.0.4",
     "opn": "5.1.0",
     "parse-link-header": "1.0.0",
     "psl": "1.1.18",
+    "punycode": "1.4.1",
```
kdzwinel: We need to transform hostname to punycode before we query the service

shared/js/background/chrome-wrapper.es6.js
```diff
@@ -58,6 +58,12 @@ let setUninstallURL = (url) => {
     chrome.runtime.setUninstallURL(url)
 }
 
+const changeTabURL = (tabId, url) => {
+    return new Promise((resolve) => {
+        chrome.tabs.update(tabId, {url}, resolve)
+    })
+}
```
kdzwinel: I use it to downgrade from https to http when needed

shared/js/background/utils.es6.js
```diff
@@ -78,12 +78,6 @@ function findParent (url) {
     }
 }
 
-function getProtocol (url) {
-    var a = document.createElement('a')
-    a.href = url
-    return a.protocol
-}
```
kdzwinel: Unrelated, but I decided to replace it with URL since creating DOM elements all the time is expensive

shared/js/shared-utils/sha1.js
```diff
@@ -0,0 +1,372 @@
+/* eslint-disable */
```
kdzwinel: we can't use native function for this because we need to be sync in Chrome :(

kdzwinel:
@jdorweiler I merged with 'develop' branch and replaced temporary negative bloom filter URLs with the production ones (I had to make some changes to ` unit-test/helpers/https.es6.js` because of that). Could you take a last look at those last commits?

PR #417, opened 2019-11-15
Don't upgrade IPs, localhost and local TLDs to HTTPS

kdzwinel:
**Reviewer:** @dharb 

## Description:
This adds a shortpath for IPs, `localhost` and local TLDs (as defined in https://tools.ietf.org/html/rfc6761) which are not upgradabe by SE.
This fixes a problem where user doesn't have access to WAN, wants to access their ruter and our extension keeps upgrading local IP url to HTTPS making it impossible to access it.

This PR also:
- upgrades us from tldjs to tldts (finally) - so that we detect IPv6
- fixes integration tests

## Steps to test this PR:
1. Go to http://1.1.1.1, http://localhost, http://[::1], http://dev.localhost, http://example.test
2. Check in the extension's console that: a) there is a message saying that those were not upgraded because host is an IP, local TLD or `localhost` b) there are no requests to SE service

## Automated tests:
- [x] Unit tests
- [ ] Integration tests

###### Reviewer Checklist:

###### PR Author Checklist:


package-lock.json
```diff
@@ -12357,9 +12357,9 @@
       "integrity": "sha1-EEqOSqym09jNFXqO+L+rLXo//bY="
     },
     "ip": {
-      "version": "1.0.1",
-      "resolved": "https://registry.npmjs.org/ip/-/ip-1.0.1.tgz",
-      "integrity": "sha1-x+NWzeoiWucbNtcPLnGpK6TkJZA="
+      "version": "1.1.5",
```
kdzwinel: looks like we were already importing this package indirectly

shared/js/background/https.es6.js
```diff
@@ -59,6 +60,20 @@ class HTTPS {
...
+    /**
+     * Checks if provided hostname is local (e.g. ::1) or a private (e.g. 192.168.1.1)
+     * @param {string} host
+     * @returns {boolean}
+     */
+    isLocalOrPrivate (host) {
+        // if it looks like IP v6 remove [ and ] that wrap it ("[::1]" -> "::1") for ip library to work
+        if (host.startsWith('[') && host.endsWith(']')) {
+            host = host.slice(1, -1)
+        }
```
kdzwinel: TIL how IPv6 works in URLs. It's always in those brackets that, unfortunately, `ip` library does not understand. Thankfully it looks like domain name can't start with `[` so we can just strip it when we see it.

shared/js/background/utils.es6.js
```diff
@@ -7,9 +7,16 @@ const browserInfo = parseUserAgentString()
 function extractHostFromURL (url, shouldKeepWWW) {
     if (!url) return ''
 
-    let urlObj = tldjs.parse(url)
+    const urlObj = tldjs.parse(url)
```
kdzwinel: `tldjs` doesn't support IPv6, so this function was always returning an empty string before. My fix is… OKish.

shared/js/background/https.es6.js
```diff
@@ -60,10 +63,28 @@ class HTTPS {
     }
 
     /**
-     * @param {string} host
+     * @param {string} url either domain (example.com) or a full URL (http://example.com/about)
      * @returns {Boolean|Promise<Boolean>} returns true if host can be upgraded, false if it shouldn't be upgraded and a promise if we don't know yet and we are checking against a remote service
      */
-    canUpgradeHost (host) {
+    canUpgradeUrl (url) {
```
kdzwinel: this function is only used in one place - this file (besides test), so I felt comfortable changing it

integration-test/background/grades.js
```diff
@@ -3,9 +3,9 @@ const harness = require('../helpers/harness')
 
 const tests = [
     { url: 'duckduckgo.com', siteGrade: 'A', enhancedGrade: 'A' },
-    { url: 'theguardian.com', siteGrade: 'D-', enhancedGrade: 'B+' },
+    { url: 'theguardian.com', siteGrade: ['D', 'D-'], enhancedGrade: 'B+' },
     { url: 'google.com', siteGrade: 'D', enhancedGrade: 'D' },
-    { url: 'amazon.com', siteGrade: ['D-'], enhancedGrade: 'C' },
```
kdzwinel: amazon started detecting puppeteer as a bot

PR #452, opened 2020-05-19
Multi-step Onboarding: onboarding messaging

sballesteros:
**Reviewer:** @jdorweiler 
**CC:** @MariagraziaAlastra @moollaza 

## Description:
This implement an alternative post install workflow where the extension does not open the post install page but instead notify a new onboarding modal on the DuckDuckGo that it was successfully installed.

### Experiment

1. Open chrome and **uninstall** the extension if you have it installed already
2. Visit [https://ddh9.duckduckgo.com?t=hx](https://ddh9.duckduckgo.com?t=hx)
3. Clear all the cookies / local storage data associated with `ddh9.duckduckgo.com`
4. Reload [https://ddh9.duckduckgo.com?t=hx](https://ddh9.duckduckgo.com?t=hx)
5. You should see a "Get Started" button
    Screen Shot 2020-05-20 at 7 30 17 PM
6. Click on the "Get Started" button, you should see a modal and an animated button automatically opening a new window for the chrome store
    Screen Shot 2020-05-20 at 7 31 55 PM
7. **DO NOT INSTALL THE EXTENSION FROM THE CHROME STORE**, instead install it from this pull request by:
    - opening a new tab and visiting `chrome://extensions/`
    - loading the extension unpacked
8. Once the extension is installed the extension should:
    - [x] refocus the duckduckgo tab
    - [x] automatically unfold the second step of the modal
            Screen Shot 2020-05-20 at 7 36 50 PM

### control

1. visit [https://ddh9.duckduckgo.com/?t=hw](https://ddh9.duckduckgo.com/?t=hw)
2. install the extension from this PR
    - [x] check that the normal post install flow takes place

## Automated tests:
- [ ] Unit tests
- [ ] Integration tests

###### Reviewer Checklist:
- [x] **Ensure the PR solves the problem**
- [x] **Review every line of code**
- [x] **Ensure the PR does no harm by testing the changes thoroughly**
- [x] **Get help if you're uncomfortable with any of the above!**
- [x] Determine if there are any quick wins that improve the implementation

###### PR Author Checklist:

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -170,6 +170,9 @@ module.exports = function (grunt) {
         exec: {
             copyjs: `cp shared/js/*.js build/${browser}/${buildType}/js/ && rm build/${browser}/${buildType}/js/*.es6.js`,
             copyContentScripts: `cp shared/js/content-scripts/*.js build/${browser}/${buildType}/public/js/content-scripts/`,
+            // replace `/* __ */ 'https://duckduckgo.com' /* __ */` in content-scripts/onboarding.js for local dev
+            // make sure that sed works on both linux and OSX (see https://stackoverflow.com/questions/5694228/sed-in-place-flag-that-works-both-on-mac-bsd-and-linux)
+            devifyOnboarding: `sed -i.bak "s/\\/\\* __ \\*\\/ 'https:\\/\\/duckduckgo\\.com' \\/\\* __ \\*\\//'*'/" build/${browser}/${buildType}/public/js/content-scripts/onboarding.js && rm build/${browser}/${buildType}/public/js/content-scripts/onboarding.js.bak`,
             tmpSafari: `mv build/${browser}/${buildType} build/${browser}/tmp && mkdir -p build/${browser}/${buildType}/`,
             mvSafari: `mv build/${browser}/tmp build/${browser}/${buildType}/ && mv build/${browser}/${buildType}/tmp build/${browser}/${buildType}/${browser}`,
             mvWatchSafari: `rsync -ar build/${browser}/${buildType}/public build/${browser}/${buildType}/${browser}/ && rm -rf build/${browser}/${buildType}/public`
@@ -195,7 +198,7 @@ module.exports = function (grunt) {
             },
             contentScripts: {
                 files: watch.contentScripts,
-                tasks: ['browserify:banner', 'exec:copyContentScripts']
+                tasks: ['browserify:banner', 'exec:copyContentScripts', 'exec:devifyOnboarding']
             }
         },
 
@@ -225,7 +228,7 @@ module.exports = function (grunt) {
 
     grunt.registerTask('build', 'Build project(s)css, templates, js', ['sass', 'browserify:ui', 'browserify:background', 'browserify:backgroundTest', 'browserify:banner', 'execute:preProcessLists', 'safari'])
 
-    const devTasks = ['build']
+    const devTasks = ['build', 'exec:devifyOnboarding']
     if (grunt.option('watch')) { devTasks.push('watch') }
 
     grunt.registerTask('dev', 'Build and optionally watch files for development', devTasks)
--- a/shared/js/background/atb.es6.js
+++ b/shared/js/background/atb.es6.js
@@ -173,6 +173,30 @@ const ATB = (() => {
             // - the user wasn't already looking at the app install page
             // - the user hasn't seen the page before
             settings.ready().then(() => {
+                // TODO XXX only run that for CPPD experiment
+                chrome.tabs.query({currentWindow: true}, (tabs) => {
+                    const tab = tabs.find((tab) => {
+                        const { hostname, searchParams } = new URL(tab.url)
+
+                        return (hostname.includes('duckduckgo.com') /*&& !!searchParams.get('natb')*/)
+                    })
+
+                    if (tab) {
+                        chrome.tabs.highlight({
+                            tabs: [tab.index],
+                            windowId: tab.windowId
+                        }, () => {
+                            chrome.tabs.executeScript(tab.id, { file: 'public/js/content-scripts/onboarding.js' })
+                        })
+                    } else {
+                        // TODO XXX recreate the tab?
+                        // chrome.tabs.create({ url })
+                    }
+                })
+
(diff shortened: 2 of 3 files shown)


shared/js/background/atb.es6.js
```diff
@@ -173,18 +176,43 @@ const ATB = (() => {
...
+                if (settings.getSetting('isMultiStepOnboarding')) {
+                    // We find the tab containing the cross product promotion modal
+                    // and inject a content script that will notify the page that the
+                    // extension was successfully installed
+                    // Note: that we don't know on which window that tab is
+                    chrome.tabs.query({}, (tabs) => {
+                        const tab = tabs.find((tab) => {
+                            const { hostname, searchParams } = new URL(tab.url)
+
+                            return (hostname.includes('duckduckgo.com') && !!searchParams.get('natb'))
```
moollaza, line 189: We might want to make this more explicit in production and specify the url must be `https://duckduckgo.com`

Otherwise something like `https://not-duckduckgo.com?natb=1` could trigger this

sballesteros: Good point, we have https://www.npmjs.com/package/psl in the package.json so I can use that and check on the domain rigorously

shared/js/background/atb.es6.js
```diff
@@ -173,18 +176,43 @@ const ATB = (() => {
...
                         })
-                    }
-                })
+
+                        const file = 'public/js/content-scripts/onboarding.js'
+                        if (tab) {
```
moollaza, line 193: If there's no DDG tab I think we'll want to open a new tab to the homepage so they continue with the multi-step onboarding?

shared/js/background/atb.es6.js
sballesteros: actually `psl` is only used for the script so will just do simple check.

sballesteros: should be fixed by https://github.com/duckduckgo/duckduckgo-privacy-extension/pull/452/commits/1ed4e4afb423f609e0bb40ec17c2e9718f4ed646

shared/js/background/atb.es6.js
sballesteros: apologies if I am missing something but I don't think that we can detect that no ?
If the user closed the tab _before_ installing the extension then `settings.getSetting('isMultiStepOnboarding')` will be `false`  and we can't detect that the install was triggered by the new onboarding workflow.

moollaza: Yup you're right. Not sure what I was thinking

shared/js/content-scripts/onboarding.js
```diff
@@ -0,0 +1,15 @@
+(function () {
+    // Note: this is injected by atb.es6.js
+    // This is currently used for the cross product promotion on desktop experiment.
```
moollaza, line 3: I think we can simplify this to say it is only used on duckduckgo.com to assist with user onboarding.

sballesteros: https://github.com/duckduckgo/duckduckgo-privacy-extension/pull/452/commits/9819655facb88e89a8382a138a8618a040e9a4df

shared/js/content-scripts/onboarding.js
```diff
@@ -0,0 +1,12 @@
+(function () {
+    // Note: this is only used on duckduckgo.com to assist with user onboarding
+    if (document.readyState === 'interactive' || document.readyState === 'complete') {
+        start()
+    } else {
+        document.addEventListener('DOMContentLoaded', start)
+    }
+
+    function start () {
+        window.postMessage({ type: 'onboarding', ready: true }, '*')
```
moollaza, line 10: According to MDN seems like we should really try to avoid using `*` for the target origin.

I'm thinking we should have `onboarding.js` and `onboarding-dev.js` where the target origin is `'https://duckduckgo.com'` and `'*'`.

We can add a `setDevMode` method to ATB.js and enable it in [debug.es6.js](https://github.com/duckduckgo/duckduckgo-privacy-extension/blob/develop/shared/js/background/debug.es6.js). With that we can load either the production or dev file accordingly.

moollaza:
@sballesteros this looks great. I left one comment about potential security issues and best practices for `window.postMessage`. Arguably its not a concern because nothing important is being passed from the extension, but seems like we should try to use a targetOrigin.

sballesteros:
@moollaza see https://github.com/duckduckgo/duckduckgo-privacy-extension/pull/452/commits/07a23e08f586a0b2a2ebc2d810977d25e2ab2df1 
I am not sure that there is an issue with the target origin here but that sets it to `https://duckduckgo.com` in prod and `*` in dev (for local testing convenience).
The content scripts are not run through the babelify transform so I couldn't leverage the existing tooling. 
Note that the new `exec:devifyOnboarding` task is only run for dev build and should not affect prod builds.

Gruntfile.js
```diff
@@ -170,6 +170,8 @@ module.exports = function (grunt) {
         exec: {
             copyjs: `cp shared/js/*.js build/${browser}/${buildType}/js/ && rm build/${browser}/${buildType}/js/*.es6.js`,
             copyContentScripts: `cp shared/js/content-scripts/*.js build/${browser}/${buildType}/public/js/content-scripts/`,
+            // replace `/* __ */ 'https://duckduckgo.com' /* __ */` in content-scripts/onboarding.js for local dev
+            devifyOnboarding: `sed -i${process.platform === 'darwin' ? " ''" : "''"} "s/\\/\\* __ \\*\\/ 'https:\\/\\/duckduckgo\\.com' \\/\\* __ \\*\\//'*'/" build/${browser}/${buildType}/public/js/content-scripts/onboarding.js`,
```
sballesteros, line 174: apologies for the horror here. The double escaping for the special charcaters is needed as we are in a JS template string and for what i can tell, the `sed` that comes standard with OSX (BSD?) needs `sed -i ''` (the suffix MUST be specified) whereas GNU sed is fine with just `sed -i`.

sballesteros: going to check quickly if we can find a version of the command that avoids the process.platform check...

moollaza:
@sballesteros Thanks! This LGTM :+1:

PR #456, opened 2020-06-10
Update README to point to Safari

brindy:
Update the README to include information about the Safari extension.


brindy:
Thanks @dharb!

PR #468, opened 2020-07-10
Fingerprint protection

Charlie-belmer:
**Reviewer:** Konrad

## Description:
Adding basic fingerprinting protection to the extension. This will prevent websites from gleaning screen, battery, and some system information provided by the window, navigator, battery, and screen API's.

## Steps to test this PR:
1. Run automated tests with npm run test-int
2. Run manual tests by installing the plugin and going to any page. Open dev tools, and check values of window.availTop or availLeft (should be 0).

## Automated tests:
- [ ] Unit tests
- [X] Integration tests

###### Reviewer Checklist:

###### PR Author Checklist:
- [X] Get advice or leverage existing code
- [X] Agree on technical approach with reviewer (if the changes are nuanced)
- [X] Ensure that there is a testing strategy (and documented non-automated tests)
- [X] Ensure there is a documented monitoring strategy (if necessary)
- [X] Consider systems implications

### DIFF
--- a/shared/data/fingerprint-protection.js
+++ b/shared/data/fingerprint-protection.js
@@ -241,20 +241,6 @@
         e.textContent = scriptToInject;
         (document.head || document.documentElement).appendChild(e)
 
-        // Inject into any and all iFrames
-        const frames = document.getElementsByTagName('iframe')
-        for (const frame of frames) {
-            try {
-                let fe = document.createElement('script')
-                fe.textContent = scriptToInject
-                frame.contentDocument.head.appendChild(fe)
-                if (removeAfterExec) {
-                    fe.remove()
-                }
-            } catch (e) {
-                console.log(`Couldn't inject into subframe. Sometimes this is due to cross domain frames, which are injected separately.`)
-            }
-        }
         if (removeAfterExec) {
             e.remove()
         }
--- a/shared/data/fingerprint-protection.js
+++ b/shared/data/fingerprint-protection.js
@@ -6,7 +6,7 @@
 
 (function protect () {
     // Property values to be set and their original values.
-    let fingerprintPropertyValues = {
+    const fingerprintPropertyValues = {
         'screen': {
             'availTop': {
                 'object': 'screen',
@@ -48,12 +48,12 @@
             'hardwareConcurrency': {
                 'object': 'navigator',
                 'origValue': navigator.hardwareConcurrency,
-                'targetValue': isMobile() ? 2 : 8
+                'targetValue': 8
             },
             'deviceMemory': {
                 'object': 'navigator',
                 'origValue': navigator.deviceMemory,
-                'targetValue': isMobile() ? 2 : 8
+                'targetValue': 8
             }
         },
         'storage': {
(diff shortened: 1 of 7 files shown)


shared/js/background/chrome-events.es6.js
```diff
@@ -201,6 +201,46 @@ chrome.runtime.onMessage.addListener((req, sender, res) => {
...
+        const experiment = settings.getSetting('experimentData')
+
+        if (experiment && experiment.fingerprint_protection) {
+            const whitelisted = settings.getSetting('whitelisted')
+            const tabURL = new URL(details.url) || {}
+            if (!whitelisted || !whitelisted[tabURL.hostname]) {
+                const scriptDetails = {
+                    'file': '/data/fingerprint-protection.js',
+                    'runAt': 'document_start',
+                    'allFrames': true
```
kdzwinel, line 223: It occurred to me that the issue that you mentioned with scirpt not being injected into certain iframes might be because they don't have any 'src' (e.g. because they were dynamically crated). This should be fixed by adding `matchAboutBlank: true` here. Could you please verify if this addresses the issue?

Charlie-belmer: It does seem to work, so the other iframe injections can be removed.

However after this, the iframes don't seem to receive screen size updates. This doesn't seem to cause any issues in testing.

kdzwinel: > It does seem to work, so the other iframe injections can be removed.

That's great, thank you for verifying - it will make the code simpler 🙌

> don't seem to receive screen size updates.

Hm, not sure what that's about, regular iframes seem to receive 'resize' events - might be a Chrome bug. I'll take a look when testing.

kdzwinel: When I played with the extension I saw some frames that are not receiving `resize` events, but I was able to reproduce this with and without the extension, so it doesn't seem to be related.

shared/data/fingerprint-protection.js
```diff
@@ -0,0 +1,270 @@
...
+     * ensuring that no information is leaked as the dimensions change, but also that the
+     * values change correctly for valid use cases.
+     */
+    function setWindowDimensions () {
+        let windowScript = ''
+        const normalizedY = normalizeWindowDimension(window.screenY, window.screen.height)
+        const normalizedX = normalizeWindowDimension(window.screenX, window.screen.width)
+
+        if (normalizedY <= fingerprintPropertyValues.screen.availTop.origValue) {
+            windowScript += setWindowPropertyValue('window.screenY', 0)
```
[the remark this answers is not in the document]
Charlie-belmer: Good call, fixed.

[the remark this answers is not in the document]
Charlie-belmer: On the jsbin - I see it working correctly on both chrome and FF. Maybe something I fixed between versions?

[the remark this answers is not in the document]
kdzwinel: Just retested and you are right - looks like you fixed it 👍

PR #469, opened 2020-07-21
Serve our own local surrogates

dharb:
**Reviewer:** @jdorweiler @kdzwinel 

## Description:
Two major changes are included in this PR:
1. Surrogates are now loaded from a local .txt file (shared/data/surrogates.txt)
2. Surrogates have been rewritten from the ground up. I plan to document and add them to https://github.com/duckduckgo/tracker-radar soon. We will also be generating a lot more of these in the future.

## Steps to test this PR:
1. build and install extension
2. visit a site that fetches a script that we have a surrogate for (eg vrbo.com, which loads google-analytics.com/analytics.js). Check that the surrogate is loaded by opening dev tools and looking in either the debugger tab (firefox) or sources tab (chrome) at the analytics.js resource.

###### Reviewer Checklist:

###### PR Author Checklist:

### DIFF
--- a/shared/data/surrogates.txt
+++ b/shared/data/surrogates.txt
@@ -100,7 +100,7 @@ google-analytics.com/inpage_linkid.js application/javascript
     window._gaq = (window._gaq === void 0) ? gaqObj : window._gaq
 })()
 
-google-analytics.com/cx/api.js application/javascript
+google-analytics.com/api.js application/javascript
 (() => {
     const noop = () => {}
     const cxApiHandler = {
@@ -255,7 +255,7 @@ googlesyndication.com/adsbygoogle.js application/javascript
     'use strict'
 })()
 
-doubleclick.net/instream/ad_status.js application/javascript
+doubleclick.net/ad_status.js application/javascript
 (() => {
     'use strict'
     window.google_ad_status = 1
@@ -329,7 +329,7 @@ outbrain.com/outbrain.js application/javascript
     window.outbrain_rater = (window.outbrain_rater === void 0) ? noopProxy : window.outbrain_rater
 })()
 
-amazon-adsystem.com/aax2/amzn_ads.js application/javascript
+amazon-adsystem.com/amzn_ads.js application/javascript
 (() => {
     'use strict'
     if (window.amznads) { return }
--- a/shared/data/surrogates.txt
+++ b/shared/data/surrogates.txt
@@ -1,0 +1,360 @@
+google-analytics.com/ga.js application/javascript
+(() => {
+    'use strict'
+    const noop = () => {}
+    const noopReturnEmptyArray = () => { return [] }
+    const noopHandler = {
+        get: function(target, prop) {
+            if (typeof target[prop] !== 'undefined') {
+                return Reflect.get(...arguments)
+            }
+            return noop
+        }
+    }
+    const trackerTarget = {
+        _getLinkerUrl: function(arg) { return arg }
+    }
+    const gaqTarget = {
+        push: function(arg) {
+            if (typeof arg === 'function') {
+                try {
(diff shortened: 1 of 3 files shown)


shared/data/surrogates.txt
```diff
@@ -0,0 +1,360 @@
+google-analytics.com/ga.js application/javascript
```
kdzwinel, line 1: I'm not a fan of this format - these should be separate JS files IMO. Is there any good reason to keep it? (thats something for org days, not for this PR)

shared/data/surrogates.txt
```diff
@@ -0,0 +1,360 @@
...
+
+google-analytics.com/inpage_linkid.js application/javascript
+(() => {
+    const gaqObj = {
+        push: () => {}
+    }
+    window._gaq = (window._gaq === void 0) ? gaqObj : window._gaq
+})()
+
+google-analytics.com/cx/api.js application/javascript
```
kdzwinel, line 103: As far as I can tell this surrogate will never match anything (and it haven't matched anything before this PR either). ID extracted from this URL is `cx` while ID used in `https://staticcdn.duckduckgo.com/trackerblocking/v2.1/tds.json` is `api.js`. I say we drop it.

ids

shared/data/surrogates.txt
```diff
@@ -0,0 +1,360 @@
...
+        }
+    }
+})()
+
+googlesyndication.com/adsbygoogle.js application/javascript
+(() => {
+    'use strict'
+})()
+
+doubleclick.net/instream/ad_status.js application/javascript
```
kdzwinel, line 258: ditto - extracted id is `instream` while actual id should be `ad_status.js`

failed-to-fetch

shared/data/surrogates.txt
```diff
@@ -0,0 +1,360 @@
...
+    }
+    window['OBR$'] = noop
+    window.OB_releaseVer = '200037'
+    window.OBR = (window.OBR === void 0) ? obrObj : window.OBR
+    window.OB_PROXY = (window.OB_PROXY === void 0) ? noopProxy : window.OB_PROXY
+    window.outbrain = (window.outbrain === void 0) ? noopProxy : window.outbrain
+    window.outbrain_rater = (window.outbrain_rater === void 0) ? noopProxy : window.outbrain_rater
+})()
+
+amazon-adsystem.com/aax2/amzn_ads.js application/javascript
```
kdzwinel, line 332: ditto - extracted id is `aax2` while referenced id is `amzn_ads.js`

shared/data/surrogates.txt
jdorweiler: I don't like it either, but I'd change the format later.

shared/data/surrogates.txt
jdorweiler: We could keep them and change the mapping here https://github.com/duckduckgo/tracker-radar/blob/main/build-data/static/surrogates.json

dharb: Good catch. For the time being, what do you think about updating these to:
google-analytics.com/api.js
doubleclick.net/ad_status.js
amazon-adsystem.com/amzn_ads.js
so that the mapping in Tracker Radar matches on them? That would make them work right away, and this change is only going out to the extension immediately. Then when we get these things into a better format (split up, checked into Tracker Radar) we can revisit and improve the logic?

jdorweiler: 👍 that makes more sense

shared/data/surrogates.txt
dharb: Agreed, this format is bad and these should be in individual files. Let's plan to fix it as a follow up when we check everything into Tracker Radar?

PR #486, opened 2020-08-26
User agent

Charlie-belmer:
This PR adds user agent spoofing for third party requests to the extension. Agents are chosen based on real world usage metrics, and are updated over time. 

**Reviewer:** Lucas

## Automated tests:
- [X] Unit tests
- [ ] Integration tests

###### Reviewer Checklist:


shared/data/fingerprint-protection.js
```diff
@@ -68,6 +68,18 @@
...
+        'useragent': {
+            'userAgent': {
+                'object': 'navigator',
+                'origValue': navigator.userAgent,
+                'targetValue': `"${ua}"` // Defined in chrome-events.es6.js and injected as a variable
+            },
+            'appVersion': {
+                'object': 'navigator',
+                'origValue': navigator.appVersion,
+                'targetValue': /Firefox/i.test(navigator.userAgent) ? `"${navigator.appVersion}"` : `"${ua.replace('Mozilla/', '')}"`
```
kdzwinel, line 80: hm, I'm not sure if I understand what's going on here and why 🤔 Might be worth extracting to a seprate method with a comment?

shared/js/background/chrome-events.es6.js
```diff
@@ -217,6 +219,17 @@ chrome.webNavigation.onCommitted.addListener(details => {
             const whitelisted = settings.getSetting('whitelisted')
             const tabURL = new URL(details.url) || {}
             if (!whitelisted || !whitelisted[tabURL.hostname]) {
+                // Set variables, which are used in the fingerprint-protection script.
+                const variableScript = {
+                    'code': `
```
kdzwinel, line 224: is this injected to the page context or extension's "isolated world" i.e. will this variable be visible to the page? Ideally it shouldn't.

shared/data/constants.js
```diff
@@ -97,6 +97,20 @@ module.exports = {
...
+        {
+            "name": "agents",
+            "url": "https://staticcdn.duckduckgo.com/useragents/random_useragent.json",
+            "format": "json",
+            "source": "external"
+        },
+        {
+            "name": "excludeList",
+            "url": "https://staticcdn.duckduckgo.com/useragents/useragent_excludes.json",
+            "format": "text",
```
kdzwinel, line 110: why it's text and not json?

shared/js/background/chrome-events.es6.js
```diff
@@ -217,6 +219,17 @@ chrome.webNavigation.onCommitted.addListener(details => {
...
+                        try {
+                            var ua='${agentSpoofer.getAgent()}'
```
kdzwinel, line 226: 🦹‍♂️ So… what's going to happen if I visit DDG 1000s of times using spoofed UA `' eval("hello world!");'`? Will it make it into here? Should we sanitize this?

shared/data/fingerprint-protection.js
```diff
@@ -68,6 +68,18 @@
                 'targetValue': undefined
             }
         },
+        'useragent': {
+            'userAgent': {
+                'object': 'navigator',
+                'origValue': navigator.userAgent,
+                'targetValue': `"${ua}"` // Defined in chrome-events.es6.js and injected as a variable
```
kdzwinel, line 75: hm, why not `"targetValue": ua`? Do we really need those `"`?

shared/js/background/classes/agentspoofer.es6.js
```diff
@@ -0,0 +1,145 @@
...
+const tabManager = require('../tab-manager.es6')
+
+class AgentSpoofer {
+    constructor () {
+        this.realAgent = navigator.userAgent
+        this.parsedAgent = agentparser.lookup(this.realAgent)
+        this.spoofedAgent = this.realAgent
+        this.selectAgent()
+        this.needsRotation = true
+        this.tabs = {} // maintain cache of tab root URL's (chrome only)
```
kdzwinel, line 14: Is that something that we should merge into tabManager? 🤔

shared/js/background/classes/agentspoofer.es6.js
```diff
@@ -0,0 +1,145 @@
...
+    }
+
+    /**
+     * Tests whether the two URL's belong to the same
+     * first party set.
+     */
+    isFirstParty (url1, url2) {
+        const first = tldts.parse(url1).domain
+        const second = tldts.parse(url2).domain
+        return first === second ||
```
kdzwinel, line 119: isn't our definition of first party too broad here? Do we have to include entities? 🤔

unit-test/data/random_useragent.json
```diff
@@ -0,0 +1,6357 @@
+{
+    "linux-samsungbrowser": [
```
kdzwinel, line 2: is it still useful if there is only one entry here?

unit-test/data/random_useragent.json
```diff
@@ -0,0 +1,6357 @@
...
+        {
+            "percentage": "0.008%",
+            "total": "233",
```
kdzwinel, line 5: this doesn't seem to be used - I'd drop it

kdzwinel (commented):
I did a quick pass. Excited for this!

package.json
```diff
@@ -84,15 +88,17 @@
         "tabular-json": "0.1.3",
         "tldts": "^5.6.2",
         "underscore": "1.8.3",
+        "useragent": "^2.3.0",
         "web-ext": "^2.7.0",
-        "xvfb": "0.2.3"
+        "xvfb": "0.2.3",
+        "yamlparser": "0.0.2"
```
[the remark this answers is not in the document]
Charlie-belmer: It's true, we don't. Unfortunately, the useragent library requires it for some optional functionality, but doesn't include it in the dependency chain automatically.

shared/data/fingerprint-protection.js
Charlie-belmer: Thanks for noticing this - definitely it's confusing (even going back over now a few days after I wrote it I had to remind myself of a couple of things). I expanded it into a function with liberal comments.

shared/js/background/chrome-events.es6.js
Charlie-belmer: It will be visible to the page. Unfortunately, it's the only way I have found to pass variables in to the on page script before all other scripts load. I could potentially make the variable name more generic, and maybe obfuscate a few more things. Unfortunately, this runs into some of the same issues I hit with fingerprinting defense - any script we inject can potentially be observed by the page itself.

shared/data/constants.js
Charlie-belmer: great catch, forgot to update that when I changed the list format to json :)

shared/js/background/chrome-events.es6.js
Charlie-belmer: This is an awesome catch! An eval won't execute here, but a carefully crafted UA can get XSS on visited pages. So, if enough traffic was generated with a malicious UA (would need to be very significant), or a malicious userAgent file was served from our domain, this could be an issue. 

Unfortunately, UserAgents by RFC can contain any valid character, so it isn't as simple as a normal XSS filter.

What I did is strip out agent that contains single quotes in the UA, which I have not seen any valid user agents that include it, and is needed to break out of the assignment.

shared/data/fingerprint-protection.js
Charlie-belmer: Unfortunately, yes, though it's ugly. The eventual function it gets passed to will throw an error if a string value doesn't have quotes surrounding it, because it is part of an object (line 135)

unit-test/data/random_useragent.json
Charlie-belmer: Yeah this is all data just for testing, I updated this file.

shared/js/background/classes/agentspoofer.es6.js
Charlie-belmer: I don't know if we _have_ to, but it might increase breakage where entities are doing something. Maybe it doesn't matter - I'll run some tests without entities and see if I find any breakage.

shared/js/background/classes/agentspoofer.es6.js
Charlie-belmer: I should have been using tabs from the start for this - much simpler.

package.json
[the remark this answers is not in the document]
kdzwinel: 🤦‍♂️

shared/js/background/chrome-events.es6.js
kdzwinel: Sounds good! Maybe we could also do some filtering on the back-end if it's simple enough?

shared/js/background/chrome-events.es6.js
kdzwinel: Gotcha. Detectability is one thing, but I'm worried about name conflicts with variables defined by the page - let's make that variable name a bit longer/complicated.

unit-test/data/random_useragent.json
kdzwinel: oh yeah, but as discussed - lets also drop it from the real-thing™️

unit-test/background/fingerprint.es6.js
```diff
@@ -18,7 +20,8 @@ describe('User-Agent replacement', () => {
         agentStorage.family = agentSpoofer.parsedAgent.family
         agentStorage.os = agentSpoofer.parsedAgent.os.family
         agentStorage.processAgentList(agentData)
-        spy = spyOn(Tab, 'constructor')
+        tabSpy = spyOn(Tab, 'constructor')
```
kdzwinel, line 23: 👀 How about `tabObserver` 😅

Charlie-belmer: I updated anyway - yeah these are spys to replace the original function :)

shared/js/background/classes/agentspoofer.es6.js
jdorweiler: `findParent` will return the entity for a url if you do to compare entities https://github.com/duckduckgo/duckduckgo-privacy-extension/blob/develop/shared/js/background/utils.es6.js#L30

shared/js/background/chrome-events.es6.js
```diff
@@ -246,6 +279,10 @@ chrome.alarms.create('updateLists', { periodInMinutes: 30 })
 chrome.alarms.create('updateUninstallURL', { periodInMinutes: 10 })
 // remove expired HTTPS service entries
 chrome.alarms.create('clearExpiredHTTPSServiceCache', { periodInMinutes: 60 })
+// Update userAgent lists
+chrome.alarms.create('updateUserAgentData', { periodInMinutes: 12 * 60 })
+// Rotate the user agent spoofed
+chrome.alarms.create('rotateUserAgent', { periodInMinutes: 24 * 60 })
```
jdorweiler, line 285: Just double checking... It looks like this will rotate in two cases:
1. 24 hr alarm triggers
2. browser restarts and there's a new etag

Is that correct?

Charlie-belmer: That's right - new data or alarm trigger

PR #488, opened 2020-09-09
Enable fingerprint protection for all users

Charlie-belmer:
Removes the fingerprint protection experiment gates. The experiment did not show any increase in breakage or reported issues, so I'd like to enable this feature for all users.


kdzwinel (approved):
LGTM 🚢

PR #489, opened 2020-09-09
Improve css positioning in popup

GioSensation:

## Description:

Previously the order of elements in the popup was not guaranteed because they were slotted one after the other on a first-come-first-served basis and the underlying logic had network and other async calls. In order to maintain visual hierarchy, we were relying on `position: absolute` to explicitly set their position. This was hard to maintain and fragile.

With this PR, I have created hardcoded containers in the markup and render items inside them, regardless of when they become available. This allowed me to remove the fixed and absolute positioning and just let elements follow the normal flow, so that I could also remove the fixed heights of the body and other elements.

This PR also runs a minor review on elements heights to make them all more uniform in light of future developments.

## Steps to test this PR:
Just make sure that the popup is looking right. Click around and make sure that panels slide in as expected and everything sits right where it should.

###### Reviewer Checklist:

###### PR Author Checklist:

### DIFF
--- a/shared/html/popup.html
+++ b/shared/html/popup.html
@@ -26,8 +26,12 @@
 <body class="body--neutral">
 
     <div id="popup-container"
-         class="sliding-subview sliding-subview--root sliding-subview--has-fixed-header">
-         <div class="site-info site-info--placeholder card"></div>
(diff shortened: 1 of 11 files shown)


MariagraziaAlastra:
LGTM, thanks :+1:

PR #495, opened 2020-10-02
Global Privacy Control

dharb:
**Reviewer:** @kdzwinel 

**CC:** @MariagraziaAlastra 

## Description:
This PR adds the ability for users to set the proposed Global Privacy Control standard, which allows users to signal to websites their preference to:
1. Not sell your personal data
2. Limit sharing of your personal data to other companies.

## Steps to test this PR:
1. Install extension
2. Visit [test site](https://global-privacy-control.glitch.me/)
3. Header should not be present. DOM signal should be present and have value of `false`
4. Go to extension settings, turn on GPC
5. Reload test site
6. Header should be present and have value of `"1"`. DOM signal should be present and have value of `true`
7. Visit site with `script-src` in their CSP like https://twitter.com. Check that DOM signal is set.
8. Repeat steps for Firefox and Chrome.

## Automated tests:
- [ ] Unit tests
- [ ] Integration tests

###### Reviewer Checklist:

###### PR Author Checklist:


shared/js/background/GPC.es6.js
```diff
@@ -0,0 +1,57 @@
...
+
+// Add navigator.globalPrivacyControl property to all frames if setting enabled
+// The script injected in Firefox is different from the script injected in Chromium-
+// based browsers due to slight differences in how the browsers interpret script-src
+// CSP directives, and the existence of Xray vision in Firefox.
+function injectDOMSignal (tabId) {
+    const supportedBrowsers = ['chrome', 'moz', 'edg']
+    let browserName = utils.getBrowserName()
+
+    if (!supportedBrowsers.includes(browserName)) return
```
kdzwinel, line 28: What browsers are unsupported? Maybe we should use chromium script as a default and only make an exception for 'moz'?

shared/js/background/chrome-events.es6.js
```diff
@@ -262,6 +262,31 @@ chrome.webRequest.onBeforeSendHeaders.addListener(
...
+        if (GPCHeader) {
+            let requestHeaders = request.requestHeaders
+            requestHeaders.push(GPCHeader)
+            return {requestHeaders: requestHeaders}
+        }
+    },
+    {urls: ['<all_urls>']},
+    ['blocking', 'requestHeaders']
+)
+
```
kdzwinel, line 289: for majority of users those two listeners will do nothing, it'd be better to wrap them in `if (settings.getSetting('GPCEnabled')) {` but I understand that there is no clean way for us to start/stop those listeners whenever the setting changes?

shared/js/content-scripts/GPC-chromium.js
```diff
@@ -0,0 +1,18 @@
...
+(function setDOMSignal () {
+    const scriptString = `
+        // Catch errors if signal is already set by user agent or other extension
+        try {
+            Object.defineProperty(navigator, "globalPrivacyControl", {
+                value: ${globalPrivacyControlValue},
+                enumerable: true
+            });
+            // Remove script tag after execution
+            document.currentScript.parentElement.removeChild(document.currentScript);
```
kdzwinel, line 11: nice 👍

shared/js/content-scripts/GPC-moz.js
```diff
@@ -0,0 +1,12 @@
+// Set Global Privacy Control property on DOM
+(function setDOMSignal () {
+    let wrappedNavigator = navigator.wrappedJSObject;
```
kdzwinel, line 3: nit
```suggestion
    const wrappedNavigator = navigator.wrappedJSObject;
```

shared/scss/options.scss
```diff
@@ -19,3 +19,13 @@
     margin-bottom: 10px;
     color: $slate;
 }
+
+.options-content__privacy {
```
kdzwinel, line 23: what's up with all those css changes 🤔 are those related to this task?

shared/scss/views/_whitelist.scss
```diff
@@ -4,11 +4,11 @@
     font-weight: 600;
 
     .menu-title {
-        padding-bottom: 4px;
+        padding: 20px 0 4px;
```
kdzwinel, line 7: ditto - those seem unrelated

kdzwinel (commented):
Nice work @dharb ! That `wrappedJSObject` stuff in firefox is very neat, wish we had that in Chrome.

I left couple of comments, no show stoppers though.

kdzwinel:
(also - build failures are due to linting issues)

shared/js/background/GPC.es6.js
dharb: TBH I'm not sure anything is unsupported, I put this in here more to prevent it from blowing anything up if for some reason `utils.getBrowserName()` returns an empty string (see https://github.com/duckduckgo/duckduckgo-privacy-extension/blob/develop/shared/js/shared-utils/parse-user-agent-string.es6.js#L19).

I'm happy to make chromium the default though, that makes sense to me (since only the FF content script has the potential to break things due to FF-specific code).

shared/scss/options.scss
dharb: Yeah, it's for the settings page. There are some inconsistencies in the spacing there where the actual design didn't look at all like the mock up, so I tweaked a few things outside of the GPC section for consistency. There are plans to revamp this whole page, so there will be more seemingly unrelated css changes incoming shortly.

shared/js/background/chrome-events.es6.js
dharb: Definitely the case for the header, but to match the most recent spec we have to inject a DOM signal regardless of whether setting is on or not--`navigator.globalPrivacyControl` will be true if enabled, and false otherwise.

I guess we could potentially wrap the header listener in an if statement, then add some logic to the settings toggle to add/remove the listener on change. What do you think, worth doing?

shared/js/background/GPC.es6.js
kdzwinel: Yep, I think we should go ahead with chromium script being the default.

shared/js/background/chrome-events.es6.js
kdzwinel: Ah good point about the DOM prop. Yeah, going after the header is probably not worth the effort.

dharb:
@kdzwinel it appears the settings page style tweaks are going to be more than just tweaks. Let's merge this one now and do them in a different PR.

kdzwinel:
Great job 🙌

PR #497, opened 2020-10-16
Fix issue with processing User Agent Safe List

Charlie-belmer:
This fixes an issue where if the user agent safe list is longer than a single entry, it won't correctly apply the safe listing.


jdorweiler:
👍

Issue #498, opened 2020-10-17
Extension causes Firefox to ignore size in window.open()


hkockerbeck:
## Description
When I open a new window with the JavaScript function `window.open()`, I can set the width and height the new window should have. The extension seems to cause Firefox to ignore those parameters.

## Steps to Reproduce
1. Deactivate the extension
2. Click on a link like `<a href="javascript:window.open('https://www.duckduckgo.com', 'name_of_window', 'height=500,width=500')">Test link</a>`. The new window should be of the specified size.
3. Activate the extension
4. Click on the link again. The new window probably is not of the specified size, but uses all available width and height on the display (but isn't maximized, either). 

**Expected behavior:**
I expect Firefox with the extension enabled to honor the specified width and height of new windows.

**Actual behavior:**
The extension seems to filter out the `width` and `height` parameters at least. I didn't test other window features that `window.open()` supports in the third argument.

**Versions**
- Extension: 2020.10.13
- Browser: Firefox
- OS: Linux

## Additional Information
The issue was probably introduced in the latest version of the extension, or at least not too long ago. I've been using the extension for some time now, and the problem only appeared quite recently.

Charlie-belmer:
I tested this in the latest release and it appears resolved.

PR #499, opened 2020-10-20
Issue 498

Charlie-belmer:
This fixes Issue #498. The problem was that on Linux, Firefox would resize the window.

I also increased the UA list fetch frequency and made it leverage etags.

### DIFF
--- a/shared/js/background/chrome-events.es6.js
+++ b/shared/js/background/chrome-events.es6.js
@@ -213,30 +213,35 @@ chrome.webNavigation.onCommitted.addListener(details => {
     const whitelisted = settings.getSetting('whitelisted')
     const tabURL = new URL(details.url) || {}
     let tab = tabManager.get({ tabId: details.tabId })
-    if (tab.site.isBroken) {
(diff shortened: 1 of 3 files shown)


shared/js/background/chrome-events.es6.js
```diff
@@ -237,6 +243,12 @@ chrome.webNavigation.onCommitted.addListener(details => {
 // Replace UserAgent header on third party requests.
 chrome.webRequest.onBeforeSendHeaders.addListener(
     function spoofUserAgentHeader (e) {
+        let tab = tabManager.get({ tabId: e.tabId })
+        if (tab.site.isBroken) {
```
dharb, line 247: Nice, good call here.

dharb: Actually, mind checking that tab is defined here? And in the one above.
`if (tab && tab.site.isBroken) {`
Sorry, I just realized this was throwing some errors on the initial extension load.

dharb:
Thanks for adding the extra checks! 👍

PR #501, opened 2020-10-20
Fix iFrame cross origin console errors

Charlie-belmer:
Fix occasional console error where cross domain iFrame gives console errors on anti-fingerprinting script


dharb (approved):
👍

Issue #503, opened 2020-10-21
Firefox XML view: JavaScript and other output


ActionScripted:
## Description

Firefox XML view broken; contains JavaScript and other extraneous output.

## Steps to Reproduce

1. Open Firefox
2. Add/enable this extension
3. View an XML feed or file

**Expected behavior:**

Standard XML browser.

**Actual behavior:**

Mangled output with JavaScript.

**Versions**

- Extension: 2020.10.13
- Browsers: Firefox 82.0 (64-bit) and Firefox 83.0b2 (64-bit)
- OS: macOS

## Additional Information

Screenshot attached showing expected (left; plugin disabled) versus what happens (right; plugin enabled). Have tested this with all other plugins disabled.

![ddg-privacy-bug](https://user-images.githubusercontent.com/708612/96741556-ec430b80-138f-11eb-91d6-455435b8603b.png)

ActionScripted:
This is an issue even when the URL/file ends with `*.xml` e.g., https://feedforall.com/sample.xml

Charlie-belmer:
Thanks for reporting!

This looks like it's a result of the anti-fingerprinting features we released recently. I just made a PR which should fix this by filtering out xml and json content types. 

If you have some sites I can test with, I would appreciate it, otherwise this should go out in a future release.

ActionScripted:
I don't have any specific sites but I've been noticing it on anything with an RSS/XML feed. Noticed it first when setting up our new company site/blog and testing Gatbsy's RSS generation (via https://merit-network.github.io/rss.xml).

Thanks so much!

vpArth:
What is date for this merge(fix xml/json data) to be released?

vpArth:
Also, is there any chance to have a gui switch to turn off that antifingerprint injection feature?

Charlie-belmer:
Unfortunately there isn't a disable toggle at this time for that feature. 

Though I don't have an exact date yet, this fix will be included in the next feature release, which is likely to be in the next couple of weeks.

I know it's not an ideal solution, but until then, disabling the site protections when you want to use the XML view should work.

Thanks!

vpArth:
Main issue for me is not just direct xml view. It injects into ajax loaded xml-data on some services I use and leaks into gui =)

![][1]

[1]: https://imgur.com/fKGStLE.png

Charlie-belmer:
[...]

PR #504, opened 2020-10-21
Stop injecting anti-fingerprint protections into xml and json documents

Charlie-belmer:
This adds some content checks into the anti-fingerprinting script injection, and will not inject scripts if the content type is json or xml. 

Fixes issue #503

### DIFF
--- a/shared/data/fingerprint-protection.js
+++ b/shared/data/fingerprint-protection.js
@@ -5,6 +5,23 @@
  */
 
 (function protect () {
+    // Exclude some content types from injection
+    const elem = document.head || document.documentElement
+    try {
+        const contentType = elem.ownerDocument.contentType
+        if (contentType === 'application/xml' ||
+            contentType === 'application/json' ||
+            contentType === 'text/xml' ||
+            contentType === 'text/json' ||
+            contentType === 'text/rss+xml' ||
+            contentType === 'application/rss+xml'
+        ) {
+            return
+        }
+    } catch (e) {
+        // if we can't find content type, go ahead with injection.
+    }
+
     // Property values to be set and their original values.
     const fingerprintPropertyValues = {
         'screen': {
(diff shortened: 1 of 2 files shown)


dharb (commented):
This looks good, nice fix! I just added the same fix for our GPC injection to this same PR.

shared/data/fingerprint-protection.js
```diff
@@ -283,6 +283,17 @@
...
+                const contentType = elem.ownerDocument.contentType
+                if (contentType === 'application/xml' ||
+                    contentType === 'application/json' ||
+                    contentType === 'text/xml' ||
+                    contentType === 'text/json') {
+                    return
+                }
+            } catch (e) {
+                // if we can't find content type, go ahead with injection.
+            }
```
kdzwinel, line 296: instead of doing this check here, I think it'd be better to do it couple of lines below and avoid creating a window resize listener and the injection script in the first place e.g.

```js
if (!isXMLorJSON()) {
    window.addEventListener('resize', function () {
        const windowScript = setWindowDimensions()
        inject(windowScript, true)
    })

    const injectionScript = buildInjectionScript()
    inject(injectionScript)
}
```

or do it at the very beginning of the script and just `return` early like Dave did below.

kdzwinel (commented):
I left one comment. Also, please test if we are not breaking the PDF reader on any of the browsers 🙏

shared/data/fingerprint-protection.js
Charlie-belmer: Good thinking, made the changes.

Charlie-belmer:
I updated and also added a couple of other content types for RSS feeds which might have display issues. The larger thing to think about here is how much of this content type selectivity we want - there are many potential content types for structured data which the browser may treat in special ways: https://www.iana.org/assignments/media-types/media-types.xhtml

It seems this change will work for the most common use cases, but it's something we may want to think about.

dharb (approved):
👍

PR #511, opened 2020-10-29
Referrer Truncation

Charlie-belmer:
**Reviewer: Konrad

## Description:
This adds referrer truncation features to the extension, similar to what Safari has implemented. This will help limit information leaks while browsing the web.

## Automated tests:
- [X] Unit tests
- [ ] Integration tests

###### Reviewer Checklist:

###### PR Author Checklist:

### DIFF
--- a/shared/data/fingerprint-protection.js
+++ b/shared/data/fingerprint-protection.js
@@ -23,7 +23,7 @@
     }
 
     // Property values to be set and their original values.
-    const fingerprintPropertyValues = {
+    let fingerprintPropertyValues = {
         'screen': {
             'availTop': {
                 'object': 'screen',
@@ -103,12 +103,19 @@
                 'origValue': navigator.doNotTrack,
                 'targetValue': /Firefox/i.test(navigator.userAgent) ? '"unspecified"' : null
             }
-        },
-        'document': {
+        }
+    }
+
+    // ddg_referrer is defined in chrome-events.es6.js and injected as a variable if referrer should be modified
+    if (ddg_referrer &&
+        ddg_referrer !== 'undefined' &&
+        document.referrer &&
+        document.referrer !== '') {
+        fingerprintPropertyValues['document'] = {
             'referrer': {
                 'object': 'document',
                 'origValue': document.referrer,
-                'targetValue': `"${ddg_referrer || document.referrer}"` // Defined in chrome-events.es6.js and injected as a variable
+                'targetValue': `"${ddg_referrer || document.referrer}"` 
             }
         }
     }
--- a/shared/data/fingerprint-protection.js
+++ b/shared/data/fingerprint-protection.js
@@ -69,11 +69,11 @@
             }
         },
         'useragent': {
-            'userAgent': {
-                'object': 'navigator',
-                'origValue': navigator.userAgent,
-                'targetValue': `"${ddg_ext_ua}"` // Defined in chrome-events.es6.js and injected as a variable
-            },
+//            'userAgent': {
+//                'object': 'navigator',
+//                'origValue': navigator.userAgent,
+//                'targetValue': `"${ddg_ext_ua}"` // Defined in chrome-events.es6.js and injected as a variable
+//            },
             'appVersion': {
                 'object': 'navigator',
                 'origValue': navigator.appVersion,
@@ -139,7 +139,10 @@
             for (const [name, prop] of Object.entries(fingerprintPropertyValues[category])) {
                 // Don't update if existing value is undefined or null
                 if (!(prop.origValue === undefined)) {
-                    script += `Object.defineProperty(${prop.object}, "${name}", { value: ${prop.targetValue} });\n`
+                    script += `try {
+                        Object.defineProperty(${prop.object}, "${name}", { value: ${prop.targetValue} });
+                    } catch (e) {}
+                    `
                 }
             }
         }
@@ -161,12 +164,14 @@
             for (const prop of battery.properties) {
                 // Prevent setting events via event handlers
                 batteryScript += `
+                try {
                     Object.defineProperty(battery, '${prop}', {
                         enumerable: true,
                         configurable: false,
                         writable: false,
                         value: undefined
                     })
+                } catch (e) {}
                 `
             }
 
@@ -214,15 +219,12 @@
         return value
     }
 
-    function setWindowPropertyValue (property, value, catchErrors=false) {
-        let script = `window.${property} = ${value}\n`
-        if (catchErrors) {
-            script = `
-                try {
-                    ${script}
-                } catch (e) {}
-            `
-        }
+    function setWindowPropertyValue (property, value) {
+        let script = `
+            try {
+                window.${property} = ${value}
+            } catch (e) { }
+        `
         return script
     }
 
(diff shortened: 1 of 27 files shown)


shared/js/background/utils.es6.js
```diff
@@ -17,6 +18,29 @@ function extractHostFromURL (url, shouldKeepWWW) {
     return hostname
 }
 
+// Removes information from a URL, such as path, user information, and optionally sub domains
+function extractLimitedDomainFromURL (url, keepSubdomains) {
+    if (!url) return ''
```
kdzwinel, line 23: In most cases we return undefined but here we are returning empty string - is there a difference? should we use the same thing for consistency?

shared/js/background/chrome-events.es6.js
```diff
@@ -271,6 +274,54 @@ chrome.webRequest.onBeforeSendHeaders.addListener(
...
+    function limitReferrerData (e) {
+        let referrer = e.requestHeaders.filter(header => header.name.toLowerCase() === 'referer')[0] || ''
+        if (referrer) {
+            referrer = referrer.value
+        } else {
+            return
+        }
+
+        // Check if origin is safe listed
+        const tab = tabManager.get({ tabId: e.tabId })
```
kdzwinel, line 304: we are already doing all the checks inside of `truncateReferrer` so I wonder if we really need this one?

shared/js/background/chrome-events.es6.js
```diff
@@ -271,6 +274,54 @@ chrome.webRequest.onBeforeSendHeaders.addListener(
...
+ *   Don't modify the header when:
+ *   - If the header is blank, it will not be modified.
+ *   - If the referrer domain OR request domain are safe listed, the header will not be modified
+ *   - If the referrer domain and request domain are part of the same entity (as defined in our
+ *     entities file for first party sets), the header will not be modified.
+ *
+ *   Modify the header when:
+ *   - If the destination is in our tracker list, we will trim it to eTLD+1 (remove path and subdomain information)
+ *   - In all other cases (the general case), the header will be modified to only the referrer origin (includes subdomain).
+ */
```
kdzwinel, line 288: Nice description 👏

shared/js/background/tab-manager.es6.js
```diff
@@ -116,6 +116,7 @@ class TabManager {
         let tab = tabManager.get({tabId: request.tabId})
 
         if (tab) {
+            tab.initiator = request.initiator
```
kdzwinel, line 119: initiator is the url of the previous page?

shared/js/background/utils.es6.js
```diff
@@ -114,6 +138,15 @@ function getAsyncBlockingSupport () {
     return false
 }
 
+// return true if the given url is in the safelist. For checking if the current tab is in the safelist,
+// tabManager.site.whitelisted is the preferred method.
+function isSafeListed (url) {
+    const hostname = extractHostFromURL(url)
+    const tld = tldts.parse(url).domain
+    const safeList = settings.getSetting('whitelisted')
+    return safeList && (safeList[hostname] || safeList[tld])
```
kdzwinel, line 147: should we also check stuff in between hostname and tld? like for a.b.c.example.com hostname is a.b.c.example.com and tld is example.com, but I think c.example.com can be safelisted?

unit-test/background/tracker-utils-tests.js
```diff
@@ -0,0 +1,275 @@
...
+            referrer: 'http://subdomain.siteA.com/article/1',
+            target: 'https://google-analytics.com/some/path',
+            expectedReferrer: 'http://sitea.com'
+        },
+        {
+            name: 'target is not a tracker, referrer should keep subdomain',
+            referrer: 'http://subdomain.siteA.com/article/1',
+            target: 'http://siteB.comm',
+            expectedReferrer: 'http://subdomain.sitea.com'
+        }
```
kdzwinel, line 196: lets also test stuff with multiple subdomain "levels" (a.b.c.example.com) and that "www." thing

kdzwinel (commented):
Looks really good! I left a bunch of small comments, but nothing critical as far as I can tell.

shared/js/background/chrome-events.es6.js
Charlie-belmer: I think the only check that isn't done in the truncateReferrer is against the broken site list. It could probably be done in truncateReferrer as well if we'd rather put everything there, but it would need to be refactored to accept a tab parameter.

shared/js/background/chrome-events.es6.js
```diff
@@ -249,7 +252,7 @@ chrome.webNavigation.onCommitted.addListener(details => {
 chrome.webRequest.onBeforeSendHeaders.addListener(
     function spoofUserAgentHeader (e) {
         let tab = tabManager.get({ tabId: e.tabId })
-        if (tab && tab.site.isBroken) {
+        if (!!tab && (tab.site.whitelisted || tab.site.isBroken)) {
```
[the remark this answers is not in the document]
Charlie-belmer: Yeah, this just brings it in line with other check methods and cleans it up a little.

shared/js/background/chrome-events.es6.js
```diff
@@ -271,6 +274,54 @@ chrome.webRequest.onBeforeSendHeaders.addListener(
...
+ *   - In all other cases (the general case), the header will be modified to only the referrer origin (includes subdomain).
+ */
+let ReferrerListenerOptions = ['blocking', 'requestHeaders']
+if (browser !== 'moz') {
+    ReferrerListenerOptions.push('extraHeaders') // Required in chrome type browsers to receive referrer information
+}
+
+chrome.webRequest.onBeforeSendHeaders.addListener(
+    function limitReferrerData (e) {
+        let referrer = e.requestHeaders.filter(header => header.name.toLowerCase() === 'referer')[0] || ''
```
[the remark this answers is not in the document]
Charlie-belmer: Nice find, thanks!

shared/js/background/utils.es6.js
```diff
@@ -17,6 +18,29 @@ function extractHostFromURL (url, shouldKeepWWW) {
...
+    if (!url) return ''
+    try {
+        const parsedURL = new URL(url)
+        const tld = tldts.parse(url)
+        let finalURL = tld.domain
+        if (!parsedURL || !tld) return ''
+        if (keepSubdomains) {
+            if (tld.subdomain) {
+                finalURL = tld.subdomain + '.' + tld.domain
+            }
```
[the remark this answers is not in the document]
Charlie-belmer: Another good simplification!

shared/js/background/utils.es6.js
```diff
@@ -17,6 +18,29 @@ function extractHostFromURL (url, shouldKeepWWW) {
...
+        } else if (tld.subdomain && tld.subdomain.toLowerCase() === 'www') {
+            finalURL = 'www.' + tld.domain
```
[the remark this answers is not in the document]
Charlie-belmer: Added a comment here - basically subdomain would be exactly 'www', so shouldn't match your example, and some sites require it, so stripping it off could cause issues.

shared/js/background/tracker-utils.js
```diff
@@ -0,0 +1,81 @@
+// Utility functions for dealing with tracker information
+const utils = require('./utils.es6')
+const trackers = require('./trackers.es6')
+const tldts = require('tldts')
+const tdsStorage = require('./storage/tds.es6')
+
+// Determine if two URL's belong to the same entity.
+function isSameEntity (url1, url2) {
```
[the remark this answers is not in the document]
Charlie-belmer: I agree. It's here because putting it into utils creates a circular dependency, utils is a dependency of trackers, and trackers is a dependency for this function.

shared/js/background/tracker-utils.js
```diff
@@ -0,0 +1,81 @@
...
+    if (utils.isSafeListed(referrer) || utils.isSafeListed(target)) {
+        return undefined
+    }
+
+    if (isSameEntity(referrer, target)) {
+        return undefined
+    }
+
+    if (tdsStorage.ReferrerExcludeList && tdsStorage.ReferrerExcludeList.excludedReferrers) {
+        const excludedDomains = tdsStorage.ReferrerExcludeList.excludedReferrers.map(e => e.domain)
```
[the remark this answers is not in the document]
Charlie-belmer: The exclude list will be public, so the format has space for human information: 
```
[
         {
              "domain": "example.com",
              "reason": "this domain won't login without the correct referrer"
          }
]
```

shared/data/fingerprint-protection.js
```diff
@@ -103,6 +103,13 @@
                 'origValue': navigator.doNotTrack,
                 'targetValue': /Firefox/i.test(navigator.userAgent) ? '"unspecified"' : null
             }
+        },
+        'document': {
+            'referrer': {
+                'object': 'document',
+                'origValue': document.referrer,
+                'targetValue': `"${ddg_referrer || document.referrer}"` // Defined in chrome-events.es6.js and injected as a variable
```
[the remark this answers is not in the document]
Charlie-belmer: I added several checks to determine whether to do this - see what you think.

shared/js/background/tracker-utils.js
```diff
@@ -0,0 +1,81 @@
...
+        if (entity1 === undefined && entity2 === undefined) return false
+        return entity1 === entity2
+    } catch (TypeError) {
+        // tried to parse invalid URL
+        return false
+    }
+}
+
+// return true if URL is in our tracker list
+function isTracker (url) {
```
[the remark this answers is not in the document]
Charlie-belmer: This is another one that would create a circular dependency unfortunately.

shared/js/background/tab-manager.es6.js
Charlie-belmer: I updated this to attach referrer directly to the tab - passing it around isn't perfect, and the initiator was not always what I wanted here.

Charlie-belmer:
Thanks for all the great thoughts and feedback Konrad! I should have updated everything - let me know how it looks.

shared/data/fingerprint-protection.js
```diff
@@ -106,6 +106,20 @@
         }
     }
 
+    // ddg_referrer is defined in chrome-events.es6.js and injected as a variable if referrer should be modified
+    if (ddg_referrer &&
+        ddg_referrer !== 'undefined' &&
```
kdzwinel, line 111: undefined as a string? This can happen?

shared/data/fingerprint-protection.js
```diff
@@ -23,7 +23,7 @@
     }
 
     // Property values to be set and their original values.
-    const fingerprintPropertyValues = {
+    let fingerprintPropertyValues = {
```
kdzwinel, line 26: nit - if we are not overwriting whole `fingerprintPropertyValues` with a new object (and just adding a property to existing object), this should still be const.

shared/data/fingerprint-protection.js
```diff
@@ -106,6 +106,20 @@
         }
     }
 
+    // ddg_referrer is defined in chrome-events.es6.js and injected as a variable if referrer should be modified
```
kdzwinel, line 109: nice 👏

shared/js/background/chrome-events.es6.js
```diff
@@ -271,6 +274,55 @@ chrome.webRequest.onBeforeSendHeaders.addListener(
...
+ *   - If the header is blank, it will not be modified.
+ *   - If the referrer domain OR request domain are safe listed, the header will not be modified
+ *   - If the referrer domain and request domain are part of the same entity (as defined in our
+ *     entities file for first party sets), the header will not be modified.
+ *
+ *   Modify the header when:
+ *   - If the destination is in our tracker list, we will trim it to eTLD+1 (remove path and subdomain information)
+ *   - In all other cases (the general case), the header will be modified to only the referrer origin (includes subdomain).
+ */
+let ReferrerListenerOptions = ['blocking', 'requestHeaders']
```
kdzwinel, line 289: nit - this should be in the linter, but I'd avoid capitalizing variables that are not classes

shared/js/background/tracker-utils.js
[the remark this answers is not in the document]
kdzwinel: Ah, I see, that makes sense.

shared/js/background/utils.es6.js
```diff
@@ -114,6 +140,22 @@ function getAsyncBlockingSupport () {
...
+    const tld = tldts.parse(url).domain
+    const safeList = settings.getSetting('whitelisted')
+    let subdomains = hostname.split('.')
+    while (subdomains.length > 1) {
+        if (safeList && safeList[subdomains.join('.')]) {
+            return true
+        }
+        subdomains.shift()
+    }
+    return safeList && (safeList[hostname] || safeList[tld])
```
kdzwinel, line 156: so now that you have that `while` loop, I think we can change this last line to `return false;`? In fact, you probably don't need `tldts.parse` anymore either.

shared/js/background/chrome-events.es6.js
kdzwinel: Hm, yeah, passing tab to `truncateReferrer` doesn't seem right. Should we update `isSafeListed` instead to check through all safelists?

kdzwinel (commented):
Thank you for addressing my comments and all those explanations! I left a few more final comments and played a bit with the extension (haven't seen any bugs).

shared/data/fingerprint-protection.js
Charlie-belmer: In the original case it actually was a string, because of how it was being inserted and read. The whole things is moved around now and is properly undefined (along with several other changes here).

Charlie-belmer:
Thanks Konrad, I refactored a few things, and shifted around some of how the injection works to deal with the iFrame issue we discussed. Please take a look and let me know your thoughts. Thanks!

shared/js/background/utils.es6.js
```diff
@@ -114,6 +140,45 @@ function getAsyncBlockingSupport () {
...
+    // Check user safe list
+    while (subdomains.length > 1) {
+        if (safeList && safeList[subdomains.join('.')]) {
+            return true
+        }
+        subdomains.shift()
+    }
+
+    // Check broken sites
+    if (isBroken(hostname)) {
```
kdzwinel, line 175: it looks like `isBroken` takes full url and not a hostname. Not sure if that matters.

Charlie-belmer: That's fine - it only needs the hostname, and if it gets a full URL it drops the rest, so it should be fine here. Thanks for everything!

kdzwinel (approved):

PR #513, opened 2020-11-11
Break-fix: Disable UserAgent spoofing & fix issue with Temp Storage FP protection

Charlie-belmer:
Brave generally doesn't include a special UA, so JS agent parsing libraries don't correctly identify Brave user agent strings. It looks like there was a logic error in how excluded regex patterns were being matched, which this should fix, so we can also ignore certain Brave strings are are DDG internal only.

**Reviewer:**

## Automated tests:
- [ ] Unit tests
- [ ] Integration tests

###### Reviewer Checklist:

###### PR Author Checklist:

### DIFF
--- a/shared/data/fingerprint-protection.js
+++ b/shared/data/fingerprint-protection.js
@@ -290,14 +290,14 @@
         const script = `
             if (navigator.webkitTemporaryStorage) {
                 try {
-                    navigator.webkitTemporaryStorage.${randomFunctionName} = navigator.webkitTemporaryStorage.queryUsageAndQuota
+                    ${randomFunctionName} = navigator.webkitTemporaryStorage.queryUsageAndQuota
                     navigator.webkitTemporaryStorage.queryUsageAndQuota = function queryUsageAndQuota (callback, err) {
                         const modifiedCallback = function (usedBytes, grantedBytes) {
                             const maxBytesGranted = 4 * 1024 * 1024 * 1024
-                            grantedBytes = Math.min(grantedBytes, maxBytesGranted)
+                            const grantedBytes = Math.min(grantedBytes, maxBytesGranted)
                             callback(usedBytes, grantedBytes)
                         }
-                        navigator.webkitTemporaryStorage.${randomFunctionName}(modifiedCallback, err)
+                        ${randomFunctionName}.call(navigator.webkitTemporaryStorage, modifiedCallback, err)
                     }
                 }
                 catch(e) {}
--- a/shared/data/fingerprint-protection.js
+++ b/shared/data/fingerprint-protection.js
@@ -292,7 +292,7 @@
         const script = `
             if (navigator.webkitTemporaryStorage) {
                 try {
-                    ${randomFunctionName} = navigator.webkitTemporaryStorage.queryUsageAndQuota
+                    const org = navigator.webkitTemporaryStorage.queryUsageAndQuota
                     navigator.webkitTemporaryStorage.queryUsageAndQuota = function queryUsageAndQuota (callback, err) {
                         const modifiedCallback = function (usedBytes, grantedBytes) {
                             const maxBytesGranted = 4 * 1024 * 1024 * 1024
--- a/shared/data/fingerprint-protection.js
+++ b/shared/data/fingerprint-protection.js
@@ -73,18 +73,6 @@
                 'targetValue': 8
             }
         },
-        'storage': {
-            'webkitTemporaryStorage': {
-                'object': 'navigator',
-                'origValue': navigator.webkitTemporaryStorage,
-                'targetValue': undefined
-            },
-            'webkitPersistentStorage': {
-                'object': 'navigator',
-                'origValue': navigator.webkitPersistentStorage,
-                'targetValue': undefined
-            }
-        },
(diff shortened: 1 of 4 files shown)


shared/data/fingerprint-protection.js
```diff
@@ -291,6 +280,32 @@
...
+     * Temporary storage can be used to determine hard disk usage and size.
+     * This will limit the max storage to 4GB without completely disabling the
+     * feature.
+     */
+    function modifyTemporaryStorage () {
+        const randomFunctionName = 'A' + Math.random().toString(36).substring(2) // random string of alphanumeric chars
+        const script = `
+            if (navigator.webkitTemporaryStorage) {
+                try {
+                    navigator.webkitTemporaryStorage.${randomFunctionName} = navigator.webkitTemporaryStorage.queryUsageAndQuota
```
kdzwinel, line 293: hm, why are we storing it as a property and not a variable in this scope? Sth like:

```js
                    const org = navigator.webkitTemporaryStorage.queryUsageAndQuota
///…
                    org.call(navigator.webkitTemporaryStorage, modifiedCallback, err)
```

Charlie-belmer: Good catch Konrad, should have used call from the start, updated.

shared/data/fingerprint-protection.js
```diff
@@ -291,6 +280,32 @@
...
+    function modifyTemporaryStorage () {
+        const randomFunctionName = 'A' + Math.random().toString(36).substring(2) // random string of alphanumeric chars
+        const script = `
+            if (navigator.webkitTemporaryStorage) {
+                try {
+                    ${randomFunctionName} = navigator.webkitTemporaryStorage.queryUsageAndQuota
+                    navigator.webkitTemporaryStorage.queryUsageAndQuota = function queryUsageAndQuota (callback, err) {
+                        const modifiedCallback = function (usedBytes, grantedBytes) {
+                            const maxBytesGranted = 4 * 1024 * 1024 * 1024
+                            const grantedBytes = Math.min(grantedBytes, maxBytesGranted)
```
dharb, line 297: I think we're redeclaring `grantedBytes` here which is throwing some console errors.

dharb (changes requested):
UA change looks good, just one minor issue with the webkitTemporaryStorage solution then this should be gtg.

shared/data/fingerprint-protection.js
dharb: Hm. I might have spoken too soon, I'm actually not seeing the permission dialog prompt for temporary storage on wevideo.com once I changed the var name of `grantedBytes`.

Charlie-belmer: You were right here - var needed renaming. Updated

Charlie-belmer:
Overall, I also disabled UA spoofing globally. There were some issues with Brave and Edge changing UA per request, which doesn't match the design of this feature.

The plan for now is to disable UA spoofing, and spin up a new project to rework it to be spoofed per request instead.

shared/data/fingerprint-protection.js
```diff
@@ -73,30 +73,20 @@
                 'targetValue': 8
             }
         },
-        'storage': {
-            'webkitTemporaryStorage': {
-                'object': 'navigator',
-                'origValue': navigator.webkitTemporaryStorage,
-                'targetValue': undefined
-            },
-            'webkitPersistentStorage': {
```
kdzwinel, line 82: I'm wondering why are we changing `webkitTemporaryStorage`? and why are we not overwriting `webkitPersistentStorage ` anymore? Description doesn't mention this.

shared/data/fingerprint-protection.js
```diff
@@ -291,6 +282,32 @@
...
+     * Temporary storage can be used to determine hard disk usage and size.
+     * This will limit the max storage to 4GB without completely disabling the
+     * feature.
+     */
+    function modifyTemporaryStorage () {
+        const randomFunctionName = 'A' + Math.random().toString(36).substring(2) // random string of alphanumeric chars
+        const script = `
+            if (navigator.webkitTemporaryStorage) {
+                try {
+                    ${randomFunctionName} = navigator.webkitTemporaryStorage.queryUsageAndQuota
```
kdzwinel, line 295: I'm still unsure why we need this random name 🤔 If we did:

```suggestion
                    const org = navigator.webkitTemporaryStorage.queryUsageAndQuota
```

no one, outside of the try/catch block, would have access to that `org` variable and we could still use it inside `queryUsageAndQuota`.

shared/data/fingerprint-protection.js
Charlie-belmer: Removing temp storage was causing breakage on some sites that use it (it's a Chrome only feature, and that site would tell users that it only works on chrome...)

On further reviewing during troubleshooting, persistent storage always returns 0, and prompts for permission to change it, so removing it seemed a safer option since I don't think it can be used to FP without permission.

kdzwinel: Ah, good find re webkitPersistentStorage, I didn't know that.

kdzwinel (approved):
LGTM

PR #520, opened 2020-12-03
Unit test repair

Charlie-belmer:
Attempting to fix the integration tests that frequently fail during the CI/CD test process. This upgrades puppeteer and associated API's, and fixes the tests that commonly fail.

Note: No review needed here until tests consistently pass.


integration-test/background/atb.js
```diff
@@ -1,7 +1,7 @@
 /* global dbg:false */
 const harness = require('../helpers/harness')
 const wait = require('../helpers/wait')
-const request = require('request-promise-native')
```
kdzwinel, line 4: just out of curiosity, what was wrong with this library?

integration-test/background/atb.js
```diff
@@ -137,27 +146,35 @@ describe('search workflow', () => {
...
     })
     it('should not update set_atb if a repeat search is made on the same day', async () => {
         // set set_atb to today's version
         await bgPage.evaluate((todaysAtb) => dbg.settings.updateSetting('set_atb', todaysAtb), todaysAtb)
 
         // run a search
         const searchPage = await browser.newPage()
-        searchPage.goto('https://duckduckgo.com/?q=test')
-
-        await bgPage.waitForResponse(res => res.url().match(/atb\.js/))
```
kdzwinel, line 159: this part was not moved over - not 100% sure if we need it, but I guess it's checking if atb.js has loaded?

integration-test/background/atb.js
```diff
@@ -167,13 +184,15 @@ describe('search workflow', () => {
     it('should update set_atb if a repeat search is made on a different day', async () => {
         // set set_atb to an older version
         await bgPage.evaluate((lastWeeksAtb) => dbg.settings.updateSetting('set_atb', lastWeeksAtb), lastWeeksAtb)
-
         // run a search
         const searchPage = await browser.newPage()
-        searchPage.goto('https://duckduckgo.com/?q=test')
-
-        await bgPage.waitForResponse(res => res.url().match(/atb\.js/))
```
kdzwinel, line 175: ditto

integration-test/background/atb.js
```diff
@@ -187,10 +206,13 @@ describe('search workflow', () => {
 
         // run a search
         const searchPage = await browser.newPage()
-        searchPage.goto('https://duckduckgo.com/?q=test')
-
-        await bgPage.waitForResponse(res => res.url().match(/atb\.js/))
```
kdzwinel, line 192: ditto

kdzwinel (commented):
I left couple of questions. BTW I should be testing `npm run test-int` correct?

integration-test/background/atb.js
Charlie-belmer: It was deprecated in 2020, so I figured I would replace it with the most recommended currently (node-fetch).

integration-test/background/atb.js
Charlie-belmer: Maybe there was an intended test in there, but it always times out and can cause errors (I don't think this is the right way to test whether a sub-request was made, but I could be wrong). The rest of the test succeeds without it. 

If we want to check that the atb.js file loaded, we could probably add a separate test for that?

integration-test/background/atb.js
kdzwinel: Ah, good call 👍

shared/js/background/atb.es6.js
```diff
@@ -126,6 +126,7 @@ const ATB = (() => {
         // builds a new query string containing only accepted params
         getAcceptedParamsFromURL: (url) => {
             const validParams = new URLSearchParams()
+            if (url === '') return validParams
```
dharb, line 129: Not sure I understand what could cause url to be blank here, since it's pulling from [getDDGTabUrls](https://github.com/duckduckgo/duckduckgo-privacy-extension/blob/develop/shared/js/background/chrome-wrapper.es6.js#L41). @Charlie-belmer did you encounter this case while testing?

dharb (approved):
This is great, thanks so much for getting things passing again. Left a couple questions, otherwise lgtm 👍

PR #522, opened 2020-12-04
Fix occaisional referrer error message

Charlie-belmer:
Add try/catch block around ddg_referrer code block, in case it errors out (guard was not enough in every case).

### DIFF
--- a/shared/data/fingerprint-protection.js
+++ b/shared/data/fingerprint-protection.js
@@ -98,27 +98,31 @@
 
     // ddg_referrer is defined in chrome-events.es6.js and injected as a variable if referrer should be modified
     // Unfortunately, we only have limited information about the referrer and current frame. A single
-    // page may load many requests and sub frames, all with different referrers. Since we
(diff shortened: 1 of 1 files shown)


shared/data/fingerprint-protection.js
```diff
@@ -98,27 +98,31 @@
...
+            fingerprintPropertyValues['document'] = {
+                'referrer': {
+                    'object': 'document',
+                    'origValue': document.referrer,
+                    'targetValue': `"${trimmedReferer}"`
+                }
             }
         }
+    } catch (e) {
+        // If for some reason ddg_referrer is not set, don't error out.
```
kdzwinel, line 125: 1. Do we know when `ddg_referrer` is not being set? 
2. nit: As far as I understand there was an error because we referred to a global, and sometimes not existing, `ddg_referrer` variable. We could do it safely with `if(window['ddg_referrer'] …` instead of try/catch, but maybe try/catch is safer?

Charlie-belmer:
Fixed instead with PR#523

PR #524, opened 2020-12-07
Improve browser detection for edge and brave

Charlie-belmer:
This improves browser detection to include Brave and Edge (previously a bug prevented Edge detection) for some behavior. It also adds a parameter to atb requests with this browser value (Firefox, Chrome, Brave, Edge).

Asana: https://app.asana.com/0/inbox/1172258661922799/1199403702264245/1199505030150494

### DIFF
--- a/shared/js/background/atb.es6.js
+++ b/shared/js/background/atb.es6.js
@@ -39,7 +39,7 @@ const ATB = (() => {
             }
 
             let randomValue = Math.ceil(Math.random() * 1e7)
-            let url = `${ddgAtbURL}${randomValue}&atb=${atbSetting}&set_atb=${setAtbSetting}${errorParam}`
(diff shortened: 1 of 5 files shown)


jdorweiler:
Need to update https://github.com/duckduckgo/duckduckgo-privacy-extension/blob/develop/shared/scss/_mixins.scss#L221 to add `.is-browser--brave` 
Screen Shot 2020-12-08 at 8 48 27 AM

jdorweiler:
Probably a few more css changes too. Search for `.is-browser` in `shared/scss`. Maybe it's easier to leave it as chrome and check for `window.navigator.brave` in atb.js?

Charlie-belmer:
Yes, excellent find! I instead updated the code to default to chrome in most cases, so we don't duplicate a lot of css. Please take another look when you have a few minutes. I think now all the various UI places this value reaches are covered.

jdorweiler:
Looks good, thanks

PR #527, opened 2020-12-10
Protect against some anti-fingerprint bypasses.

Charlie-belmer:
Improve anti-fingerprinting scripts to prevent some bypasses.

### DIFF
--- a/shared/data/fingerprint-protection.js
+++ b/shared/data/fingerprint-protection.js
@@ -78,12 +78,12 @@
 //            'userAgent': {
 //                'object': 'navigator',
 //                'origValue': navigator.userAgent,
-//                'targetValue': `"${JSON.stringify(ddg_ext_ua)}"` // Defined in chrome-events.es6.js and injected as a variable
+//                'targetValue': ddg_ext_ua // Defined in chrome-events.es6.js and injected as a variable
 //            },
               'appVersion': {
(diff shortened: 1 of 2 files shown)


palant:
Yes, this seems to solve it. For reference, there is [a better way to solve this on Firefox](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/Sharing_objects_with_page_scripts), something that I wasn’t aware of myself until recently. Sadly, Chromium doesn’t have such functionality, there running code in a hostile environment is unavoidable.

palant:
There are more places in this script affected for the same issue however. For example, `navigator.getBattery()` is being overwritten instead of `Navigator.prototype.getBattery()`. The resulting value is also a mess – not a `BatteryManager` instance and nothing like it either.

There are calls to `setWindowPropertyValue()` which all suffer from the same issue as well and more…

Charlie-belmer:
Thanks @palant,

There's way more to update here, as you mention. I'll be taking a closer look in the new year.

Charlie-belmer:
Thanks for all the help here - I think I addressed all your comments, and it's definitely looking way better.

dharb (approved):
Looks great, thanks @Charlie-belmer for making these changes. And hat-tip to @palant for the help identifying these issues, we really appreciate it.

PR #529, opened 2020-12-15
Add trailing slash to modified referrer

Charlie-belmer:
Modifies the referrer header truncation function to include a trailing '/' when the referrer is modified to a domain. This is because some sites do security checking on referrer, and may not match without this trailing slash.

### DIFF
--- a/shared/js/background/tracker-utils.js
+++ b/shared/js/background/tracker-utils.js
@@ -74,6 +74,7 @@ function truncateReferrer (referrer, target) {
     } else {
         modifiedReferrer = utils.extractLimitedDomainFromURL(referrer, {keepSubdomains: true})
     }
+    modifiedReferrer += '/'
     // If extractLimitedDomainFromURL fails (for instance, invalid referrer URL), it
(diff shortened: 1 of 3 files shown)


shared/js/background/tracker-utils.js
```diff
@@ -74,6 +74,7 @@ function truncateReferrer (referrer, target) {
     } else {
         modifiedReferrer = utils.extractLimitedDomainFromURL(referrer, {keepSubdomains: true})
     }
+    modifiedReferrer += '/'
```
kdzwinel, line 77: I think we should move this to `utils.extractLimitedDomainFromURL` (specifically to `${parsedURL.protocol}//${finalURL}**/**` part)? We are using this function only here so there is no risk of breaking anything else and appending something to URL here seems a bit out of place.

kdzwinel (commented):
I left one small note, feel free to merge after addressing it.

shared/js/background/tracker-utils.js
Charlie-belmer: yes great thinking, thanks!

PR #540, opened 2021-01-26
Block third party cookies

SlayterDev:
**Reviewer:** @kdzwinel 

## Description:

This PR adds blocking functionality for 3rd party cookies

## Steps to test this PR:
1. Visit https://privacy-test-pages.glitch.me/privacy-protections/storage-blocking
2. Click "Store data"
3. Click "Retrieve data"
4. Third party header and js cookie should fail

## Automated tests:
- [x] Unit tests
- [x] Integration tests

###### Reviewer Checklist:

###### PR Author Checklist:


shared/js/content-scripts/block-cookie.js
```diff
@@ -0,0 +1,55 @@
...
+                    }
+
+                    // disable setting cookies
+                    document.__defineSetter__ ('cookie', function(value) { })
+                })()
+            `
+
+            const scriptElement = document.createElement('script')
+            scriptElement.innerHTML = scriptString
+            document.documentElement.prepend(scriptElement)
```
dharb, line 52: Just a heads up that this is going to get blocked by sites with strict script-src CSPs in Firefox, take a look at what happens on twitter.com. Firefox offers a great workaround called [Xray vision](https://developer.mozilla.org/en-US/docs/Mozilla/Tech/Xray_vision). Check out how I used it in our GPC content script [here](https://github.com/duckduckgo/duckduckgo-privacy-extension/blob/develop/shared/js/content-scripts/GPC-moz.js) (as compared to the [content script](https://github.com/duckduckgo/duckduckgo-privacy-extension/blob/develop/shared/js/content-scripts/GPC.js) we use for chrome).

shared/js/background/chrome-events.es6.js
```diff
@@ -56,10 +57,26 @@ chrome.webRequest.onHeadersReceived.addListener(
...
             return ATB.updateSetAtb(request)
         }
+
+        // Strip 3rd party response header
+        const tab = tabManager.get({ tabId: request.tabId })
+        if (!tab || !request.responseHeaders || tab.site.whitelisted) return
+        if (cookieConfig.isExcluded(request.url)) return
+        const index = request.responseHeaders.findIndex(header => { return header.name.toLowerCase() === 'set-cookie' })
+        if (index !== -1) {
+            if (!utils.isFirstParty(request.url, tab.url)) {
```
kdzwinel, line 67: nit: any reason why we are doing this check here and not after `if (cookieConfig.isExcluded(request.url)) return`? It seems like we can do it before we start looking for a header.

shared/js/background/utils.es6.js
```diff
@@ -185,6 +185,16 @@ function isSafeListed (url) {
...
 }
 
+/**
+ * Tests whether the two URL's belong to the same
+ * first party set.
+ */
+function isFirstParty (url1, url2) {
+    const first = tldts.parse(url1).domain
+    const second = tldts.parse(url2).domain
+    return first === second
```
kdzwinel, line 195: hm, I wonder what will happen here for edge cases like IPs and chrome://newtab etc.

cc @sammacbeth

shared/js/background/storage/cookies.es6.js
```diff
@@ -0,0 +1,67 @@
...
+        switch (listName) {
+        case 'cookieExcludeList':
+            this.processExcludeList(listData)
+            break
+        }
+    }
+
+    async processExcludeList (data) {
+        for (const record of data.excludedDomains) {
+            this.excludedDomains.push(record.domain)
```
kdzwinel, line 58: nit: we are always adding to that array but never clear it - does it mean that the array increases in size every time list is fetched?

shared/js/background/storage/cookies.es6.js
```diff
@@ -0,0 +1,67 @@
...
+        }
+    }
+
+    isExcluded (url) {
+        const domain = tldts.parse(url).domain
+        return this.excludedDomains.find(elem => elem === domain)
```
kdzwinel, line 64: this assumes that excluded domains will always be top level and I wonder if that's correct 🤔 maybe in some cases we only want to safelist `accounts.google.com` but not the whole `google.com`?

shared/js/content-scripts/block-cookie.js
```diff
@@ -0,0 +1,92 @@
+// Set up 3rd party cookie blocker
+(function cookieBlocker () {
+    function getFrameUrl () {
+        let url = document.location.href
+        let parentFrame = (document != window.top) && window.parent
```
kdzwinel, line 5:
```suggestion
        let parentFrame = (document !== window.top) && window.parent
```

Is this file covered by our linting?

shared/js/content-scripts/block-cookie.js
```diff
@@ -0,0 +1,92 @@
...
+    if (window.top == window) {
+        return
+    }
+
+    chrome.runtime.sendMessage({
+        'checkThirdParty': true,
+        frameUrl: getFrameUrl()
+    }, function (isThridParty) {
+        if (isThridParty) {
+            const scriptString = `
```
kdzwinel, line 27: can we move this code in huge strings to a function like:

```js
function injectedClearCookies() {
   // Clear previously set cookies
  …
}

…

const scriptString = `(function() {${injectedClearCookies.toString()}()})()`

```

This way syntax will be linted and higlighted in diffs 🌈

shared/js/content-scripts/block-cookie.js
```diff
@@ -0,0 +1,92 @@
...
+                    // Clear previously set cookies
+                    var cookies = document.cookie.split("; ");
+                    for (var c = 0; c < cookies.length; c++) {
+                        var d = window.location.hostname.split(".");
+                        while (d.length > 0) {
+                            var cookieBase = encodeURIComponent(cookies[c].split(";")[0].split("=")[0]) + '=; expires=Thu, 01-Jan-1970 00:00:01 GMT; domain=' + d.join('.') + ' ;path=';
+                            var p = location.pathname.split('/');
+                            document.cookie = cookieBase + '/';
+                            while (p.length > 0) {
+                                document.cookie = cookieBase + p.join('/');
```
kdzwinel, line 38: I haven't realized it's that complicated to clear cookies. I thought it will be `document.cookie = ''` :)

BTW I remembered about https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Clear-Site-Data - we could inject this header to force browser to clear cookeis - not sure if that'd make anything easier, but just FYI

shared/js/content-scripts/block-cookie.js
```diff
@@ -0,0 +1,92 @@
...
+                    document.__defineSetter__ ('cookie', function(value) { })
+                })()
+            `
+
+            const scriptElement = document.createElement('script')
+            scriptElement.innerHTML = scriptString
+            document.documentElement.prepend(scriptElement)
+        } else {
+            const scriptString = `
+                (function() {
```
kdzwinel, line 55: ditto re string -> function

shared/js/content-scripts/block-cookie.js
```diff
@@ -0,0 +1,92 @@
...
+                    var cookie_setter = document.__lookupSetter__ ('cookie')
+                    // Enforce 7 day max expiry
+                    document.__defineSetter__ ('cookie', function(value) {
```
kdzwinel, line 58: what will happen to session cookies?

kdzwinel (commented):
Lookin' good 💪 I left couple of recommendations.

shared/js/content-scripts/block-cookie.js
SlayterDev: This is magical ✨

shared/js/background/chrome-events.es6.js
```diff
@@ -200,6 +218,27 @@ chrome.runtime.onMessage.addListener((req, sender, res) => {
...
+        var action = {
+            isThirdParty: false,
+            shouldBlock: false
+        }
+        const tab = tabManager.get({ tabId: sender.tab.id })
+        if (tab && tab.site.whitelisted) {
+            res(action)
+        }
+
+        if (!utils.isFirstParty(req.frameUrl, sender.tab.url)) {
```
sammacbeth, line 232: Do we need to get the `frameUrl` in the content script? According to [MDN](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/runtime/MessageSender) the `sender.url` will return the frame URL if it is sent from a frame:

> If the sender is a script running in a web page (including content scripts as well as normal page scripts), then url will be the web page URL. If the script is running in an iframe, url will be the iframe's URL.

shared/js/content-scripts/block-cookie.js
```diff
@@ -0,0 +1,90 @@
...
+        })
+    }
+
+    if (window.top === window) {
+        return
+    }
+
+    chrome.runtime.sendMessage({
+        'checkThirdParty': true,
+        frameUrl: getFrameUrl()
```
sammacbeth, line 76: See earlier comment in the `onMessage` listener - the frameURL should automatically be set on the `sender`.

shared/js/content-scripts/block-cookie.js
```diff
@@ -0,0 +1,90 @@
...
+            var d = window.location.hostname.split('.')
+            while (d.length > 0) {
+                var cookieBase = encodeURIComponent(cookies[c].split(';')[0].split('=')[0]) + '=; expires=Thu, 01-Jan-1970 00:00:01 GMT; domain=' + d.join('.') + ' ;path='
+                var p = location.pathname.split('/')
+                document.cookie = cookieBase + '/'
+                while (p.length > 0) {
+                    document.cookie = cookieBase + p.join('/')
+                    p.pop()
+                };
+                d.shift()
```
sammacbeth, line 31: If I understand this correctly, you are unsetting cookies for all possible subdomains of `window.location.hostname`. Why is that necessary? Can we not just use the existing domain field?

shared/js/content-scripts/block-cookie.js
```diff
@@ -0,0 +1,90 @@
...
+                    const setDate = new Date(dateStr)
+                    if (setDate > nextWeek) {
+                        comps[i] = 'expires=' + nextWeek.toUTCString()
+                    }
+
+                    break
+                }
+            }
+
+            if (!expiryFound) {
```
sammacbeth, line 60: Will this convert session cookies to normal cookies?

shared/js/content-scripts/block-cookie.js
```diff
@@ -0,0 +1,90 @@
...
+    }
+
+    function enforceCookieExpiry () {
+        var cookieSetter = document.__lookupSetter__('cookie')
+        // Enforce 7 day max expiry
+        document.__defineSetter__('cookie', function (value) {
+            var today = new Date()
+            var nextWeek = new Date(today.getFullYear(), today.getMonth(), today.getDate() + 7)
+            var comps = value.split(';')
+            var expiryFound = false
```
sammacbeth, line 46: How about using `;max-age=604800`? I believe the browser will ignore 'expiries' if 'max-age' is specified ([Source](https://mrcoles.com/blog/cookies-max-age-vs-expires/))

sammacbeth (commented):
Looks good! I left a few comments.

shared/js/background/chrome-events.es6.js
```diff
@@ -56,10 +57,27 @@ chrome.webRequest.onHeadersReceived.addListener(
             // returns a promise
             return ATB.updateSetAtb(request)
         }
+
+        // Strip 3rd party response header
+        const tab = tabManager.get({ tabId: request.tabId })
+        if (!request.responseHeaders) return
+        if (tab && tab.site.whitelisted) return
+        if (tab && utils.isFirstParty(request.url, tab.url)) return
```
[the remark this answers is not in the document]
SlayterDev: I went back and forth on this a bit but think you may be right here. cc @kdzwinel

shared/js/content-scripts/block-cookie.js
SlayterDev: Initially it was just being over cautious but based on the rest of the implementation the current domain should suffice.

shared/js/background/utils.es6.js
SlayterDev: Do we need to worry about IP addresses in this implementation? If so is there an easy check if a domain is an IP address so we can compare? Also cc @Charlie-belmer who originally wrote this method for UA rotation.

shared/js/background/storage/cookies.es6.js
```diff
@@ -0,0 +1,67 @@
...
+    async processExcludeList (data) {
+        this.excludedDomains = []
+        for (const record of data.excludedDomains) {
+            this.excludedDomains.push(record.domain)
+        }
+    }
+
+    isExcluded (url) {
+        const domain = (new URL(url)).host
+        return this.excludedDomains.find(elem => elem === domain)
```
kdzwinel, line 64: Ah, now we are not giving ourselves an option to exclude whole domain with subdomains (we will have to all full hosts). Maybe that's OK, but maybe we should do incremental check instead here? e.g. "sub.domain.com" -> check("sub.domain.com") + check("domain.com")

shared/js/background/utils.es6.js
kdzwinel: IMO we should be taking IPs into account, tldts returns `isIP` property so that should be straightforward.

As for the private domains, I think we should stick to the standard definition of a third party, as used by other browsers, and don't use private domains for now. It's a good idea for a follow up though.

shared/js/content-scripts/block-cookie.js
```diff
@@ -0,0 +1,80 @@
...
+    }
+
+    if (window.top === window) {
+        return
+    }
+
+    chrome.runtime.sendMessage({
+        'checkThirdParty': true
+    }, function (action) {
+        console.log(action)
```
kdzwinel, line 68: nit: we should remove console.logs in content scripts before merging

kdzwinel (commented):
I took another pass here and found some issues:

1. We are only blocking cookies from being set, but we should also block them from being sent. Right now, if someone goes to e.g. google.com, we will allow them to set cookies (first party rule), but we will then also allow those cookies to be sent on other sites!

2. On https://privacy-test-pages.glitch.me/privacy-protections/storage-blocking/ it looks like JS cookie in a 3p iframe is being set - it shouldn't. This is possibly related to an error that I'm getting in the console - " d.shift is not a function".

3. Tests: we should add an integration test for https://privacy-test-pages.glitch.me/privacy-protections/storage-blocking/ and, ideally, unit tests for manipulating cookies as this code looks a bit scary 😬

sammacbeth:
@kdzwinel re: tests, I've opened #544 which does integration tests against privacy-test-pages for cookie blocking. This is intended to cover the tests needed for my project, but it also easily covers 3rd party cookie blocking, so I added those tests too. We could cherry-pick that commit into the PR to add tests.

shared/js/background/chrome-events.es6.js
```diff
@@ -56,10 +57,27 @@ chrome.webRequest.onHeadersReceived.addListener(
...
+        }
+
+        return { responseHeaders: request.responseHeaders }
     },
     {
         urls: ['<all_urls>']
-    }
+    },
+    [
+        'blocking', 'responseHeaders', 'extraHeaders'
```
sammacbeth, line 79: Firefox doesn't support the `extraHeaders` option here which causes an exception when the extension is started (`Uncaught Error: Type error for parameter extraInfoSpec (Error processing 2: Invalid enumeration value "extraHeaders") for webRequest.onHeadersReceived.`)

You can feature detect support for these flags via `chrome.webRequest.OnHeadersReceivedOptions`. e.g.
```js
const extraInfoSpec = ['blocking', 'responseHeaders'];
if (chrome.webRequest.OnHeadersReceivedOptions.EXTRA_HEADERS) {
   extraInfoSpec.push(chrome.webRequest.OnHeadersReceivedOptions.EXTRA_HEADERS)
}
```

shared/js/content-scripts/block-cookie.js
```diff
@@ -0,0 +1,80 @@
...
+        for (var c = 0; c < cookies.length; c++) {
+            var d = window.location.hostname
+            var cookieBase = encodeURIComponent(cookies[c].split(';')[0].split('=')[0]) + '=; expires=Thu, 01-Jan-1970 00:00:01 GMT; domain=' + d + ' ;path='
+            var p = location.pathname.split('/')
+            document.cookie = cookieBase + '/'
+            while (p.length > 0) {
+                document.cookie = cookieBase + p.join('/')
+                p.pop()
+            };
+            d.shift()
```
sammacbeth, line 24:
```suggestion
```

(Looks like this should have been removed in refactor)

browsers/chrome/manifest.json
```diff
@@ -40,6 +40,7 @@
         "matches": ["<all_urls>"],
         "all_frames": true,
         "css": ["public/css/noatb.css"],
+        "js": ["public/js/content-scripts/block-cookie.js"],
```
sammacbeth, line 43: Should this also be included in the Firefox manifest?

shared/js/content-scripts/block-cookie.js
```diff
@@ -0,0 +1,80 @@
...
+        }
+
+        // disable setting cookies
+        document.__defineSetter__('cookie', function (value) { })
```
sammacbeth, line 28:
```suggestion
        document.__defineSetter__('cookie', function (value) { })
        document.__defineGetter__('cookie', () => '')
```

When testing this code I saw that `document.cookie` somehow became `undefined`. This broke scripts in the iframe because, according to the spec `document.cookie` always returns a string. Adding a stubbed getter seems to fix this.

browsers/chrome/manifest.json
SlayterDev: Yes 🤦‍♂️

shared/js/content-scripts/block-cookie.js
```diff
@@ -0,0 +1,80 @@
...
+        document.__defineSetter__('cookie', function (value) {
+            const nextWeek = 604800
+            var comps = value.split(';')
+            var expiryFound = false
+            for (var i = 0; i < comps.length; i++) {
+                if (comps[i].trim().startsWith('expires')) {
+                    expiryFound = true
+                    const dateStr = comps[i].trim().replace('expires=', '').trim()
+                    const setDate = new Date(dateStr)
+                    if (setDate > nextWeek) {
```
sammacbeth, line 43: This doesn't look right (comparing a `Date` with the number of seconds in a week).

SlayterDev: Yes still working on this as part of the session cookie fix 👍

SlayterDev:
I've commited a change to use the TR block list for cookie blocking. Feel free to review it but it is currently untested due to intermittent power issues here.

shared/js/background/chrome-events.es6.js
```diff
@@ -56,10 +64,32 @@ chrome.webRequest.onHeadersReceived.addListener(
...
+            if (experiment && experiment.blockingActivated) {
+                // Strip 3rd party response header
+                const tab = tabManager.get({ tabId: request.tabId })
+                if (!request.responseHeaders) return
+                if (tab && tab.site.whitelisted) return
+                if (tab && utils.isFirstParty(request.url, tab.url)) return
+                const index = request.responseHeaders.findIndex(header => { return header.name.toLowerCase() === 'set-cookie' })
+                if (index !== -1) {
+                    if (!cookieConfig.isExcluded(request.url) && trackerutils.isTracker(request.url)) {
+                        request.responseHeaders.splice(index, 1)
```
kdzwinel, line 81: I found out that this will only remove singe `set-cookie` header while in the wild there can be multiple:

3-set-cookie

We should probably `.filter()` all of those out here.

shared/js/background/chrome-events.es6.js
```diff
@@ -75,16 +76,13 @@ chrome.webRequest.onHeadersReceived.addListener(
...
-                        request.responseHeaders.splice(index, 1)
-                    }
+                if (!cookieConfig.isExcluded(request.url) && trackerutils.isTracker(request.url)) {
+                    responseHeaders = responseHeaders.filter(header => header.name.toLowerCase() !== 'set-cookie')
                 }
             }
         }
 
-        return { responseHeaders: request.responseHeaders }
+        return { responseHeaders: responseHeaders }
```
sammacbeth, line 99:
```suggestion
        return { responseHeaders }
```

shared/js/background/chrome-events.es6.js
```diff
@@ -46,8 +49,13 @@ chrome.webRequest.onBeforeRequest.addListener(
     ['blocking']
 )
 
+const extraInfoSpec = ['blocking', 'responseHeaders']
+if (chrome.webRequest.OnHeadersReceivedOptions.EXTRA_HEADERS) {
+    extraInfoSpec.push(chrome.webRequest.OnHeadersReceivedOptions.EXTRA_HEADERS)
+}
 chrome.webRequest.onHeadersReceived.addListener(
     request => {
+        utils.isFirstParty('1.2.3.4', '5.6.7.8')
```
sammacbeth, line 72: Is this left here by mistake?

shared/js/background/chrome-events.es6.js
```diff
@@ -56,10 +64,30 @@ chrome.webRequest.onHeadersReceived.addListener(
             // returns a promise
             return ATB.updateSetAtb(request)
         }
+
+        let responseHeaders = request.responseHeaders
+        const activeExperiment = settings.getSetting('activeExperiment')
+        if (activeExperiment) {
+            const experiment = settings.getSetting('experimentData')
+
+            if (experiment && experiment.blockingActivated) {
```
sammacbeth, line 87: As this block comes up 3 times in this file, could we extract this into a function that just returns if we're in the cookie blocking experiment?

shared/js/background/chrome-events.es6.js
sammacbeth: I would also consider only attaching `responseHeaders` to the `BlockingResponse` if we actually change the headers. This will prevent any potential overhead for unchanged requests.

shared/js/background/chrome-events.es6.js
```diff
@@ -200,6 +227,42 @@ chrome.runtime.onMessage.addListener((req, sender, res) => {
...
+                }
+
+                const tab = tabManager.get({ tabId: sender.tab.id })
+                if (tab && tab.site.whitelisted) {
+                    res(action)
+                }
+
+                if (!utils.isFirstParty(sender.url, sender.tab.url)) {
+                    action.isThirdParty = true
+                }
```
sammacbeth, line 267:
```suggestion
                action.isThirdParty = !utils.isFirstParty(sender.url, sender.tab.url))
```

shared/js/background/chrome-events.es6.js
```diff
@@ -200,6 +227,42 @@ chrome.runtime.onMessage.addListener((req, sender, res) => {
...
+                if (!cookieConfig.isExcluded(sender.url) && trackerutils.isTracker(sender.url)) {
+                    action.shouldBlock = true
+                }
```
sammacbeth, line 270:
```suggestion
                action.shouldBlock = !cookieConfig.isExcluded(sender.url) && trackerutils.isTracker(sender.url))
```

shared/js/content-scripts/block-cookie.js
```diff
@@ -0,0 +1,51 @@
...
+        for (var c = 0; c < cookies.length; c++) {
+            var d = window.location.hostname
+            var cookieBase = encodeURIComponent(cookies[c].split(';')[0].split('=')[0]) + '=; expires=Thu, 01-Jan-1970 00:00:01 GMT; domain=' + d + ' ;path='
+            var p = location.pathname.split('/')
+            document.cookie = cookieBase + '/'
+            while (p.length > 0) {
+                document.cookie = cookieBase + p.join('/')
+                p.pop()
+            };
+        }
```
sammacbeth, line 24: It occurs to me that this may cause issues for sites that the user also visits in first party context. Consider a site that embeds a 3rd party iframe from YouTube. This code will wipe out all cookies except those that are http-only. On a subsequent visit to Youtube, there may be some breakage caused by the absense of these cookies.

Perhaps it is sufficient just to block the cookie setter and getter in this case?

SlayterDev: Hmm this is a good point. What are your thoughts @kdzwinel?

kdzwinel: Hm, good point. Those cookies will be still saved in a browser but should not be shared in the third party context. My only worry is that, as far as I remember, the removal of all 3p cookies was helping us to avoid the timing issue (i.e. if we injected our script after some cookies were already set it's OK because we would remove them). Am I recalling this correctly? Is this still the case?

SlayterDev: Yes this is still here to take care of any race conditions

sammacbeth: I still think this will cause breakage. For example, on YouTube there is a consent cookie set as not http-only, which would therefore be wiped by this. This means that everytime you visit a page with embedded youtube iframe, your consent settings would get wiped.

I think fixes for race conditions are planned, so we can wait for a proper solution there before addressing this.

kdzwinel: Hm, OK lets drop removing cookies for now. We have to note this gap in the relevant task and update tech design.

integration-test/background/storage.js
```diff
@@ -0,0 +1,54 @@
...
+        await harness.teardown(browser)
+    })
+
+    it('does not block 1st party HTTP cookies', async () => {
+        const headerCookie = cookies.find(({ name, domain }) => name === 'headerdata' && domain === testPageDomain)
+        expect(headerCookie).toBeTruthy()
+        expect(headerCookie.expires).toBeGreaterThan(Date.now() / 1000)
+    })
+
+    it('blocks 3rd party HTTP cookies', async () => {
```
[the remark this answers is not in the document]
sammacbeth: These are the original tests I wrote in #544 against the old tech designs - these will probably need updating for the updated policy.

shared/js/background/storage/cookies.es6.js
```diff
@@ -0,0 +1,81 @@
...
+    isDomainExcluded (domain) {
+        if (this.excludedDomains.find(elem => elem === domain)) {
+            return true
+        }
+
+        const comps = domain.split('.')
+        if (comps.length > 2) {
+            comps.shift()
+            return this.isDomainExcluded(comps.join('.'))
+        }
```
sammacbeth, line 76: This function is O(N*M) in the worst case, where N is the length of `excludedDomains` and M is the subdomain depth of the given domain. As this is run for every third-party request, then most calls will run to the worst case complexity, and this function will be called often. 

I think it would make sense to optimise this a bit. Using a trie you should be able to get O(log M) performance on average.

shared/js/background/chrome-events.es6.js
```diff
@@ -70,10 +88,25 @@ chrome.webRequest.onHeadersReceived.addListener(
...
+            // Strip 3rd party response header
+            const tab = tabManager.get({ tabId: request.tabId })
+            if (!request.responseHeaders) return
+            if (tab && tab.site.whitelisted) return
+            if (tab && utils.isFirstParty(request.url, tab.url)) return
+            if (!cookieConfig.isExcluded(request.url) && trackerutils.isTracker(request.url)) {
+                responseHeaders = responseHeaders.filter(header => header.name.toLowerCase() !== 'set-cookie')
+            }
+
+            return { responseHeaders: responseHeaders }
```
sammacbeth, line 103:
```suggestion
            if (!cookieConfig.isExcluded(request.url) && trackerutils.isTracker(request.url)) {
                responseHeaders = responseHeaders.filter(header => header.name.toLowerCase() !== 'set-cookie')
                return { responseHeaders }
            }
            ```

This avoids returning if response headers are unchanged.

**Review by sammacbeth — CHANGES_REQUESTED** [review]

File: shared/js/background/storage/cookies.es6.js (hunk 32 above)
Comment (SlayterDev) [reply shared/js/background/storage/cookies.es6.js:76]: I'm having issues imagining the trie implementation (my theory is rather rusty). Would you be able to go a bit more in depth. Do you think we'll be hitting the worst case most often because most cookies will be excluded by the root domain (i.e. example.com vs test.example.com)? If so the other idea I have is recursing to the root domain first and bubbling up from there. However I think some granularity in the exclusions could be lost this way?

Comment (sammacbeth) [reply shared/js/background/storage/cookies.es6.js:76]: I would organise the exclude list by reversed domain parts:
```json
{
  "com": {
     "google": {
        "accounts": true
     },
     "googleapis": true
}
```

Then you get worst-case O(M), but constant time average case with a lookup algorithm like:
```js
const parts = domain.split('.').reverse()
let ptr = this.excludedDomainsLookup;
for (const i = 0; i < parts.length; i++) {
  ptr = ptr[parts[i]]
  if (ptr === true) { // matching subdomain
     return true
  } else if (ptr === undefined) { // empty leaf
     return false;
  }
}
```
(warning: untested code that I just dropped in here, might not work at all!)

sammacbeth:
Superseded by #570

PR #541, opened 2021-01-26
Canvas fingerprint protection

jonathanKingston:
**Reviewer:**

## Automated tests:
- [ ] Unit tests
- [ ] Integration tests

###### Reviewer Checklist:

###### PR Author Checklist:


shared/js/injected-content-scripts/fingerprint-protection.js
```diff
@@ -197,14 +197,39 @@
         }
     }
 
+    async function getSjcl () {
+        const url = chrome.runtime.getURL('/data/sjcl.js')
```
jonathanKingston, line 201: I think bundling this in the build script will be faster and also cleaner than doing it at injection time.

shared/js/background/chrome-events.es6.js
```diff
@@ -232,7 +249,7 @@ chrome.webNavigation.onCommitted.addListener(details => {
             }
             chrome.tabs.executeScript(details.tabId, variableScript)
             const scriptDetails = {
-                'file': '/data/fingerprint-protection.js',
+                'file': '/public/js/injected-content-scripts/fingerprint-protection.js',
```
jonathanKingston, line 252: It may make sense to inject this file statically always and initialize it like this:
```
initFingerprintProtection(JSON.stringify({
  agent: agentSpoofer.getAgent(),
  sessionKey,
  referrer: tab.referrer
}));
```

shared/js/injected-content-scripts/fingerprint-protection.js
```diff
@@ -4,7 +4,7 @@
  * data through obsufcation / randomness.
  */
 
-(function protect () {
+(async function protect () {
```
Charlie-belmer, line 7: Does making this async make any of the protection timing issues worse here?

Charlie-belmer (commented):
Directionally it's looking good! I think we still want to have some remote disable ability for particular sites, and I'm not sure if making the whole thing async causes any timing issues. I didn't test it myself yet, but it's looking pretty solid.

shared/js/injected-content-scripts/fingerprint-protection.js
jonathanKingston: By definition, yeah there will at least be one tick of JS that is worse; The browser APIs aren't really clear on how long a local fetch will take but it's certainly better not to do that. However I think using the build script as mentioned to combine all the content-scope scripts into this file fixes it anyway.

shared/js/background/chrome-events.es6.js
```diff
@@ -207,42 +221,61 @@ chrome.runtime.onMessage.addListener((req, sender, res) => {
...
-                'matchAboutBlank': true
+async function getContentScope () {
+    const url = chrome.runtime.getURL('/public/js/content-scope.js')
+
+    let response = await fetch(url)
+    return response.text()
+}
+
+async function init () {
+    const contentScopeScript = await getContentScope()
```
jonathanKingston, line 235: @Charlie-belmer so I moved the code to be only called here once which shouldn't slow the performance of anything, the only tiny impact will be setup of the background script listener below will be delayed on install / update of the extension.

The end goal being all code in fingerprint-protection.js will move into it's own file in: shared/js/content-scope allowing us to lint it properly.
This move eventually would give us the ability to inject fingerprint-protection.js always statically; but activate it in this listener, which should reduce the amount of chance of a race condition.

shared/data/constants.js
```diff
@@ -96,6 +96,18 @@ module.exports = {
             "format": "text",
             "source": "external"
         },
+        {
+            "name": "brokenCanvasSiteList",
+            "url": "/data/broken-canvas-sites.txt",
```
jonathanKingston, line 101: @Charlie-belmer these need to be remote lists probably. Shall I add the empty files to the alowlist repo?

Charlie-belmer: Yeah that is the way to go - let me know if you need a hand with it.

jonathanKingston:
@Charlie-belmer I removed the remote config in a commit: https://github.com/duckduckgo/duckduckgo-privacy-extension/pull/541/commits/cbcd82b0a4c3f1ff4174c6f61706288b5a218e76 until we merge: https://github.com/duckduckgo/content-blocking-whitelist/pull/111  (we can revert this out once it's merged)

Currently the CDN server responds with a code 200 HTML SERP page which confuses the code enough to start exempting most results as we regex per line etc. I suspect we will want to fix this as there is a little fragility here if the server is down, we also won't apply any protection ever (I tried treating content-type != text/plain as an error but none of the FP code ever runs)

I've fixed up the 3rd party iframe code where it was accessing `window.top` by mistake rather than using the injected `args.site` param with the domain in. I added a way to filter to a specific integration test and keep the browser open as debugging this was really difficult.

shared/js/content-scope/fingerprint.js
```diff
@@ -0,0 +1,112 @@
...
+                byte = byte >> 1
+            }
+        }
+        return imageData
+    }
+    Object.defineProperty(CanvasRenderingContext2D.prototype, 'getImageData', {
+        value: getImageData
+    })
+
+    // TODO hide toString
```
jonathanKingston, line 87: I think we should block release on this as pages will be able to stringify out code, I've not looked into the correct solution enough yet as it's a function binding to null didn't seem to work. (The solution I've used in the past is just overload toString also but there's probably a more elegant solution)

integration-test/helpers/harness.js
```diff
@@ -2,6 +2,9 @@ const puppeteer = require('puppeteer')
 const execSync = require('child_process').execSync
 
 jasmine.DEFAULT_TIMEOUT_INTERVAL = 20000
+if (process.env.KEEP_OPEN) {
+    jasmine.DEFAULT_TIMEOUT_INTERVAL = 20000 * 1000
```
Charlie-belmer, line 6: Some of the testing changes in develop might fix issues you were seeing here (I hope)

shared/js/content-scope/fingerprint.js
Charlie-belmer: Here's what I was doing on other functions, and it seemed to work to prevent stringifying functions. It binds both the replacement function and the getter to null (though here it's returning an object with subfunctions). See the temporary storage FP protections for where this example comes from.
`
someObject.subObject = function replacementFunction (callback, err) {...}.bind(null)

Object.defineProperty(Something.prototype, 'property', {get: (() => someObject).bind(null)})
`

shared/js/injected-content-scripts/fingerprint-protection.js
```diff
@@ -3,8 +3,10 @@
  * object properties and methods to reduce entropy or modify fingerprint
  * data through obsufcation / randomness.
  */
+/* globals ddg_args */
```
Charlie-belmer, line 6: nice - these were going to get unwieldy :)

Charlie-belmer (commented):
It's looking pretty good! A few things I noticed while looking it over today:

- tests likely need to be merged - there were some improvements merged separately that should remove some deprecations and failures
- I am seeing test and linting failures

Let me know when you are ready for me to take it for another spin.

integration-test/helpers/harness.js
jonathanKingston: This was mostly just for giving me an easier time testing a single test and keeping it open on the test I wanted so I could use the inspector etc.

shared/js/content-scope/fingerprint.js
```diff
@@ -0,0 +1,111 @@
+/* global sjcl */
```
sammacbeth, line 1: Is it possible to not leave this global in the document content scope? This will be
 1. Visible to other scripts on the page - could be used for extension detection
 2. Modifiable by other scripts - they could manipulate the api to break randomisation.

shared/js/content-scope/fingerprint.js
```diff
@@ -0,0 +1,111 @@
...
+    for (let line of errorLines) {
+        let res = line.match(lineTest)
+        if (res) {
+            let path = res[2]
+            // checked already
+            if (errorFiles.has(path)) {
+                continue
+            }
+            if (shouldExemptUrl(path)) {
+                console.log('Exempting script path:', path)
```
sammacbeth, line 45: We should put logging behind a debug flag before release to avoid polluting page logs.

shared/js/content-scope/fingerprint.js
```diff
@@ -0,0 +1,111 @@
...
+                    pixel = nextRandom(pixel)
+
+                    // Right shift as we use the least significant bit of it
+                    byte = byte >> 1
+                }
+            }
+            return imageData
+        }
+    })
+    CanvasRenderingContext2D.prototype.getImageData = getImageDataProxy
```
sammacbeth, line 86: What is the value of `CanvasRenderingContext2D.prototype.getImageData.toString()` after this?

shared/js/content-scope/fingerprint.js
jonathanKingston: The content-scope code is bundled together and thrown in an IFFE for this reason. The 'global' here is just external to the file in question but scoped only to the IFFE.

Perhaps having a test to verify this is always the case would be worthwhile though.

shared/js/content-scope/fingerprint.js
jonathanKingston: This is the status quo, I do agree though.

shared/js/content-scope/fingerprint.js
jonathanKingston: If we weren't using proxies, the content page can see the DDG code which is something we would like to prevent.

sammacbeth: So the proxy trick preserves the original `toString()` for this method? Nice.

shared/js/content-scope/fingerprint.js
jonathanKingston: Added a test.

shared/js/content-scope/fingerprint.js
Charlie-belmer: Maybe we should remove all logging calls from the scripts that run on page? Other ones might have been left in by mistake, unless they are informative to a user debugging for themselves I suppose.

jonathanKingston:
Making this dependant on using a json list as it's a little hard to get all the list code changed. It's actually a change I wanted to make anyway.

Either way to test change the constants file to:

```
            //"url": "https://duckduckgo.com/contentblocking/fingerprinting.json",
            "url": "http://127.0.0.1:8080/fingerprinting.json",
```

Check out: https://github.com/duckduckgo/content-blocking-lists/pull/2 and run `http-server` in that directory or similar cli servers.

Charlie-belmer (approved):
LGTM now, thanks for everything! Feel free to merge when ready.

PR #545, opened 2021-02-11
Expiration policy for first party cookies

sammacbeth:
**Reviewer:** @kdzwinel 

**Depends on:** #540 

## Description:

Sets the `max-age` of cookies set via `document.cookie` from a 3rd party script to 7 days, if they set an expiry longer than 7 days.

Implementation notes:
 * A script is injected in all frames to override the `document.cookie` setter.
 * After injecting this script, we fetch a policy from the extension background, and push it to the injected script. This message also includes the eTLD+1 for the current tab (used to determine if scripts are cross-site).
 * When a cookie is set, we look at the call stack to determine which scripts set the cookie. We compare this to the 1st party domain to determine if the script is 3rd party (this checks same site rather than same origin).
 * If the cookie is set by a 3rd party, and the expiry exceeds the policy threshold, we modify the cookie to set max-age as defined by the policy.

Note this PR currently targets the branch for #540 as it builds on top of those changes.

## Steps to test this PR:
1. Open a website with Google Analytics
2. Check for the `_ga` cookie in the storage tab of devtools
3. The cookie should expire in 2 weeks, rather than 2 years.

## Automated tests:
- [ ] Unit tests
- [ ] Integration tests - Can be added pending deployment of https://github.com/duckduckgo/privacy-test-pages/pull/24

###### Reviewer Checklist:

###### PR Author Checklist:

### DIFF
--- a/integration-test/background/storage.js
+++ b/integration-test/background/storage.js
@@ -4,8 +4,9 @@ const wait = require('../helpers/wait')
 const testPageDomain = 'privacy-test-pages.glitch.me'
 
 describe(`On https://${testPageDomain}/privacy-protections/storage-blocking/`, () => {
-    const thirdPartyDomain = 'broken.third-party.site'
-    let cookies
+    const thirdPartyDomain = 'good.third-party.site'
+    const thirdPartyTracker = 'broken.third-party.site'
+    let cookies = []
 
     beforeAll(async () => {
         const { browser } = await harness.setup()
@@ -14,7 +15,7 @@ describe(`On https://${testPageDomain}/privacy-protections/storage-blocking/`, (
         try {
             page.on('requestfinished', (req) => {
                 // once we see this url, we can consider the test completed
-                if (req.url().startsWith(`https://${thirdPartyDomain}/privacy-protections/storage-blocking/3rdparty.js`)) {
+                if (req.url().startsWith(`https://${thirdPartyTracker}/privacy-protections/storage-blocking/3rdparty.js`)) {
                     iframeFullyLoaded = true
                 }
             })
@@ -23,7 +24,7 @@ describe(`On https://${testPageDomain}/privacy-protections/storage-blocking/`, (
             while (!iframeFullyLoaded) {
                 await wait.ms(100)
             }
-            await wait.ms(400) // allow cookies to be set
+            await wait.ms(1000) // allow cookies to be set
             // collect all browser cookies
             cookies = (await page._client.send('Network.getAllCookies')).cookies
         } finally {
@@ -40,6 +41,12 @@ describe(`On https://${testPageDomain}/privacy-protections/storage-blocking/`, (
 
     it('does not block 3rd party HTTP cookies not on block list', async () => {
         const headerCookie = cookies.find(({ name, domain }) => name === 'headerdata' && domain === thirdPartyDomain)
+        expect(headerCookie).toBeTruthy()
+        expect(headerCookie.expires).toBeGreaterThan(Date.now() / 1000)
+    })
+
+    it('blocks 3rd party HTTP cookies for trackers', async () => {
+        const headerCookie = cookies.find(({ name, domain }) => name === 'headerdata' && domain === thirdPartyTracker)
         expect(headerCookie).toBeUndefined()
     })
 
@@ -50,7 +57,21 @@ describe(`On https://${testPageDomain}/privacy-protections/storage-blocking/`, (
     })
 
     it('does not block 3rd party JS cookies not on block list', async () => {
-        const headerCookie = cookies.find(({ name, domain }) => name === 'jsdata' && domain === thirdPartyDomain)
-        expect(headerCookie).toBeUndefined()
+        const jsCookie = cookies.find(({ name, domain }) => name === 'jsdata' && domain === thirdPartyDomain)
+        expect(jsCookie).toBeTruthy()
+        expect(jsCookie.expires).toBeGreaterThan(Date.now() / 1000)
+    })
+
+    it('does not block 1st party JS cookies set by non-trackers', () => {
+        const jsCookie = cookies.find(({ name, domain }) => name === 'tpsdata' && domain === testPageDomain)
+        expect(jsCookie).toBeTruthy()
+        expect(jsCookie.expires).toBeGreaterThan(Date.now() / 1000 + 864000) // 10 days in the future
+    })
+
+    it('reduces the expiry of 1st party JS cookies set by trackers to 8 days', () => {
+        const jsCookie = cookies.find(({ name, domain }) => name === 'tptdata' && domain === testPageDomain)
+        expect(jsCookie).toBeTruthy()
+        expect(jsCookie.expires).toBeGreaterThan(Date.now() / 1000)
+        expect(jsCookie.expires).toBeLessThan(Date.now() / 1000 + 691200) // 8 days in the future
     })
 })
--- a/integration-test/background/storage.js
+++ b/integration-test/background/storage.js
@@ -15,10 +15,13 @@ describe(`On https://${testPageDomain}/privacy-protections/storage-blocking/`, (
         try {
             page.on('requestfinished', (req) => {
                 // once we see this url, we can consider the test completed
-                if (req.url().startsWith(`https://${thirdPartyTracker}/privacy-protections/storage-blocking/3rdparty.js`)) {
+                if (req.url().startsWith(`https://${thirdPartyTracker}/privacy-protections/storage-blocking/iframe.js`)) {
                     iframeFullyLoaded = true
                 }
             })
+            // Load the test pages home first to give some time for the extension background to start
+            // and register the content-script-message handler
+            await page.goto(`https://${testPageDomain}/`, { waitUntil: 'networkidle0' })
             await page.goto(`https://${testPageDomain}/privacy-protections/storage-blocking/?store`, { waitUntil: 'networkidle0' })
             // eslint-disable-next-line no-unmodified-loop-condition
             while (!iframeFullyLoaded) {
--- a/integration-test/background/storage.js
+++ b/integration-test/background/storage.js
@@ -18,9 +18,7 @@ describe(`On https://${testPageDomain}/privacy-protections/storage-blocking/`, (
                     iframeFullyLoaded = true
                 }
             })
-            await page.goto(`https://${testPageDomain}/privacy-protections/storage-blocking/`, { waitUntil: 'networkidle0' })
-            await page.evaluate('storeData()')
-            await page.waitForNavigation({ waitUntil: 'networkidle0' })
+            await page.goto(`https://${testPageDomain}/privacy-protections/storage-blocking/?store`, { waitUntil: 'networkidle0' })
             // eslint-disable-next-line no-unmodified-loop-condition
             while (!iframeFullyLoaded) {
                 await wait.ms(100)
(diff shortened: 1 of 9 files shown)


integration-test/background/storage.js
```diff
@@ -0,0 +1,54 @@
...
+        const headerCookie = cookies.find(({ name, domain }) => name === 'jsdata' && domain === testPageDomain)
+        expect(headerCookie).toBeTruthy()
+        expect(headerCookie.expires).toBeGreaterThan(Date.now() / 1000)
+    })
+
+    it('blocks 3rd party JS cookies', async () => {
+        const headerCookie = cookies.find(({ name, domain }) => name === 'jsdata' && domain === thirdPartyDomain)
+        expect(headerCookie).toBeUndefined()
+    })
+})
```
kdzwinel, line 54: we need test for the modified expiry, but I understand that is waiting for https://github.com/duckduckgo/privacy-test-pages/pull/24, right?

shared/js/content-scripts/block-cookie.js
```diff
@@ -28,72 +28,109 @@
...
-                        comps[i] = 'max-age=' + nextWeek
-                    }
+    /**
+     * Apply an expiry policy to cookies set via document.cookie.
+     * @param {string} secret Used to detect messages sent from the extension content-script
+     */
+    function applyCookieExpiryPolicy (secret) {
+        const cookieSetter = document.__lookupSetter__('cookie')
+        const cookieGetter = document.__lookupGetter__('cookie')
+        const lineTest = /(\()?(http[^)]+):[0-9]+:[0-9]+(\))?/
```
kdzwinel, line 38: Can you please check with @jonathanKingston how he is parsing the stack trace? I know he spent quite some time testing different variants in chrome and firefox (https://privacy-test-pages.glitch.me/features/stack-tracing/).
We should probably consolidate both methods at some point.

shared/js/content-scripts/block-cookie.js
```diff
@@ -28,72 +28,109 @@
...
+                    resolve(event.data)
+                    window.removeEventListener('message', messageListener)
                 }
             }
-
-            if (!expiryFound) {
-                comps.push('max-age=' + nextWeek)
+            window.addEventListener('message', messageListener)
+        })
+        console.log('inject cookie test', cookieSetter, secret)
```
kdzwinel, line 50: we probably want to drop logging before merging or put it behind a flag (there are bunch of console.log's and console.warn's in the file)

shared/js/content-scripts/block-cookie.js
```diff
@@ -28,72 +28,109 @@
...
+                // determine the origins of the scripts in the stack
+                const scriptOrigins = (new Error().stack.split('\n')).reduce((origins, line) => {
+                    const res = line.match(lineTest)
+                    if (res && res[2]) {
+                        origins.add(new URL(res[2]).hostname)
+                    }
+                    return origins
+                }, new Set())
+                // wait for config before doing same-site tests
+                loadPolicy.then(({ tabRegisteredDomain, policy }) => {
```
kdzwinel, line 66: it looks like tabRegisteredDomain can come undefined if tab is not found. Not sure in what scenarios this can happen, but maybe we want a short path in such case?

shared/js/content-scripts/block-cookie.js
```diff
@@ -28,72 +28,109 @@
...
+                    const sameSiteScript = [...scriptOrigins].every((host) => host === tabRegisteredDomain || host.endsWith(`.${tabRegisteredDomain}`))
```
kdzwinel, line 67: I wonder if this should be `.every` or if we should look only at the last item (one that actually does the setting) or the first item (one that invokes the action) 🤔 Do we know how those stacks look like on various sites? e.g. what happens when some CDN hosted libraries are involved? This might be worth discussing a bit more in tech design.

shared/js/content-scripts/block-cookie.js
```diff
@@ -28,72 +28,109 @@
...
+                    // apply cookie policy
+                    if (expiry > policy.threshold) {
+                        if (maxAgeIdx === -1) {
+                            cookieParts.push(`max-age=${policy.maxAge}`)
+                        } else {
+                            cookieParts.splice(maxAgeIdx, 1, `max-age=${policy.maxAge}`)
+                        }
+                        console.log('update cookie', cookieParts.join(';'), scriptOrigins)
+                        cookieSetter.apply(document, [cookieParts.join(';')])
+                    }
```
kdzwinel, line 93: gosh, that cookie code is scary. Wish Cookie Storage API (https://wicg.github.io/cookie-store/) was available in FF. But since it isn't… can we somehow unit test this code? 🤔

shared/js/content-scripts/block-cookie.js
```diff
@@ -28,72 +28,109 @@
...
+                })
+            } catch (e) {
+                console.warn('error in cookie override', e)
             }
-
-            const cookieStr = comps.join(';')
-
-            cookieSetter.apply(document, cookieStr)
         })
+        document.__defineGetter__('cookie', cookieGetter)
```
kdzwinel, line 99: hm, are we doing anything with that getter? Seems like we don't need to touch it at all?

shared/js/content-scripts/block-cookie.js
```diff
@@ -28,72 +28,109 @@
...
+                    const expiresIdx = cookieParts.findIndex(part => part.trim().toLowerCase().startsWith('expires'))
+                    if (maxAgeIdx === -1 && expiresIdx === -1) {
+                        // session cookie
+                        return
+                    }
+                    const expiry = maxAgeIdx >= 0
+                        ? parseInt(cookieParts[maxAgeIdx].split('=')[1])
+                        : (new Date(cookieParts[expiresIdx].split('=')[1]) - new Date()) / 1000
+                    // apply cookie policy
+                    if (expiry > policy.threshold) {
```
kdzwinel, line 85: to be clear, we are modifying the expiry even if page is safelisted, correct? Let's note this in the tech design (I know you proposed it in the comments).

kdzwinel (commented):
I left some comments, but overall this looks really solid 💪

shared/js/background/chrome-events.es6.js
```diff
@@ -238,11 +239,21 @@ chrome.runtime.onMessage.addListener((req, sender, res) => {
             if (req.checkThirdParty) {
                 const action = {
                     isThirdParty: false,
-                    shouldBlock: false
+                    shouldBlock: false,
+                    tabRegisteredDomain: tldts.parse(sender.tab.url).hostname,
+                    policy: {
+                        threshold: 604800,
```
kdzwinel, line 245: I thought we discussed threshold to be 16d? Let's clarify in tech design.

integration-test/background/storage.js
sammacbeth: Correct.

integration-test/background/storage.js
```diff
@@ -0,0 +1,54 @@
...
+        const { browser } = await harness.setup()
+        const page = await browser.newPage()
+        let iframeFullyLoaded = false
+        try {
+            page.on('requestfinished', (req) => {
+                // once we see this url, we can consider the test completed
+                iframeFullyLoaded |= req.url() === `https://${thirdPartyDomain}/privacy-protections/storage-blocking/iframe.js`
+            })
+            await page.goto(`https://${testPageDomain}/privacy-protections/storage-blocking/`, { waitUntil: 'networkidle0' })
+            await page.evaluate('storeData()')
```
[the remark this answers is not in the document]
sammacbeth: Nice. That should also improve the test as it might catch potential race condition issues.

shared/js/content-scripts/block-cookie.js
sammacbeth: This is the regex copied from his PR.

I'm not sure the best way to share code for this, given this is in a function injected into the page context via `Function.toString()`, so needs to be self contained.

shared/js/content-scripts/block-cookie.js
sammacbeth: It shouldn't ever be undefined - the code that sends it will use the `sender.tab.url` from the `runtime.onMessage` callback if the tab store doesn't have the tab. This case happens when the extension reloads - we don't have tab info for existing open tabs.

shared/js/content-scripts/block-cookie.js
sammacbeth: Again, this code is tricky to extract in a way that we can unittest because of the injection method. However, if we can extract it we could certainly test it by throwing a load of cookie strings see in the wild at it (e.g. from tracker-radar data) to see if there are any missed edge cases.

shared/js/content-scripts/block-cookie.js
sammacbeth: I found if I don't do this, then `document.cookie` is `undefined`.

shared/js/content-scripts/block-cookie.js
sammacbeth: Yes, this policy applies _always_, though with this implementation it would be easy to skip this on pages that the user has allowed.

shared/js/content-scripts/block-cookie.js
jonathanKingston: Sorry missed this, we could generate a file into the `content-scope` directory from a utils directory perhaps?

shared/js/content-scripts/block-cookie.js
jonathanKingston: Can we use the `content-scope` injection method I added in my patch? This then keeps the injected code into it's own file which a unit test could load in the same way / import or do whatever without all the other utility code.

shared/js/content-scripts/block-cookie.js
kdzwinel: TIL!

kdzwinel: In fact, can we leave a comment that this is the case? Someone (me in 1mo) may be tempted to remove this thinking it's useless.

shared/js/content-scripts/block-cookie.js
kdzwinel: Resolving since we discussed it in Tech Design and changed direction to skip safelisted pages for now.

shared/js/content-scripts/block-cookie.js
```diff
@@ -123,12 +110,18 @@
...
+        const scriptString = `{
+            ${imports};
+            (${func.toString()})('${arg}')
+        }`
         const doc = window.wrappedJSObject ? window.wrappedJSObject.document : document
         const scriptElement = doc.createElement('script')
         scriptElement.innerHTML = scriptString
         doc.documentElement.prepend(scriptElement)
+        // remove element immediately so it is not visible to scripts on the page
+        doc.documentElement.removeChild(scriptElement)
```
kdzwinel, line 124: oh, we can do it sync? I thought this would require `setTimeout(,0)` or `script.onload`?

unit-test/background/cookie.es6.js
```diff
@@ -0,0 +1,48 @@
...
+        })
+        it('updates existing max-age', () => {
+            const cki = new Cookie('jsdata=; expires=Thu, 01-Jan-1970 00:00:01 GMT; domain=good.third-party.site ;max-age=50')
+            expect(cki.getExpiry()).toBe(50)
+            cki.maxAge = 100
+            expect(cki.getExpiry()).toBe(100)
+            expect(cki.toString().indexOf('max-age=100')).not.toBe(-1)
+        })
+    })
+})
```
kdzwinel, line 48: Nicely done 👏 Can we throw at it some more "weird but valid" cookie values from https://github.com/nfriedly/set-cookie-parser/blob/master/test/set-cookie-parser.js#L49 ?

shared/js/content-scripts/block-cookie.js
sammacbeth: I also thought the same, but when I tested it sync also worked.

sammacbeth:
Closing in favour of #570

PR #551, opened 2021-02-19
Fix canvas issues

jonathanKingston:
**Reviewer:** @Charlie-belmer 

## Description:

- First pixel data returning '0' created div by 0 issues which was pretty common
- The image data wasn't being used correctly, but stringifying the key full length was really expensive with the crypto algo
  - We could consider some method similar to nextRandom to return a short determinisitic string
- Verifying I'm hitting all the channels (I think I am anyway)
- TODO the same response appears to be returned toDataURL across origins.

## Automated tests:
- [ ] Unit tests
- [ ] Integration tests

###### Reviewer Checklist:

###### PR Author Checklist:

### DIFF
--- a/shared/js/background/utils.es6.js
+++ b/shared/js/background/utils.es6.js
@@ -170,6 +170,7 @@ function isBrokenList (url, lists) {
         if (brokenSiteDomain) {
             return hostname.match(new RegExp(brokenSiteDomain + '$'))
         }
+        return false
     })
 }
 
--- a/shared/js/background/utils.es6.js
+++ b/shared/js/background/utils.es6.js
@@ -167,7 +167,9 @@ function isBrokenList (url, lists) {
 
     // If root domain in temp unprotected list, return true
     return lists.some((brokenSiteDomain) => {
-        return hostname.match(new RegExp(brokenSiteDomain + '$'))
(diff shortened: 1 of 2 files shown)


Charlie-belmer (commented):
Looks good to me. I suggest that we add some comment documentation. It isn't obvious to me why sumData is needed. You mentioned this is WIP, so let me know when you want a closer look!

jonathanKingston:
> Looks good to me. I suggest that we add some comment documentation. It isn't obvious to me why sumData is needed. You mentioned this is WIP, so let me know when you want a closer look!

The main reason it's needed is due to the speed of using the entire data string. Using the privacy page example there was very obvious slow down in even drawing to the canvas. Previously this was inconsistent across browsers with Firefox using something like "[Object object]" as the key.
This is something I want to tear apart and make unit tests of when we have the ability to so this doesn't happen.

jonathanKingston:
The main rationale to using something more pseudo random than just summing the pixel data together, is to prevent collisions which might be exposable in the fingerprint and timing attacks would be less predictable too. But the counter case it comes at more cost. (however nothing on the cost of using a 30k byte key)

Charlie-belmer (approved):
Looks good to me, thanks for doing this!

PR #555, opened 2021-02-26
minimal messaging

sballesteros:
**Reviewer:**

## Automated tests:
- [ ] Unit tests
- [ ] Integration tests

###### Reviewer Checklist:
- [ ] **Ensure the PR solves the problem**
- [ ] **Review every line of code**
- [ ] **Ensure the PR does no harm by testing the changes thoroughly**
- [ ] **Get help if you're uncomfortable with any of the above!**
- [ ] Determine if there are any quick wins that improve the implementation

###### PR Author Checklist:
- [ ] Get advice or leverage existing code
- [ ] Agree on technical approach with reviewer (if the changes are nuanced)
- [ ] Ensure that there is a testing strategy (and documented non-automated tests)
- [ ] Ensure there is a documented monitoring strategy (if necessary)
- [ ] Consider systems implications

PR #568, opened 2021-03-12
Onboarding infrastructure (v2)

sballesteros:
**Reviewer:** @jonathanKingston  @kdzwinel 

Asana: https://app.asana.com/0/11984721910118/1199989403323564/f

## Description:

Alternative to https://github.com/duckduckgo/duckduckgo-privacy-extension/pull/555.
👉  See https://dub.duckduckgo.com/duckduckgo/ddg/pull/21857 for technical design and testing instructions.

## Automated tests:
- [ ] Unit tests
- [ ] Integration tests

###### Reviewer Checklist:

###### PR Author Checklist:


shared/js/background/chrome-events.es6.js
```diff
@@ -38,6 +45,111 @@ chrome.runtime.onInstalled.addListener(function (details) {
...
+    }
+}
+
+let onBeforeNavigateTimeStamp = null
+chrome.webNavigation.onBeforeNavigate.addListener(details => {
+    onBeforeNavigateTimeStamp = details.timeStamp
+})
+
+chrome.webNavigation.onCommitted.addListener(details => {
+    if (details.url.includes('duckduckgo.com/?')) {
```
jonathanKingston, line 77: Can you make this a [url filter](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/events/UrlFilter) instead?

The performance improvement might not be significant but it should reduce risk also.

shared/js/background/onboarding.es6.js
```diff
@@ -0,0 +1,117 @@
...
+        browser,
+        extensionId
+    } = JSON.parse(__injectedJSON__) // eslint-disable-line
+
+    function getDocumentStartData (cb) {
+        if (browser !== 'chrome') {
+            return cb(null)
+        }
+
+        window.postMessage({ type: 'documentStartDataRequest' }, window.location.origin)
```
jonathanKingston, line 39: Can all of these `postMessage` go through a central function?

Also could the location be hardcoded to the one in the filter?

shared/js/background/onboarding.es6.js
```diff
@@ -0,0 +1,117 @@
...
+        isAddressBarQuery,
+        showWelcomeBanner,
+        showCounterMessaging,
+        extensionId,
+        browser
+    })
+
+    // TODO: upgrade to `chrome.scripting.executeScript` when we upgrade to manifest v3
+    // as it allows to inject a function with _arguments_. Here we simulate that in a hacky way
+    return `(${onboarding.toString().replace('__injectedJSON__', `'${params}'`)})()`
```
jonathanKingston, line 22: This actually I **think** this could suffer from an escaping bug if the scope of these variables changed (although it appears these are very limited in scope so shouldn't). The arguments above having a single quote will break out of this escaping and be turned into real code.

You could actually just pass the stringified params into this:
```
`(${onboarding.toString())(${params})`
```

Then just change the function signature below:
```
function onboarding ({
        isAddressBarQuery,
        showWelcomeBanner,
        showCounterMessaging,
        browser,
        extensionId
    } = {}) {
```
(untested)

shared/js/background/onboarding.es6.js
```diff
@@ -0,0 +1,117 @@
...
+                showWelcomeBanner,
+                showCounterMessaging
+            }, documentStartData)
+
+            // the content script do not share the same `window` as the page
+            // so we inject a `<script>` to be able to access the page `window`
+            const script = document.createElement('script')
+            script.textContent = `
+                    if (window.onFirstSearchPostExtensionInstall) {
+                        window.onFirstSearchPostExtensionInstall(${JSON.stringify(args)})
```
jonathanKingston, line 103: Can this not be done with post messaging also?

jonathanKingston (changes requested):

shared/js/background/chrome-events.es6.js
sballesteros: Thanks, I didn't know about `URLFilter`. This should now be fixed 👍

shared/js/background/onboarding.es6.js
sballesteros: Sorry not sure that I fully understand what you mean by "the one in the filter".
Note that there are 2 different things here:
- The post message calls from the `getDocumentStartData` function. This is to get data from another content script injected by the extension (unfortunately, we need to capture this data _before_ `chrome.webNavigation.onCommitted` fires so we need to inject another script declaratively from the extension manifest).
- The response from messages received from the site

shared/js/background/onboarding.es6.js
sballesteros: Thanks, good catch this should now be fixed 👍

shared/js/background/onboarding.es6.js
sballesteros: So my understanding is that we wanted to avoid adding an event listener to the SERP (we would need that event listener for all queries which was considered wasteful).
I fully agree that the current approach is gross but I think that it's the only way to avoid an event listener on the SERP and make sure that the SERP is only impacted on the first search post extension install.

shared/js/background/chrome-events.es6.js
```diff
@@ -38,6 +45,118 @@ chrome.runtime.onInstalled.addListener(function (details) {
...
+  */
+function getOnboardingParams (cb) {
+    function onReady () {
+        cb(null, {
+            showWelcomeBanner: settings.getSetting('showWelcomeBanner'),
+            showCounterMessaging: settings.getSetting('showCounterMessaging')
+        })
+    }
+
+    if (settings.isReady()) {
```
jonathanKingston, line 64: Cool thanks for highlighting this: https://app.asana.com/0/892838074342800/1200071625925781

This code could be made more JavaScript like with:
```
async function getOnboardingParams () {
    if (!settings.isReady()) {
        await settings.ready()
    }
    return {
        showWelcomeBanner: settings.getSetting('showWelcomeBanner'),
        showCounterMessaging: settings.getSetting('showCounterMessaging')
    }
}
```

With usage like:
```
const { showWelcomeBanner, showCounterMessaging } = await getOnboardingParams();
```

shared/js/background/onboarding.es6.js
jonathanKingston: Thanks for the explanation!

shared/js/background/onboarding.es6.js
jonathanKingston: > Regarding hardcoding the location, I used window.location.origin to facilitate local dev. I think that this is safe as those scripts can only be injected to *.duckduckgo.com (per the manifest) so in practice window.location.origin is akin to *.duckduckgo.com here.

Note that the url filter will make this harder etc but as you mention we can fix this later.

shared/js/background/onboarding.es6.js
```diff
@@ -0,0 +1,115 @@
...
+                cb(null, e.data.payload)
+            }
+        })
+    }
+
+    function start () {
+        // Consolidate data: grab value of the Chrome Dialogue content-script
+        // this script was injected earlier to capture value as early as possible
+        getDocumentStartData((err, documentStartData) => {
+            if (err) {
```
jonathanKingston, line 50: Can we remove this argument if it's not currently used?

shared/js/background/onboarding.es6.js
jonathanKingston: I'm not sure if there is an obvious place to document this, bit I think it's worthwhile perhaps making a few comments point to a central README or comment perhaps?

For example it wasn't immediately clear on the moving parts:
- `onFirstSearchPostExtensionInstall` is exposed (sometimes) on SERP
- We inject `chrome-dialogue.js` on document_start
- Later we message to the script injected on start to understand if the document was focused
- We merge this data into the calling of the `onFirstSearchPostExtensionInstall`  method on SERP

Perhaps `chrome-dialogue.js` should be renamed now it's changed to `serp-document-start-handler.js`

shared/js/background/chrome-events.es6.js
```diff
@@ -38,6 +45,118 @@ chrome.runtime.onInstalled.addListener(function (details) {
...
+    }
+}
+
+let onBeforeNavigateTimeStamp = null
+chrome.webNavigation.onBeforeNavigate.addListener(details => {
+    onBeforeNavigateTimeStamp = details.timeStamp
+})
+
+chrome.webNavigation.onCommitted.addListener(details => {
+    getOnboardingParams((err, { showWelcomeBanner, showCounterMessaging }) => {
```
jonathanKingston, line 77: Can we remove this arg if not used?

shared/js/background/onboarding.es6.js
```diff
@@ -0,0 +1,115 @@
...
+}
+
+function onboarding ({
+    isAddressBarQuery,
+    showWelcomeBanner,
+    showCounterMessaging,
+    browser,
+    extensionId
+}) {
+    function getDocumentStartData (cb) {
```
jonathanKingston, line 32: This function could be made more JavaScript like by using a promise instead:
```
function getDocumentStartData (cb) {
    if (browser !== 'chrome') {
        return null
    }
    return new Promise((resolve) => {
        window.postMessage({ type: 'documentStartDataRequest' }, window.location.origin)
            window.addEventListener('message', function handleMessage (e) {
                if (e.origin === window.location.origin && e.data.type === 'documentStartDataResponse') {
                    window.removeEventListener('message', handleMessage)
                    resolve(e.data.payload)
                }
            })
        }
    })
}
```

Then usage:
```
const documentStartData = await getDocumentStartData();
```

jonathanKingston (changes requested):

shared/js/background/chrome-events.es6.js
```diff
@@ -38,6 +45,118 @@ chrome.runtime.onInstalled.addListener(function (details) {
...
+
+    chrome.runtime.onStartup.addListener(() => {
+        const init = () => {
+            if (settings.getSetting('rescheduleCounterMessagingOnStart')) {
+                settings.removeSetting('rescheduleCounterMessagingOnStart')
+                settings.updateSetting('showCounterMessaging', true)
+            }
+        }
+        if (settings.isReady()) {
+            init()
```
jonathanKingston, line 153: Sorry missed this one too, this could use a conditional `await` instead here

shared/js/background/chrome-events.es6.js
```diff
@@ -38,6 +45,118 @@ chrome.runtime.onInstalled.addListener(function (details) {
...
+    if (settings.isReady()) {
+        onReady()
+    } else {
+        settings.ready().then(onReady)
+    }
+}
+
+let onBeforeNavigateTimeStamp = null
+chrome.webNavigation.onBeforeNavigate.addListener(details => {
+    onBeforeNavigateTimeStamp = details.timeStamp
```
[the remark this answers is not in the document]
sballesteros: This is intended. We want to be sure that the user didn't start another navigation before injecting our script

shared/js/background/chrome-events.es6.js
sballesteros: Both `showWelcomBanner` and `showCounterMessaging` are used see:

```js
createOnboardingCode({
  isAddressBarQuery,
  showWelcomeBanner,
  showCounterMessaging,
  browser,
  extensionId: chrome.runtime.id
})
```

shared/js/background/chrome-events.es6.js
jonathanKingston: Based on: https://app.asana.com/0/inbox/1199237043596759/1200071625925781/1200075771973335

This can be further shortened to:
```
async function getOnboardingParams () {
    await settings.ready()
    return {
        showWelcomeBanner: settings.getSetting('showWelcomeBanner'),
        showCounterMessaging: settings.getSetting('showCounterMessaging')
    }
}
```

shared/js/background/settings.es6.js
```diff
@@ -95,5 +95,8 @@ module.exports = {
     updateSetting: updateSetting,
     removeSetting: removeSetting,
     logSettings: logSettings,
-    ready: ready
+    ready: ready,
+    isReady () {
+        return isReady
```
jonathanKingston, line 100: I think we can drop this given: https://app.asana.com/0/inbox/1199237043596759/1200071625925781/1200075771973335

shared/js/background/chrome-events.es6.js
jonathanKingston: Sorry I wasn't clear, I meant `err`.
I think this is covered by other comments were we can just make it async instead.

shared/js/background/onboarding.es6.js
jonathanKingston: >> Can we remove this argument if it's not currently used?

> documentStartData is used see:

Again I meant `err` sorry for the confusion.

--- 

As for the other comment...

> Note that the url filter will make this harder etc

Essentially we have the ddg url hardcoded here is what I meant (harder was a wrong choice in words) so we still have the local development issue currently.

Ah! so the regex is only there for local development. Makes sense; we have this issue elsewhere with most things in `constants.js` I suspect we should hardcode to "https://duckduckgo.com" and also put it in that file.

This reduces the risk IMO as that regex could be wrongly interpreted and also we won't inject it into other properties like chat.duckduckgo.com etc.

shared/js/background/onboarding.es6.js
sballesteros: I hard-coded location (using a constant when possible), renamed `chrome-dialogue.js` to `serp-document-start-handler.js`, and added comments to the file.
LMK if that still needs further changes or clarifications.

shared/js/background/chrome-events.es6.js
```diff
@@ -38,6 +45,118 @@ chrome.runtime.onInstalled.addListener(function (details) {
...
+                        console.error(chrome.runtime.lastError)
+                    }
+                })
+            }
+        }
+    })
+}, {
+    url: [
+        {
+            originAndPathMatches: '^https:\\/\\/(?:[^/]*\\.)?duckduckgo\\.com/$',
```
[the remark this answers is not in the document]
sballesteros: We are now using `hostEquals` and using a hardcoded value (defined in constants.js) for the hostname. This is not great for local dev, but we agreed to fix that in a subsequent PR.

shared/js/background/chrome-events.es6.js
sballesteros: I have simplified this logic.

shared/js/background/chrome-events.es6.js
sballesteros: I have refactored the code to use async/await instead of callbacks

shared/js/background/onboarding.es6.js
sballesteros: I have kept a callback (instead of a Promise + async/await) to avoid needing to import 'regenerator-runtime/runtime' until we configure `@babel/preset-env` more precisely. I think that this is OK for now but I can refactor to Promises + async/await if you feel strongly about it.

shared/js/background/chrome-events.es6.js
sballesteros: this has been refactored to use async/await

shared/js/background/settings.es6.js
sballesteros: yep, done.

shared/js/background/onboarding.es6.js
sballesteros: We are now using `hostEquals: constants.duckDuckGoSerpHostname` (we will improve the DX in a subsequent PR)

browsers/chrome/manifest.json
```diff
@@ -41,7 +41,12 @@
         "all_frames": true,
         "css": ["public/css/noatb.css"],
         "run_at": "document_start"
-      }
+      },
+      {
+        "matches": ["https://duckduckgo.com/*"],
+        "js": ["public/js/content-scripts/serp-document-start-handler.js"],
+        "run_at": "document_start"
```
kdzwinel, line 48: I'm not stoked that this script will always be injected (we need it once in most cases), but I understand that alternative of injecting it programmatically had timing issues?

shared/js/background/atb.es6.js
```diff
@@ -176,47 +173,18 @@ const ATB = (() => {
             // - the user wasn't already looking at the app install page
             // - the user hasn't seen the page before
             settings.ready().then(() => {
-                // Special case for the cross product promotion on desktop (cppd) experiment
-                if (settings.getSetting('isMultiStepOnboarding')) {
```
kdzwinel, line 180: wait, this experiment is gone? is this a unrelated cleanup?

shared/js/background/chrome-events.es6.js
```diff
@@ -4,10 +4,14 @@
  * on FF, we might actually miss the onInstalled event
  * if we do too much before adding it
  */
+import 'regenerator-runtime/runtime' // needed for async/await untill we config @babel/preset-env more precisely
```
kdzwinel, line 7:
```suggestion
import 'regenerator-runtime/runtime' // needed for async/await until we config @babel/preset-env more precisely
```

shared/js/background/onboarding.es6.js
```diff
@@ -0,0 +1,112 @@
...
+    }
+
+    function start () {
+        getDocumentStartData((err, documentStartData) => {
+            if (err) {
+                console.error(err)
+            }
+
+            // DDG privacy policy prevents us to use `chrome.runtime` on the SERP so we
+            // setup a relay here so that the SERP can access the background process
```
kdzwinel, line 54:
```suggestion
            // setup a relay here so that the SERP can communicate with the background process
```

kdzwinel (commented):
I left couple of comments. I'll try to run the whole setup on Monday.

shared/js/background/atb.es6.js
sballesteros: Yes, this is an unrelated cleanup. All the cleanup is done in [this commit](https://github.com/duckduckgo/duckduckgo-privacy-extension/pull/568/commits/c42419d88fc0f1aacf49f6921f857ac624a02e7d).

browsers/chrome/manifest.json
sballesteros: Yes not a fan either. I did that as manifest + `run_at: document_start` is the most reliable way to inject a script as early as possible in the page lifecycle (`onBeforeNavigate` is too early).
That being said thinking more I think that `onCommitted` + `document_start` should be good enough for our case so I went ahead with that change in this [commit](https://github.com/duckduckgo/duckduckgo-privacy-extension/pull/568/commits/f209a89f1095fbec630d045a163c031bf945e65f). We lose a few ms but it's cleaner/less wasteful.

kdzwinel (approved):
I got a chance to play with it and it LGTM! @jonathanKingston's review is still in the "request changes" state, so I'm not sure if there is something left to discuss? If not, please feel free to merge @jonathanKingston or @sballesteros .

jonathanKingston:
Good from me, I thought I'd wait for you.

sballesteros:
@kdzwinel  @jonathanKingston  I just fixed the merge conflict (sorry for the multiple commits 🙈). @jonathanKingston  could you approve / merge the PR (it still needs your approval). Thanks!

jonathanKingston (approved):

PR #569, opened 2021-03-16
Stop UA from attempting to rotate (disabled already)

Charlie-belmer:
Stop User Agent rotations from occurring. The UA spoofing is already disabled, this just stops support code from executing until all UA fixes are implemented.

### DIFF
--- a/shared/js/background/classes/agentspoofer.es6.js
+++ b/shared/js/background/classes/agentspoofer.es6.js
@@ -14,6 +14,7 @@ class AgentSpoofer {
 
     getAgent () {
         // TODO: Remove this when UA fixes occur
+        /* eslint-disable no-unreachable */
         return this.realAgent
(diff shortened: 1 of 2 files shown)


jonathanKingston (approved):

dharb (changes requested):
Thanks for doing this @Charlie-belmer. Looks like the linter is hollering about unreachable code now. You should be able to silence it by either commenting out the remainder of these methods after returning, or by adding `/* eslint-disable no-unreachable */` to the top of this file.

sammacbeth (commented):
Tbh I'd rather the code was removed rather than leaving a load of commented code in. When we want to restore the old code we can just revert this PR to undo the code deletion.

Charlie-belmer:
Closing this in favor of PR #593

PR #578, opened 2021-03-23
Fix Service Worker Cookie Issue

SlayterDev:
**Reviewer:** @sammacbeth 

## Description:

Fix service worker issues related to cookies

## Steps to test this PR:
See this task: https://app.asana.com/0/892838074342800/1200099548987553/f

## Automated tests:
- [ ] Unit tests
- [ ] Integration tests

###### Reviewer Checklist:

###### PR Author Checklist:


shared/js/background/chrome-events.es6.js
```diff
@@ -106,7 +106,14 @@ chrome.webRequest.onHeadersReceived.addListener(
...
             if (tab && tab.site.whitelisted) return
-            if (tab && utils.isFirstParty(request.url, tab.url)) return
+
+            if (!tab && request.tabId === -1) {
+                if (utils.isFirstParty(request.initiator, request.url)) {
+                    return
+                }
+            } else if (tab && utils.isFirstParty(request.url, tab.url)) {
+                return
+            }
```
sammacbeth, line 116: So if we don't have info on the tab (`!tab`) then we will always block tracker cookies. Will this not break when restoring a session, and the tabManager may not yet have tab info?

shared/js/background/chrome-events.es6.js
```diff
@@ -486,7 +493,13 @@ chrome.webRequest.onBeforeSendHeaders.addListener(
             const tab = tabManager.get({ tabId: request.tabId })
             if (!requestHeaders) return
             if (tab && tab.site.whitelisted) return
-            if (tab && utils.isFirstParty(request.url, tab.url)) return
+            if (!tab && request.tabId === -1) {
+                if (utils.isFirstParty(request.initiator, request.url)) {
```
sammacbeth, line 497: Why not put this all into one if statement?

shared/js/background/chrome-events.es6.js
SlayterDev: We'd only be removing the cookie header for tracker urls which I believe is the policy here.

shared/js/background/chrome-events.es6.js
SlayterDev: Felt messier but I have no qualms about it

shared/js/background/chrome-events.es6.js
jonathanKingston: Aren't some internal requests at least in Firefox in this category also?

sammacbeth: What I meant is that if `!tab && request.tabId > -1` and the url is a tracker URL we will still remove the cookie, without doing a first party check. I think we should handle that, just in case that happens and we end up blocking all the first party cookies on a site until the tab is updated.

SlayterDev: Fixed these scenarios. I encourage ya'll to give it a test 👍

jonathanKingston (approved):

Issue #530, opened 2020-12-16
Fallback behavior for referrer truncation results in an invalid value


palant:
The fallback behavior for referrer truncation in the `fingerprint-protection.js` content script looks as follows:

https://github.com/duckduckgo/duckduckgo-privacy-extension/blob/e15a3fba9fc46bca8f38114c574207e68b0446d6/shared/data/fingerprint-protection.js#L111-L114

This sets the referrer to hostname only which isn’t a valid URL. I guess that the idea here was to use `new URL(document.referrer).origin` instead which makes more sense.

Charlie-belmer:
Thanks for this report as well @palant! Nice find, I'll include a fix for this with the other fingerprint items as well.

kberridge:
I discovered your browser extension is stripping the port off the HTTP Referer request header (should be http://localhost:65314 but is http://localhost). Which breaks APIs that expect the Referer to match! In chrome, edge, and firefox. Disabling duck duck go extension fixes it.

Looks like this issue could be the same thing?  I didn't want to open a duplicate if so.

Charlie-belmer:
Thanks for mentioning this @kberridge. It's a slightly different issue, but I'll take a look!

sammacbeth:
This looks likely - the `hostname` attribute of a `URL` strips the port (while `host` includes it). It looks like the `origin` fix described here will also work with ports:
```javascript
const url = new URL('http://localhost:1234')
url.hostname // == "localhost"
url.host // == "localhost:1234"
url.origin // == "http://localhost:1234"
```

weaversam8:
Is it normal for the DDG privacy extension to trim subdomains too? The `Referer` header for an iframe was being set to `site.com` when I was making a request from `subdomain.site.com`, which was breaking permitted embeddable domains on said iframe.

sammacbeth:
> Is it normal for the DDG privacy extension to trim subdomains too? The `Referer` header for an iframe was being set to `site.com` when I was making a request from `subdomain.site.com`, which was breaking permitted embeddable domains on said iframe.

Yes. If a request is made to a third-party tracker domain the referer will be truncated to eTLD+1.

weaversam8:
[...]

PR #579, opened 2021-03-24
Fix referrer truncation edge cases

sammacbeth:
**Reviewer:**

## Description:
#530 

## Automated tests:
- [ ] Unit tests
- [ ] Integration tests

###### Reviewer Checklist:

###### PR Author Checklist:

### DIFF
--- a/shared/js/background/utils.es6.js
+++ b/shared/js/background/utils.es6.js
@@ -25,7 +25,9 @@ function extractLimitedDomainFromURL (url, { keepSubdomains } = {}) {
         const parsedURL = new URL(url)
         const tld = tldts.parse(url)
         if (!parsedURL || !tld) return ''
-        let finalURL = tld.domain
+        // tld.domain is null if this is an IP or the domain does not use a known TLD (e.g. localhost)
+        // in that case use the hostname (no truncation)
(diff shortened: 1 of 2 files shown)


Charlie-belmer (changes requested):
Looks good to me! The logic changes here also need to be reflected in injected-content-scripts/fingerprint-protection.js, so that the JS API returns the same thing as the header.

sammacbeth:
> Looks good to me! The logic changes here also need to be reflected in injected-content-scripts/fingerprint-protection.js, so that the JS API returns the same thing as the header.

@Charlie-belmer I think this already the case as that code uses the value created by the function I changed in the PR:
https://github.com/duckduckgo/duckduckgo-privacy-extension/blob/70ed6eab778a958a8c91009a5bdb43cc84242189/shared/js/injected-content-scripts/fingerprint-protection.js#L109-L116

Charlie-belmer:
I think you're right - as mentioned in Asana, I was thinking about line 115 here, where it creates a new referrer if there is some issue using Origin. This should already include the port, so we should be OK here.

Charlie-belmer (approved):

PR #581, opened 2021-03-24
Facebook Click to play

Charlie-belmer:
Reviewer: Dave

## Description:
Adds click to play functionality for Facebook. This replaces visible facebook elements with an element that a user can click to enable if they wish to view the content. After this change, no facebook requests will be sent until the user interacts with a facebook element.

Note that in this PR is a background HTML page for "learn more". This is likely to be removed before final merging, and instead be placed on our public help pages (still working with copy team).

## Steps to test this PR:
There is a test page that loads all the facebook elements (though login doesn't work on the test page, due to an invalid application ID from facebook) at http://privacy-test-pages.glitch.me/privacy-protections/click-to-load/ (note that metric counts are not accurate when using FireFox) 

In the network tab of dev tools, all facebook requests should be blocked until a user interacts with an element, at which point the element should load normally (may be slightly visually modified from it's original version). Login with facebook should work transparently on custom login buttons.

## Automated tests:
- [x] Unit tests
- [x] Integration tests

###### Reviewer Checklist:

###### PR Author Checklist:


shared/js/background/tracker-utils.js
```diff
@@ -30,6 +31,114 @@ function isTracker (url) {
...
+    return excludes
+}
+
+// Return true if URL is in our click to load tracker list
+function getSocialTracker (url) {
+    if (!facebookExperimentIsActive()) {
+        return
+    }
+    const parsedDomain = tldts.parse(url)
+    for (const [entity, data] of Object.entries(tdsStorage.ClickToLoadConfig)) {
```
sammacbeth, line 69: How many entries are in this config, and how large do we expect it to become? Does it make sense to build a lookup table to get O(1) lookups instead of O(N)? i.e. use an object with the keys being domains and the values being `data` from the `ClickToLoadConfig`?

shared/js/background/tracker-utils.js
```diff
@@ -30,6 +31,114 @@ function isTracker (url) {
...
+        if (data.domains.includes(parsedDomain.domain) && !data.excludedSubdomains.includes(parsedDomain.hostname)) {
+            let redirect
+            if (data.surrogates) {
+                for (const surrogate of data.surrogates) {
+                    if (url.match(surrogate.rule)) {
+                        redirect = surrogate.surrogate
```
sammacbeth, line 75: In the case of multiple matching rules, the _last_ one will be used. Is that intended or should there be a break here?

sammacbeth (commented):
Looks good! I just had a couple of nits and comments.

browsers/firefox/manifest.json
```diff
@@ -58,5 +58,5 @@
         "<all_urls>",
         "alarms"
     ],
-    "web_accessible_resources": ["web_accessible_resources/*"]
+    "web_accessible_resources": ["web_accessible_resources/fb-sdk.js", "web_accessible_resources/*", "public/font/*"]
```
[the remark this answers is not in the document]
Charlie-belmer: This is a remnant of some early surrogate work that I forgot about. Thanks for pointing it out!

shared/data/web_accessible_resources/fb-sdk.js
```diff
@@ -0,0 +1,144 @@
...
+
+    function loginPopup () {
+        const width = Math.min(window.screen.width, 450)
+        const height = Math.min(window.screen.height, 450)
+        const popupParams = `width=${width},height=${height},scrollbars=1,location=1`
+        window.open('about:blank', popupName, popupParams)
+    }
+
+    window.addEventListener('LoadFacebookSDK', enableFacebookSDK)
+    window.addEventListener('RunFacebookLogin', runFacebookLogin)
```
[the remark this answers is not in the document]
Charlie-belmer: They can emit these events and trigger the behavior, but it won't actually enable FB calls. That's because the flow requires a user click before it sends a message to the extension to actually allow FB traffic. In the worst case, a site falsely generating these events will break the surrogate function, as the surrogate will try to unsuccessfully load the real SDK.

shared/js/background/tracker-utils.js
```diff
@@ -30,6 +31,114 @@ function isTracker (url) {
...
+            const excludedDomains = data.excludedDomains.map(e => e.domain)
+            for (const domain of excludedDomains) {
+                excludes.push({
+                    entity: entity,
+                    domain: domain
+                })
+            }
+        }
+    }
+    return excludes
```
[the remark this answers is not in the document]
Charlie-belmer: Good call, added caching features here.

shared/js/background/tracker-utils.js
Charlie-belmer: It's a great idea, but I'm not sure it's worth it here. There is only going to be one entry per social site, so N > 20 or so is unlikely (right now, N=1). Each entry can have multiple domain/subdomains included or excluded, so the extra complexity doesn't seem worth it here to me, though happy to discuss more.

shared/js/background/tracker-utils.js
Charlie-belmer: Good thinking!

shared/data/experiments-out.js
```diff
@@ -30,5 +30,20 @@ module.exports = {
                 }
             }
         }
+    },
+    u: {
+        name: 'Facebook click to load experiment',
+        description: 'Test blocking all Facebook and replacing it with clickable elements.',
+        active: true,
+        atbExperiments: {
+            d: {
```
sammacbeth, line 39: Do you need a control group here, or are we just using the experiment as a feature flag?

shared/js/background/chrome-events.es6.js
```diff
@@ -603,6 +709,62 @@ chrome.webRequest.onBeforeSendHeaders.addListener(
...
+    { urls: ['<all_urls>'] },
+    ['responseHeaders']
+)
+
+// Inject our content script to overwite FB elements
+chrome.webNavigation.onCommitted.addListener(details => {
+    const tab = tabManager.get({ tabId: details.tabId })
+    if (tab && tab.site.isBroken) {
+        console.log('temporarily skip embedded object replacements for site: ' + details.url +
+          'more info: https://github.com/duckduckgo/content-blocking-whitelist')
```
sammacbeth, line 754: We changed this repo.
```suggestion
          'more info: https://github.com/duckduckgo/content-blocking-lists')
```

sammacbeth (approved):
LGTM. Just one v minor change and a question about the experiment setup.

shared/data/experiments-out.js
Charlie-belmer: There is a control variant defined in the experiment setup. If a user is in that group, they will still be in the experiment group, but the feature will be disabled. It's not really clear from the code in the extension that it exists though.

PR #582, opened 2021-03-25
Add caching for exclusion list

SlayterDev:
**Reviewer:** @sammacbeth @jonathanKingston @kdzwinel 

## Description:
Cookie exclusion list was not being cached. If the list endpoint returned a 304 there would be no response and the list would fail to load. This PR adds caching based on the UserAgent list implementation.

## Steps to test this PR:
1. Load extension
2. Visit site with excluded cookie domain (e.g. google 3rd party login)
3. Login should work
4. Reload extension (to force reloading exclusion list)
5. Repeat steps 2-3
**Note:** Since this relies on getting a 304 response you may need to reload the extension in quick succession

## Automated tests:
- [ ] Unit tests
- [ ] Integration tests

###### Reviewer Checklist:

###### PR Author Checklist:

### DIFF
--- a/shared/js/background/chrome-events.es6.js
+++ b/shared/js/background/chrome-events.es6.js
@@ -389,12 +389,12 @@ chrome.runtime.onMessage.addListener((req, sender, res) => {
             const parsed = tldts.parse(tabUrl)
             action.tabRegisteredDomain = parsed.domain === null ? parsed.hostname : parsed.domain
 
+            if (req.documentUrl && trackerutils.isTracker(req.documentUrl) && sender.frameId !== 0) {
+                action.isTrackerFrame = true
+            }
+
             action.isThirdParty = !utils.isFirstParty(sender.url, sender.tab.url)
             action.shouldBlock = !cookieConfig.isExcluded(sender.url)
-
-            if (req.documentUrl && trackerutils.isTracker(req.documentUrl) && sender.frameId !== 0) {
-                action.isTrackerFrame = true && action.shouldBlock
-            }
 
             res(action)
         } else {
--- a/shared/js/background/chrome-events.es6.js
+++ b/shared/js/background/chrome-events.es6.js
@@ -389,12 +389,13 @@ chrome.runtime.onMessage.addListener((req, sender, res) => {
             const parsed = tldts.parse(tabUrl)
             action.tabRegisteredDomain = parsed.domain === null ? parsed.hostname : parsed.domain
 
-            if (req.documentUrl && trackerutils.isTracker(req.documentUrl) && sender.frameId !== 0) {
-                action.isTrackerFrame = true
-            }
-
             action.isThirdParty = !utils.isFirstParty(sender.url, sender.tab.url)
             action.shouldBlock = !cookieConfig.isExcluded(sender.url)
+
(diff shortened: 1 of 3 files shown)


sammacbeth (commented):
Why do we not just store this data in `chrome.storage.local`? This is a simpler API and less error prone than indexeddb IMO.

SlayterDev:
@sammacbeth ready for review again.

shared/js/background/chrome-events.es6.js
```diff
@@ -389,12 +389,13 @@ chrome.runtime.onMessage.addListener((req, sender, res) => {
             const parsed = tldts.parse(tabUrl)
             action.tabRegisteredDomain = parsed.domain === null ? parsed.hostname : parsed.domain
 
+            action.isThirdParty = !utils.isFirstParty(sender.url, sender.tab.url)
+            action.shouldBlock = !cookieConfig.isExcluded(sender.url)
+
             if (req.documentUrl && trackerutils.isTracker(req.documentUrl) && sender.frameId !== 0) {
-                action.isTrackerFrame = true
+                action.isTrackerFrame = true && action.shouldBlock
```
sammacbeth, line 396: This change seems unrelated to the list caching. Did you mean to include it?

sammacbeth (commented):
Storage changes look good - I'm just a bit confused about the other changes in chrome events.

shared/js/background/chrome-events.es6.js
SlayterDev: I did. The way the content script was set up was ignoring exclusions. This will check if the url is on the exclusion list then apply it to `isTrackerFrame` accordingly. Google login for slack was broken before this change.

sammacbeth: Ok, rather than changing the semantics of `isTrackerFrame`, can you maybe change the check here to not inject if `shouldBlock` is false? https://github.com/duckduckgo/duckduckgo-privacy-extension/blob/develop/shared/js/content-scripts/block-cookie.js#L132

shared/js/content-scripts/block-cookie.js
```diff
@@ -129,8 +129,8 @@
         checkThirdParty: true,
         documentUrl: window.location.href
     }, function (action) {
-        if (!action) {
-            // response is undefined if the background has not yet registered the message listener
+        if (!action || !action.shouldBlock || !action.isThirdParty) {
+            // action is undefined if the background has not yet registered the message listener
             return
         }
         if (window.top !== window && action.isTrackerFrame) {
```
sammacbeth, line 136:
```suggestion
        if (window.top !== window && action.isTrackerFrame && action.shouldBlock) {
```

shared/js/content-scripts/block-cookie.js
```diff
@@ -129,8 +129,8 @@
         checkThirdParty: true,
         documentUrl: window.location.href
     }, function (action) {
-        if (!action) {
-            // response is undefined if the background has not yet registered the message listener
+        if (!action || !action.shouldBlock || !action.isThirdParty) {
```
sammacbeth, line 132: This should be reverted as it will break first party cookie code.

sammacbeth (changes requested):

sammacbeth (approved):
Thanks for addressing the comments!

PR #523, opened 2020-12-04
Don't overinject scripts into frames

dharb:
**Reviewer:** @Charlie-belmer @kdzwinel 

## Description:

We were injecting into all frames on every onCommitted event, which was leading to injecting multiple instances into each frame. This PR fixes that.

## Steps to test this PR:
1. Install extension
2. Navigate to onet.pl, check console. There shouldn't be any errors related to content scripts

## Automated tests:
- [ ] Unit tests
- [ ] Integration tests

###### Reviewer Checklist:
- [ ] **Ensure the PR solves the problem**
- [ ] **Review every line of code**
- [ ] **Ensure the PR does no harm by testing the changes thoroughly**
- [ ] **Get help if you're uncomfortable with any of the above!**
- [ ] Determine if there are any quick wins that improve the implementation

###### PR Author Checklist:
- [ ] Get advice or leverage existing code
- [ ] Agree on technical approach with reviewer (if the changes are nuanced)
- [ ] Ensure that there is a testing strategy (and documented non-automated tests)
- [ ] Ensure there is a documented monitoring strategy (if necessary)
- [ ] Consider systems implications

PR #591, opened 2021-04-07
Disable FLoC

kdzwinel:
**Tech Design:** https://app.asana.com/0/481882893211075/1200159597529958/f

## Description:
This disables FLoC API by opting all sites out of FLoC and removing the JS API.

## Steps to test this PR:

1. run chrome (89+) with flags that force FLoC on:
```
/Applications/Google\ Chrome\ Canary.app/Contents/MacOS/Google\ Chrome\ Canary --enable-blink-features=InterestCohortAPI --enable-features="FederatedLearningOfCohorts:update_interval/10s/minimum_history_domain_size_required/2,FlocIdSortingLshBasedComputation,InterestCohortFeaturePolicy"
```

1. clear browsing history

1. visit couple of websites

1. verify that document.interestCohort() doesn't return a cohort number (in context where we don't block the API - e.g. safelisted or any chrome:// page)

1. verify that document.interestCohort() is not available (in context where we block the API)

Make sure that:
- this works on sites that alrady have permissions-policy header (e.g. https://scotthelme.co.uk/goodbye-feature-policy-and-hello-permissions-policy/ , https://securityheaders.com/)
- permissions-policy opt-out header should be on all frames (main and sub)
- safelisting the site should bring back the JS API but the opt out header should still be there
- verify that it's not causing issues when running in Chrome with FLoC not force-enabled

## Automated tests:
- [ ] Unit tests
- [ ] Integration tests

###### Reviewer Checklist:

###### PR Author Checklist:

### DIFF
--- a/shared/js/background/chrome-events.es6.js
+++ b/shared/js/background/chrome-events.es6.js
@@ -175,26 +175,6 @@ function blockingExperimentActive () {
     // return false
 }
 
-function stripSetCookieHeadersIfNeeded (request, responseHeaders) {
-    // Strip 3rd party response header
-    const tab = tabManager.get({ tabId: request.tabId })
-    if (!request.responseHeaders) return responseHeaders
-    if (tab && tab.site.whitelisted) return responseHeaders
-    if (!tab) {
-        const initiator = request.initiator || request.documentUrl
-        if (utils.isFirstParty(initiator, request.url)) {
-            return responseHeaders
-        }
-    } else if (tab && utils.isFirstParty(request.url, tab.url)) {
-        return responseHeaders
-    }
-    if (!cookieConfig.isExcluded(request.url) && trackerutils.isTracker(request.url)) {
-        return responseHeaders.filter(header => header.name.toLowerCase() !== 'set-cookie')
-    }
-
-    return responseHeaders
-}
-
 // we determine if FLoC is enabled by testing for availability of its JS API
 const isFlocEnabled = ('interestCohort' in document)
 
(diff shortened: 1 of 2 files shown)


shared/js/background/chrome-events.es6.js
```diff
@@ -201,26 +219,32 @@ chrome.webRequest.onHeadersReceived.addListener(
...
+        }
+
         if (blockingExperimentActive()) {
-            let responseHeaders = request.responseHeaders
             // Strip 3rd party response header
             const tab = tabManager.get({ tabId: request.tabId })
-            if (!request.responseHeaders) return
-            if (tab && tab.site.whitelisted) return
+            if (!request.responseHeaders) return { responseHeaders: responseHeaders }
+            if (tab && tab.site.whitelisted) return { responseHeaders: responseHeaders }
```
kdzwinel, line 233: that's a bit ugly but I didn't want to completely rewrite that code

shared/js/background/chrome-events.es6.js
```diff
@@ -201,26 +219,32 @@ chrome.webRequest.onHeadersReceived.addListener(
...
+            responseHeaders.push({ name: 'permissions-policy', value: 'interest-cohort=()' })
+        }
+
         if (blockingExperimentActive()) {
-            let responseHeaders = request.responseHeaders
             // Strip 3rd party response header
             const tab = tabManager.get({ tabId: request.tabId })
-            if (!request.responseHeaders) return
-            if (tab && tab.site.whitelisted) return
+            if (!request.responseHeaders) return { responseHeaders: responseHeaders }
```
jonathanKingston, line 232:
```suggestion
            if (!request.responseHeaders) return { responseHeaders }
```

jonathanKingston: This also suggests sometimes we have no headers, is this on non document like requests? Like `javascript:` perhaps?

Perhaps we have to handle that above too when using responseHeaders.push?

kdzwinel: Can we safely use `{ responseHeaders }`? I haven't seen that in other parts of the code.

> This also suggests sometimes we have no headers, is this on non document like requests? Like javascript: perhaps?
> Perhaps we have to handle that above too when using responseHeaders.push?

Hm, this is good point, I'll ask Brad - he added that condition.

kdzwinel: OK we do use the `{ x }` notation in other places 👍

Brad told me that he can't recall in which cases headers were empty, but he remember this happening. I added a safeguard.

jonathanKingston (approved):
LGTM now! Thanks

shared/js/background/chrome-events.es6.js
```diff
@@ -175,6 +175,24 @@ function blockingExperimentActive () {
...
+
+// Overwrite FLoC JS API
+if (isFlocEnabled) {
+    chrome.webNavigation.onCommitted.addListener(details => {
+        const tab = tabManager.get({ tabId: details.tabId })
+        if (tab && tab.site.whitelisted) return
+
+        chrome.tabs.executeScript(details.tabId, {
+            file: 'public/js/content-scripts/floc.js',
+            allFrames: true,
```
dharb, line 189: See https://github.com/duckduckgo/duckduckgo-privacy-extension/pull/523. Confusingly, `onCommitted` is called when any frame changes location, not just the main frame.

dharb (changes requested):
Looking good! Just one tiny fix, then we should be ready to ship.

shared/js/background/chrome-events.es6.js
kdzwinel: Thanks Dave, good catch! I addressed it.

dharb (approved):
👍 💥

PR #602, opened 2021-04-14
Content Script Resiliance

jonathanKingston:
**Reviewer:**

## Automated tests:
- [ ] Unit tests
- [ ] Integration tests

###### Reviewer Checklist:

###### PR Author Checklist:


shared/js/inject/mozilla.js
```diff
@@ -0,0 +1,11 @@
+SCRIPT_TO_REPLACE
```
jonathanKingston, line 1: @sammacbeth  I couldn't think of a cleaner way to inject in the stringified code into both files. Ideall this also would be processed by rollup which would allow us to include helpers here too.

shared/js/background/chrome-events.es6.js
```diff
@@ -435,60 +448,84 @@ chrome.runtime.onMessage.addListener((req, sender, res) => {
...
+        })
+*/
+    } else {
     // Inject fingerprint protection into sites when
     // they are not whitelisted.
-    chrome.webNavigation.onCommitted.addListener(details => {
-        const tab = tabManager.get({ tabId: details.tabId })
-        if (tab && tab.site.isBroken) {
-            console.log('temporarily skip fingerprint protection for site: ' + details.url +
+        chrome.webNavigation.onCommitted.addListener(details => {
```
jonathanKingston, line 480: All this code is to be removed...

shared/js/inject/chrome.js
```diff
@@ -0,0 +1,63 @@
+/**
+ * Inject all the overwrites into the page.
```
jonathanKingston, line 2: This is the core file for protecting our code as we inject it into the page scope for chrome. We create a global function and add a password to prevent a page tampering.

shared/js/inject/chrome.js
```diff
@@ -0,0 +1,63 @@
...
+          // configurable, To allow for deletion later
+          configurable: true,
+          writable: false,
+          // Use proxy to ensure stringification isn't possible
+          value: new Proxy(function () {}, {
+              apply(target, thisArg, args) {
+                  if ('${randomPassword}' === args[0]) {
+                      protections.initProtection(args[1])
+                  } else {
+                      // TODO force enable all protections if password is wrong
```
jonathanKingston, line 41: @sammacbeth I think this is reasonable to apply protections when a page has tampered with our code right?

shared/js/content-scope/utils.js
```diff
@@ -0,0 +1,111 @@
...
+    return args.site.brokenFeatures.includes(feature)
+}
+
+// TODO make rollup aware of this so it can tree shake
+const mozProxies = 'wrappedJSObject' in window
+
+export class DDGProxy {
+    constructor (objectScope, property, proxyObject) {
+        if (mozProxies) {
+            this._native = objectScope.wrappedJSObject[property]
```
jonathanKingston, line 88: This should be the only real functional change assuming everything was implemented correctly. We use wrappedJSObjects for firefox and load it into the content script rather than an injected content script into the page.

shared/js/inject/chrome.js
sammacbeth: IMO it depends if it would be possible for them to prevent our valid call to this function.

shared/js/inject/mozilla.js
sammacbeth: Can this not be an import statement, with a preprocessor set up to which between chrome/firefox before rollup parses it?

bundle.js
```diff
@@ -0,0 +1,641 @@
+(function () {
```
sammacbeth, line 1: Did you mean to commit this? (and build/protections.js)

shared/js/inject/mozilla.js
jonathanKingston: @sammacbeth not for the chrome "SCRIPT_TO_REPLACE" it's a stringified import of the code. I'm not a huge fan of stringifying code like: `functionName.toString()` as I've seen it have different results per browser.

Perhaps I could check this further if that's still an issue.

bundle.js
jonathanKingston: Not really but there's a lot of mess like this still needing cleaning :)

shared/js/inject/chrome.js
jonathanKingston: I'm pretty sure they can given timing attacks, I think the risk here is low to always force the protections on right?

shared/js/background/chrome-events.es6.js
```diff
@@ -308,6 +308,19 @@ const browserWrapper = require('./chrome-wrapper.es6')
 chrome.runtime.onMessage.addListener((req, sender, res) => {
     if (sender.id !== chrome.runtime.id) return
 
+    if (req.registeredContentScript) {
+        const argumentsObject = getArgumentsObject(sender.tab.id)
+        if (argumentsObject.site?.isBroken) {
+            console.log('temporarily skip fingerprint protection for site: ' + sender.tab.url +
```
sammacbeth, line 314: just fingerprint, or all protections?

shared/js/background/chrome-events.es6.js
```diff
@@ -308,6 +308,19 @@ const browserWrapper = require('./chrome-wrapper.es6')
...
+        'more info: https://github.com/duckduckgo/content-blocking-whitelist')
+            return
+        }
+        if (!argumentsObject.site.whitelisted) {
```
sammacbeth, line 318: What is the difference between `isBroken` and `whitelisted`?.

Do we also need the optional chaining on `site` here? If not, we can also remove it in the above if statement.

EDIT: Looking at `getArgumentsObject` it looks like `site` will always be an object.

shared/js/background/chrome-events.es6.js
```diff
@@ -308,6 +308,19 @@ const browserWrapper = require('./chrome-wrapper.es6')
 chrome.runtime.onMessage.addListener((req, sender, res) => {
     if (sender.id !== chrome.runtime.id) return
 
+    if (req.registeredContentScript) {
+        const argumentsObject = getArgumentsObject(sender.tab.id)
+        if (argumentsObject.site?.isBroken) {
```
sammacbeth, line 313:
```suggestion
        if (argumentsObject.site.isBroken) {
```

shared/js/background/chrome-events.es6.js
```diff
@@ -426,59 +439,18 @@ chrome.runtime.onMessage.addListener((req, sender, res) => {
...
-                }
-                chrome.tabs.executeScript(details.tabId, scriptDetails)
-            } catch (e) {
-                console.log(`Failed to inject fingerprint protection into ${details.url}: ${e}`)
-            }
-        }
-    })
+function getArgumentsObject (tabId) {
+    const tab = tabManager.get({ tabId })
+    const site = tab?.site || {}
```
sammacbeth, line 444: So if the tabManager has no info on the tab, protections are on?

shared/js/content-scope/protections.js
```diff
@@ -0,0 +1,33 @@
...
+        'floc',
+        'gpc'
+    ]
+
+    initStringExemptionLists(args)
+    for (const protection of protections) {
+        const { init } = await import(`./${protection}-protection.js`)
+        if (!isFeatureBroken(args, protection)) {
+            init(args)
+        }
```
sammacbeth, line 31: If the import is `await`ed is could this cause a race condition? Or does this get made sync by the bundler?

shared/js/content-scope/utils.js
```diff
@@ -0,0 +1,156 @@
...
+            exportFunction(this.internal, objectScope, { defineAs: property })
+        } else {
+            this._native = objectScope[property]
+            const handler = {}
+            handler.apply = proxyObject.apply
+            this.internal = new window.Proxy(objectScope[property], handler)
+            objectScope[property] = this.internal
+        }
+    }
+}
```
sammacbeth, line 148: Does this need to be a class? We never assign this to a variable, nor is there any class method which we might want to use? I also don't see the object properties being used.

jonathanKingston:
Working on a fix for when the site is allow-listed, nothing is broken but it should disable itself cleaner.

shared/js/inject/chrome.js
```diff
@@ -0,0 +1,64 @@
...
+        e.textContent = `(() => {
+            ${code}
+        })();`
+        elem.appendChild(e)
+        e.remove()
+    } catch (e) {
+    }
+}
+
+inject()
```
sammacbeth, line 18: Is this here by mistake?

.gitignore
```diff
@@ -9,6 +9,7 @@ public/js/*.*
 public/css/*.*
 *.swp
 *.swo
+build/protections.js
```
sammacbeth, line 12: Do you generate this file?

sammacbeth (changes requested):

shared/js/content-scope/utils.js
jonathanKingston: this._native is used in canvas

build/config.gypi
```diff
@@ -0,0 +1,85 @@
+# Do not edit. File was generated by node-gyp's "configure" step
+{
```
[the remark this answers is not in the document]
jonathanKingston: No idea, removing

shared/js/background/chrome-events.es6.js
jonathanKingston: Old wording, changing

shared/js/background/chrome-events.es6.js
jonathanKingston: whitelisted is the existing user disabling, isBroken comes from the list

shared/js/content-scope/protections.js
jonathanKingston: Assuming we were disabling on page load, there is one tick that each of these will cause. I think it's a risk we can take but I'm happy to add a comment for that effect when we come to that? The output code essentially does `Promise.resolve().then(code)`

.gitignore
jonathanKingston: Not anymore thanks

shared/js/background/chrome-events.es6.js
jonathanKingston: I've changed this back to how it was. I'm not sure if there is rationale in either way... perhaps it solves a race condition.

shared/js/content-scope/utils.js
jonathanKingston: Added the mentioned second step `override()` method to make the "new" not have side effects and that we always use the proxy in a variable.

shared/js/content-scope/protections.js
sammacbeth: Can we do something like:
```js
protections.forEach((protection) => ...)
```
This would mean that all imports are requested on the first tick. With the async loop you currently have the last protection won't get imported until `protections.length` ticks.

Follow up, why let the import run at parse time (i.e. when we import this file)?

sammacbeth (approved):
LGTM!

PR #621, opened 2021-04-23
Require node version 14 for building

jonathanKingston:
**Reviewer:** @sammacbeth 

## Automated tests:
- [ ] Unit tests
- [ ] Integration tests

###### Reviewer Checklist:

###### PR Author Checklist:


sammacbeth:
Thanks for this @jonathanKingston. Do you think we could also add a matrix strategy to the build CI ([here's an example](https://github.com/cliqz/url-parser/blob/31c42172b0657485b199b3e8ef4affd49a8f7972/.github/workflows/tests.yml#L15-L16)) so we can keep track of which node versions work? We could start with 14 and 16.

jonathanKingston:
@sammacbeth I think we came to the conclusion we don't need the matrix strategy, unless you still do?

sammacbeth (approved):

PR #626, opened 2021-04-28
Brad/is first party fix

SlayterDev:
**Reviewer:** @sammacbeth @dharb 

## Description:

This PR switches the 3rd party check from just tldts which only checked the hostname to using entity data from TR. This should fix any issues of first party cookies being blocked where example.com is setting cookies from otherexample.com.

## Steps to test this PR:
1. Verify tests work at https://privacy-test-pages.glitch.me/privacy-protections/storage-blocking
2. See asana task for other notes

## Automated tests:
- [x] Unit tests
- [x] Integration tests

###### Reviewer Checklist:

###### PR Author Checklist:


shared/js/background/chrome-events.es6.js
```diff
@@ -166,6 +166,17 @@ function blockTrackingCookies () {
...
+/**
+ * Checks if a tracker is a first party by checking entity data
+ * @param {string} trackerUrl
+ * @param {string} siteUrl
+ * @returns {boolean}
+ */
+function isFirstPartyByEntity (trackerUrl, siteUrl, request) {
+    const trackerData = trackers.getTrackerData(trackerUrl, siteUrl, request)
+    return trackerData ? trackerData.firstParty : utils.isFirstParty(trackerUrl, siteUrl) // fall back on hostname check if trackers is null
+}
```
sammacbeth, line 178: Would be nice to have some unit-tests for this.

sammacbeth (approved):
Approving to get this hotfix out. Could you add tests in another PR?

shared/js/background/chrome-events.es6.js
```diff
@@ -166,6 +166,17 @@ function blockTrackingCookies () {
...
 
+/**
+ * Checks if a tracker is a first party by checking entity data
+ * @param {string} trackerUrl
+ * @param {string} siteUrl
+ * @returns {boolean}
+ */
+function isFirstPartyByEntity (trackerUrl, siteUrl, request) {
+    const trackerData = trackers.getTrackerData(trackerUrl, siteUrl, request)
+    return trackerData ? trackerData.firstParty : utils.isFirstParty(trackerUrl, siteUrl) // fall back on hostname check if trackers is null
```
kdzwinel, line 177: You shouldn't need to fall back here. `trackerData.firstParty` will be true for when eTLDs match because owner will be always the same. Let's drop this function and use `trackers.getTrackerData(trackerUrl, siteUrl, request).firstParty`.

kdzwinel (commented):
1. Glad that we caught it, but it seems like a big change - is that what we designed from the start, but missed?
2. Is it discussed anywhere in asana?
3. Can we add tests for it?
4. nit: could you remove those package-lock changes - they are a bit noisy and, as far as I can see, shouldn't matter.
5. Should `uitils.isFirstParty` be modified instead? Shouldn't we do entity checks in all places it's used?

kdzwinel:
Ok, I found the discussion - https://app.asana.com/0/569473074191537/1200256363978855/f . Let's follow up with a clean up PR.

shared/js/background/chrome-events.es6.js
kdzwinel: as discussed - I read it wrong, the fallback is needed for when something is not a tracker

PR #629, opened 2021-04-29
Move new method to module and unit test

SlayterDev:
**Reviewer:** @kdzwinel @sammacbeth 

## Description:

Move `isFirstPartyByEntity` to `trackerUtils` and add unit test.

## Steps to test this PR:
1. Sanity test https://privacy-test-pages.glitch.me/privacy-protections/storage-blocking
2. Run `npm run test`

## Automated tests:
- [x] Unit tests
- [ ] Integration tests

###### Reviewer Checklist:

###### PR Author Checklist:

### DIFF
--- a/shared/js/background/chrome-events.es6.js
+++ b/shared/js/background/chrome-events.es6.js
@@ -228,10 +228,10 @@ chrome.webRequest.onHeadersReceived.addListener(
             if (tab && tab.site.whitelisted) return { responseHeaders }
             if (!tab) {
                 const initiator = request.initiator || request.documentUrl
-                if (!initiator || trackerutils.isFirstPartyByEntity(initiator, request.url, request)) {
+                if (!initiator || trackerutils.isFirstPartyByEntity(initiator, request.url)) {
                     return { responseHeaders }
                 }
-            } else if (tab && trackerutils.isFirstPartyByEntity(request.url, tab.url, request)) {
+            } else if (tab && trackerutils.isFirstPartyByEntity(request.url, tab.url)) {
                 return { responseHeaders }
             }
             if (!cookieConfig.isExcluded(request.url)) {
@@ -667,10 +667,10 @@ chrome.webRequest.onBeforeSendHeaders.addListener(
             if (tab && tab.site.whitelisted) return { requestHeaders }
             if (!tab) {
                 const initiator = request.initiator || request.documentUrl
-                if (!initiator || trackerutils.isFirstPartyByEntity(initiator, request.url, request)) {
+                if (!initiator || trackerutils.isFirstPartyByEntity(initiator, request.url)) {
                     return { requestHeaders }
                 }
-            } else if (tab && trackerutils.isFirstPartyByEntity(request.url, tab.url, request)) {
+            } else if (tab && trackerutils.isFirstPartyByEntity(request.url, tab.url)) {
                 return { requestHeaders }
             }
             if (!cookieConfig.isExcluded(request.url)) {
--- a/shared/js/background/chrome-events.es6.js
+++ b/shared/js/background/chrome-events.es6.js
@@ -166,17 +166,6 @@ function blockTrackingCookies () {
     return true
 }
 
-/**
- * Checks if a tracker is a first party by checking entity data
- * @param {string} trackerUrl
- * @param {string} siteUrl
- * @returns {boolean}
- */
-function isFirstPartyByEntity (trackerUrl, siteUrl, request) {
-    const trackerData = trackers.getTrackerData(trackerUrl, siteUrl, request)
-    return trackerData ? trackerData.firstParty : utils.isFirstParty(trackerUrl, siteUrl) // fall back on hostname check if trackers is null
-}
-
 // we determine if FLoC is enabled by testing for availability of its JS API
 const isFlocEnabled = ('interestCohort' in document)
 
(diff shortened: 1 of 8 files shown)


shared/js/background/chrome-events.es6.js
```diff
@@ -678,10 +667,10 @@ chrome.webRequest.onBeforeSendHeaders.addListener(
             if (tab && tab.site.whitelisted) return { requestHeaders }
             if (!tab) {
                 const initiator = request.initiator || request.documentUrl
-                if (utils.isFirstParty(initiator, request.url)) {
+                if (trackerutils.isFirstPartyByEntity(initiator, request.url)) {
```
sammacbeth, line 670: What's the implication of not providing the `request` parameter?

shared/js/background/chrome-events.es6.js
```diff
@@ -239,10 +228,10 @@ chrome.webRequest.onHeadersReceived.addListener(
             if (tab && tab.site.whitelisted) return { responseHeaders }
             if (!tab) {
                 const initiator = request.initiator || request.documentUrl
-                if (utils.isFirstParty(initiator, request.url)) {
+                if (trackerutils.isFirstPartyByEntity(initiator, request.url)) {
```
sammacbeth, line 231: Why is `request` omitted here, but provided in the call 3 lines below?

shared/js/background/chrome-events.es6.js
```diff
@@ -548,7 +537,7 @@ chrome.runtime.onMessage.addListener((req, sender, res) => {
                 action.isTrackerFrame = true
             }
 
-            action.isThirdParty = !isFirstPartyByEntity(sender.url, sender.tab.url)
+            action.isThirdParty = !trackerutils.isFirstPartyByEntity(sender.url, sender.tab.url)
```
sammacbeth, line 540: Again `request` parameter omitted.

shared/js/background/tracker-utils.js
```diff
@@ -198,6 +198,17 @@ function truncateReferrer (referrer, target) {
     return modifiedReferrer
 }
 
+/**
+ * Checks if a tracker is a first party by checking entity data
+ * @param {string} trackerUrl
+ * @param {string} siteUrl
```
sammacbeth, line 204: The docstring does not match the function signature.

unit-test/background/tracker-utils-tests.js
```diff
@@ -627,3 +627,39 @@ describe('Tracker Utilities', () => {
...
+
+        return tdsStorage.getLists()
+            .then(lists => {
+                return tds.setLists(lists)
+            })
+    })
+
+    const firstPartyTests = [
+        { a: 'http://google-analytics.com', b: 'http://google.com', expected: true },
+        { a: 'disqus.com', b: 'google.com', expected: false }
```
sammacbeth, line 654: These should be full URLs, not just hostnames, right?

sammacbeth (commented):
I'm a bit confused about the third argument to `isFirstPartyByEntity`. It's provided about half of the time. Can it change the result if it is provided?

SlayterDev:
> I'm a bit confused about the third argument to `isFirstPartyByEntity`. It's provided about half of the time. Can it change the result if it is provided?

A couple of those were mistakes which I'll fix. The trackers module uses it to help match rules based on request type so I pass it if available. @dharb or @jdorweiler may be able to explain it better.

unit-test/background/tracker-utils-tests.js
SlayterDev: I'll make them urls for consistency but the utility function doesn't care.

shared/js/background/tracker-utils.js
```diff
@@ -198,6 +198,19 @@ function truncateReferrer (referrer, target) {
...
+ * @param {object} request
+ * @returns {boolean}
+ */
+function isFirstPartyByEntity (trackerUrl, siteUrl, request) {
+    if (!trackerUrl) return false
+    const trackerData = trackers.getTrackerData(trackerUrl, siteUrl, request)
+    return trackerData ? trackerData.firstParty : utils.isFirstParty(trackerUrl, siteUrl) // fall back on hostname check if trackers is null
```
kdzwinel, line 211: Can we please rename `isFirstParty` helper to `isSameTopLevelDomain`, so that it's super clear what it does?

kdzwinel (commented):
Thanks for doing this @SlayterDev  🎖

shared/js/background/tracker-utils.js
```diff
@@ -198,6 +198,19 @@ function truncateReferrer (referrer, target) {
...
+/**
+ * Checks if a tracker is a first party by checking entity data
+ * @param {string} trackerUrl
+ * @param {string} siteUrl
+ * @param {object} request
+ * @returns {boolean}
+ */
+function isFirstPartyByEntity (trackerUrl, siteUrl, request) {
+    if (!trackerUrl) return false
+    const trackerData = trackers.getTrackerData(trackerUrl, siteUrl, request)
```
kdzwinel, line 210: I looked a bit into this function today and it's a bit scary and not really meant for what we are trying to do here (it does rule matching, CNAME resolution, etc. that I don't think we want/need here?). Could we use `trackers.findTrackerOwner` instead?

kdzwinel: Ah wait, we may want CNAME resolution (but probably not rule matching).

sammacbeth:
After looking into this I found that it is not safe to call `getTrackerData` with `request` undefined. I added a test case for this case. As @kdzwinel states, we should not use `getTrackerData` as we should only be doing an entity check and don't need the rule matching that `getTrackerData` does.

SlayterDev:
Ok I've basically written a stripped down version of `getTrackerData` that only does the bare minimum to check the entities. Let me know what ya'll think.

unit-test/background/tracker-utils-tests.js
```diff
@@ -627,3 +627,40 @@ describe('Tracker Utilities', () => {
         tdsStorage.ClickToLoadConfig = originalConfig
     })
 })
+
+fdescribe('trackerutils.isFirstPartyByEntity()', () => {
```
sammacbeth, line 631:
```suggestion
describe('trackerutils.isFirstPartyByEntity()', () => {
```

sammacbeth (approved):
Thanks, this looks much better!

SlayterDev:
Forgot to update a unit test after method name change

PR #630, opened 2021-04-29
Remove innerHTML calls in favor of text nodes

Charlie-belmer:
This removes innerHTML calls, replacing them with textNodes. This removes the ability to insert links or styles via configuration, so changes to link text or destination will require code changes going forward.


shared/js/content-scripts/click-to-load.js
```diff
@@ -743,14 +763,14 @@
         // inherit any class styles on the button
         container.className = 'fb-login-button FacebookLogin__button'
         const styleElement = document.createElement('style')
-        styleElement.innerHTML = `
```
jonathanKingston, line 746: You could use textContent here instead.

jonathanKingston (approved):

Issue #519, opened 2020-12-03
Prevent display of post-install page when first logging into Chrome


mtconleyuk:
## Summary
When students log into our Chromebooks, DDG Privacy Extension loads and displays the post-install page (https://duckduckgo.com/app?post=1&atb=v250-2) as Chrome launches initially, which is a bit annoying. It would be nice if there were a way to disable this page, especially if we could do it from GSuite. There's no real need to remind students every time they log into their devices that DDG is there.

## Motivation
- Almost no other extensions that we load display a page (Ghostery does, but we don't force that installation; we do force uBlock, which is silent).
- Because Chromebooks aren't particularly speedy, this just adds more time to the login process, which strains the patience of students and teachers.

dharb:
Thanks for the report! This sounds annoying, and like something that we shouldn't be doing. I don't have a Chromebook but I do have a ChromeOS VM that I can use to try to repro the issue and figure out what's going on. My initial thought is that Chrome must be deleting all localstorage (including extension localstorage) upon logout, which is making our extension forget that its already shown the post install page (see https://github.com/duckduckgo/duckduckgo-privacy-extension/blob/develop/shared/js/background/atb.es6.js#L210).

Can I ask a couple quick questions to help debug?
- Is this happening on every single Chrome launch? Or is it just on the first Chrome launch after logging in?
- Does Chrome automatically open like this on login when our extension is not installed? Or does it seem like it's our extension that's making it open?
- Is there anything special about the way the DDG extension is being installed on your students' Chromebooks? Or anything special about Chrome's configuration?

mtconleyuk:
- It happens on the first launch of Chrome following a login. So it's likely that, yes, the local storage is being cleared between logins.
- When you say 'open like this' I'm not quite sure what you mean. On login, Chrome automatically launches; that's just the way it is on a Chromebook. Or, at least, that's the way we have it set (I'm not sure whether this is a GSuite setting or just the default behaviour of the OS; I could check if that helps).
- The DDG extension is being force-installed. I don't believe there are any other special considerations (we do have a number of default settings for student accounts, but other than perhaps the local storage deletion mentioned above, I don't think any of them would affect extension loading in this case).

dharb:
Gotcha, yeah it sounds like it's due to localstorage being cleared on logout. I'll take this internally and see if we can't find a fix. Will update here when I have more info.

dharb:
[...]

PR #643, opened 2021-05-07
Allow postinstall page to be disabled by managed storage

### LINKED ISSUE #567: Support chrome.storage.managed to turn off the startup page (or customize other things)
## Summary
It would be great if DDG supports chrome.storage.managed to configure the extension (in particular to turn off the start page).

## Motivation
We have folks using policy to install DDG in situations where a new profile is created every time (schools) and they see the page at every startup.

## Additional context
We document how to do this for Firefox here:

https://extensionworkshop.com/documentation/enterprise/adding-policy-support-to-your-extension/

I'd be more than happy to help.

sammacbeth:
## Description:

Allows some settings to be pre-configured by system administrators via extension managed storage. Initially this adds support for the `hasSeenPostInstall` which can be used to prevent the post install page being shown on new installs.

Following instructions from [here](https://extensionworkshop.com/documentation/enterprise/adding-policy-support-to-your-extension/), the following policy would disable the post install page:
```json
{
  "policies": {
    "3rdparty": {
      "Extensions": {
        "jid1-ZAdIEUB7XOzOJw@jetpack": {
          "hasSeenPostInstall": true
        }
      }
    }
  }
}
```

Fixes #567 and #519

## Steps to test this PR:

For Firefox:
 1. Put the above JSON into a file call `policies.json` and put it into the appropriate location for Firefox on your OS (e.g. on mac in (`Firefox.app/Contents/Resources/distribution`).
 2. Open Firefox on a new profile, open one tab to https://duckduckgo.com, then install the extension.
 3. Verify that no post install page is opened.

### DIFF
--- a/managed-storage.md
+++ b/managed-storage.md
@@ -1,0 +1,24 @@
+# Managed Storage
+
+The extension supports pre-configuration of certain options in managed environments. See documentation for [Firefox](https://extensionworkshop.com/documentation/enterprise/adding-policy-support-to-your-extension/#distributing-your-policy) and [Chrome](https://www.chromium.org/administrators/) on how to configure extension policies.
+
+The follow options can be configured via policy:
+ * `hasSeenPostInstall` (boolean) - if true the post install page will not be shown when the extension is launched for the first time.
+
+## Example
+
+Following instructions from [here](https://extensionworkshop.com/documentation/enterprise/adding-policy-support-to-your-extension/), the following policy would disable the post install page in a Firefox install:
+
+```json
+{
+  "policies": {
+    "3rdparty": {
+      "Extensions": {
+        "jid1-ZAdIEUB7XOzOJw@jetpack": {
+          "hasSeenPostInstall": true
+        }
+      }
+    }
+  }
+}
+```
--- a/shared/js/background/chrome-wrapper.es6.js
+++ b/shared/js/background/chrome-wrapper.es6.js
@@ -22,7 +22,11 @@ const getFromStorage = (key, cb) => {
 }
 const getFromManagedStorage = (keys, cb) => {
     chrome.storage.managed.get(keys, (result) => {
-        cb(result)
+        if (!browser.runtime.lastError) {
+            cb(result)
+        } else {
+            console.warn('Managed storage not available.', browser.runtime.lastError)
+        }
     })
 }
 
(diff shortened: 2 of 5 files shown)


kdzwinel:
@sammacbeth As far as I understand this won't work on Chrome because there is no JSON schema.

From https://extensionworkshop.com/documentation/enterprise/adding-policy-support-to-your-extension/:

> Unlike Chrome, Firefox does not require a managed_schema that describes your data.

From https://developer.chrome.com/docs/extensions/mv2/manifest/storage/:

> the managed storage area requires its structure to be declared as JSON Schema and is strictly validated by Chrome.

kdzwinel:
Ah, and also - should we document this somewhere in the readme / other doc? If so, lets make it a part of this PR.

shared/js/background/settings.es6.js
```diff
@@ -38,6 +42,15 @@ function buildSettingsFromLocalStorage () {
     })
 }
 
+function buildSettingsFromManagedStorage () {
+    return new Promise((resolve) => {
```
mrbaguvix, line 46: Why remove the `reject`?

managed-storage.md
```diff
@@ -0,0 +1,24 @@
+# Managed Storage
```
kdzwinel, line 1: 👏 Looks great.

shared/js/background/chrome-wrapper.es6.js
```diff
@@ -20,6 +20,11 @@ const getFromStorage = (key, cb) => {
         cb(result[key])
     })
 }
+const getFromManagedStorage = (keys, cb) => {
+    chrome.storage.managed.get(keys, (result) => {
+        cb(result)
```
kdzwinel, line 25:
```suggestion
        if (!browser.runtime.lastError) {
            cb(result)
        } else {
            console.warn('Managed storage not available.', browser.runtime.lastError)
        }
```

We need to check for error because if there is no policy manifest available Firefox is throwing this error:

> Unchecked lastError value: Error: Managed storage manifest not found

shared/js/background/settings.es6.js
kdzwinel: We are not using it in this function, eslint would complain about it if we left it.

kdzwinel (commented):
Sorry for a slow turn around. Everything looks good to me - I just found one small nit, but once addressed we can merge this.

kdzwinel (approved):
Nicely done!

PR #668, opened 2021-05-19
Inject into blank frames

jonathanKingston:
**Reviewer:** @sammacbeth 

## Description:

See: https://app.asana.com/0/inbox/1199237043596759/1200327531868645/1200327707684516

## Automated tests:
- [ ] Unit tests
- [x] Integration tests

###### Reviewer Checklist:

###### PR Author Checklist:


sammacbeth (approved):
LGTM

integration-test/content-scripts/gpc.js
```diff
@@ -0,0 +1,89 @@
+/**
```
kdzwinel, line 1: Should we add this to privacy test pages? I guess other platforms could also use such a test.

sammacbeth (commented):
1. You have a lint error to fix.
 2. It seems there is not a way to add a site exception just for about:blank, it can only be done for all frames. Do you think that gap is an issue, or that when we have an issue with an about:blank frame it's ok to disable the protection for the whole site?

jonathanKingston:
> It seems there is not a way to add a site exception just for about:blank, it can only be done for all frames. Do you think that gap is an issue, or that when we have an issue with an about:blank frame it's ok to disable the protection for the whole site?

I've allowed them to individually be turned off and also a site list too. Both keys don't need to be present in the format also now so we could ditch the update here: https://github.com/duckduckgo/content-blocking-lists/pull/35

sammacbeth (approved):

Issue #680, opened 2021-05-25
Cookie causes Plugin to freeze Chrome tab


zone5Per:
## Description
First of all, I have already lodged a report using the 'Report Broken Site' link on the extension popup. But on further investigation on my end I think this is actually an issue with the Chrome plugin rather than just a broken site.

When accessing one of our web applications the DuckDuckGo plugin causes the tab to become unresponsive. Only a restart of Chrome or manually killing the Google Chrome Helper (Renderer) task in Activity Monitor that is consuming a very high amount of CPU solves it. 

The issue does not appear when turning off the plugin, site protection or running the plugin in Incognito mode. 

## Steps to Reproduce
1. Go to https://portal.zone5cloud.com/en/sign-up and sign up for a new account, ensure you tick the **"remember me"** checkbox
2. Once logged in either click on your name in the top right corner of screen and select account or manually navigate too https://portal.zone5cloud.com/en/manager/account
3. Page freezes, either restart Chrome or kill Google Chrome Helper (Renderer) in activity monitor to regain control

**Expected behavior:**
User should be navigated to user account page.

**Actual behavior:**
The tab freezes, only by killing the process can you regain control again.

**Versions**
- Extension: 2021.5.10
- Browser: Chrome Version 90.0.4430.212 (Official Build) (x86_64)
- OS: macOS Catalina 10.15.6 (19G2021)

## Additional Information
1. Does not happen in Incognito mode
2. Does not happen when turning off site protection
3. On login we store an email in either session/Cookie storage depending on if user ticked the remember me checkbox or not, but it only breaks when using Cookie storage. By not setting this email cookie it starts working as expected again. The only difference between the session/Cookie storage code is that I use a service to set the Cookie which automatically URI encodes the email when setting it while it is stored without encoding in session storage. By manually setting the Cookie value without the library so it is stored without being URI encoded it seems to work as expected. Like so:

sessionStorage - per@email.com <-- does not cause issue
cookie - per%40email.com <-- DuckDuckGo freezes page
cookie - per@email.com <-- Page loads as expected, this scenario not testable on the website because requires a code change.

For some reason it does not seem to like that encoding of the email in the Cookie value. Also it only seems to be an issue with this particular cookie and not the rest, even though we are setting them in the same way.

[...]

PR #682, opened 2021-05-26
Prevent polyfill side-effects from 1p cookie protection

sammacbeth:
**Reviewer:**

## Description:
When we load a content-scope script on a page, after the first tick any global could be polluted or changed by other scripts in the page. This can break some of our protection implementations.

In the case of #680, the page loaded a polyfill which changed `Promise` and `Promise.prototype.then`. This then caused an issue when we subsequently called the polyfilled `Promise.then` with a native `Promise` instance.

The PR fixes this specific case by creating the `then` function during the script's first tick, where we are sure (are we?) that we have the native `Promise.prototype.then`.

## Steps to test this PR:
See #680 

## Automated tests:
- [ ] Unit tests
- [ ] Integration tests

###### Reviewer Checklist:

###### PR Author Checklist:

### DIFF
--- a/shared/js/content-scope/cookie-protection.js
+++ b/shared/js/content-scope/cookie-protection.js
@@ -26,6 +26,9 @@ function applyCookieExpiryPolicy () {
     const loadPolicy = new Promise((resolve) => {
         loadedPolicyResolve = resolve
     })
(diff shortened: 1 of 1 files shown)


jonathanKingston (approved):
Makes me sad this is a thing, but LGTM thanks.

PR #692, opened 2021-06-03
Remove remote styling & fix comment issue on some sites

Charlie-belmer:
This removes the usage of the style data in remote config, and also fixes an issue where some comment sections don't load after enabling (due to a FB update)

### DIFF
--- a/shared/data/web_accessible_resources/fb-sdk.js
+++ b/shared/data/web_accessible_resources/fb-sdk.js
@@ -5,6 +5,7 @@
     let siteInit = function () {}
     let fbIsEnabled = false
     let initData = {}
+    let runInit = false
     const parseCalls = []
     const popupName = Math.random().toString(36).replace(/[^a-z]+/g, '').substr(0, 12)
 
@@ -30,7 +31,7 @@
             window.FB = undefined
 
             window.fbAsyncInit = function () {
-                if (initData) {
+                if (runInit && initData) {
                     window.FB.init(initData)
                 }
                 siteInit()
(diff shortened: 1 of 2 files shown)


shared/data/web_accessible_resources/fb-sdk.js
```diff
@@ -30,7 +31,7 @@
             window.FB = undefined
 
             window.fbAsyncInit = function () {
-                if (initData) {
+                if (runInit && initData) {
```
jonathanKingston, line 34: Sorry I just want to ensure I understand:
The callstack is:
- `window.FB.init`
- `window.fbAsyncInit`

Where `window.FB.init` sets `runInit = true`? I'm a little lost as to how the code will ever be triggered, is this the page doing the work?

Or should this code be `ranInit` instead of 'run' and be:
```suggestion
                if (!ranInit && initData) {
```

Charlie-belmer: Not exactly. When setting up the Facebook SDK, the site may define a function called `window.fbAsyncInit`. Once the SDK loads, it searches for and calls `window.fbAsyncInit`. However, some sites may not use this, and just call `FB.init` directly at some point (after ensuring that the script has loaded).

What our surrogate does is capture calls made to init by page scripts. If at a later point we load the real sdk, we then re-call init with whatever arguments the page passed in originally. This runInit param should be true when a page has called init directly. Because we put it in asyncInit, the flow will be something like:

FB SDK loads -> SDK calls window.fbAsyncInit -> Our function calls window.FB.init (maybe) -> our function calls original fbAsyncInit (if it existed)

jonathanKingston: Ah I see, would it make it simpler to understand if `siteInit = window.fbAsyncInit` was captured before defining this function here?

Also maybe verbatim dunk this comment above into the code?

Charlie-belmer: Good call on the comment, added it to the overall function. siteInit is captured when the surrogate initializes, when we also have to call the original so the site can do whatever it was planning to do with the SDK.

jonathanKingston (approved):
LGTM thanks for clarifying things!

PR #703, opened 2021-06-25
La/ref tests

ladamski:
**Reviewer:** @jdorweiler 

## Description:
Implements initial reference tests from https://github.com/duckduckgo/privacy-reference-tests

## Steps to test this PR:
1.  `git submodule update --init --recursive`
2.  `npm test`

## Automated tests:
- [ ] Unit tests
- [ ] Integration tests

###### Reviewer Checklist:

###### PR Author Checklist:


unit-test/background/tracker-reference-tests.js
```diff
@@ -0,0 +1,47 @@
...
+            const reason = result && result.reason
+            if (action === test.expectAction) {
+                console.log(`PASS ${test.name}`)
+            } else {
+                console.log('FAIL Test Case', test.name, requestURL, rootURL)
+                console.log(`-- expected ${test.expectAction}, got ${action}, reason ${reason}`)
+            }
+            expect(test.expectAction).toEqual(action)
+        }
+    })
```
sammacbeth, line 46: Instead of a single test with assertions for each of the domain tests, could we have a separate `it` for each test? This would allow the `it` to contain the test description so it is easier to locate which is failing.

CONTRIBUTING.md
```diff
@@ -65,6 +65,7 @@ Run the dev build task for your browser from the 'Build' section above. The gene
 After running the build task it will continue watching for changes to any of the source files. After saving any changes to these files it will automatically rebuild the `dev` directory for you.
 
 ### Testing
+- Install the required test submodules: `git submodule update --init --recursive`
```
sammacbeth, line 68: This is already done in the postinstall step, does this need to also be explicitly run in the testing instructions?

.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "unit-test/background/reference-tests"]
+	path = unit-test/background/reference-tests
+	url = https://github.com/duckduckgo/privacy-reference-tests
```
sammacbeth, line 3: Can this by pinned to a specific tag or commit? I'd prefer updates to the tests to be explicit, otherwise a commit with passing tests could break in the future if the tests are updated.

Alternatively, could we publish the tests on npm (e.g. in the @duckduckgo namespace). This would allow us to pin the version, and newer versions of the tests would trigger dependabot to automatically open a pull request.

unit-test/background/tracker-reference-tests.js
jdorweiler: 👍 agreed, it takes some work to figure out which test failed otherwise.

CONTRIBUTING.md
ladamski: I added the postinstall after the above, but its probably still worth calling out if you have an existing install

.gitmodules
ladamski: This question was raised in the tech design and its definitely an interesting tradeoff.  There are a few intended mitigations:
- add a platform exception to the test definition.  This isn't yet supported the extension currently as its the reference implementation but looks like we'll have to have one for iOS at least at launch.  This is better than pinning because it makes it explicit which test is disabled for a given platform, and you still get the benefit of all of the other test being up to date.  My experience with pinning to a commit is that it tends to be a temporary stopgap that often never gets removed because its not obvious what the issue is that caused the pin in the first place.
- pinning to a commit is a fallback but we should not be changing ref tests without validating them across our products beforehand, and adding any unsupported platforms to the exception list before the new tests are merged.

It sounds like a more specific concern is around tests changing for features that are in progress.  I would argue there though that its the lesser of two evils.  If the tests are validated across main/master on each product, and a particular WIP branch goes red once they are merged to the test repo, I would rather that branch go red before being merged (vs after) as it has regressed something?

jdorweiler (approved):

sammacbeth (approved):

PR #704, opened 2021-06-28
Fix counting of trackers

jonathanKingston:
Ensure we count broken site trackers and displaying this to users.

https://app.asana.com/0/0/1200517965702747/f

**Reviewer:** @sammacbeth 

## Automated tests:
- [ ] Unit tests
- [ ] Integration tests

###### Reviewer Checklist:

###### PR Author Checklist:

### DIFF
--- a/shared/js/background/redirect.es6.js
+++ b/shared/js/background/redirect.es6.js
@@ -81,16 +81,6 @@ function handleRequest (requestData) {
          */
         if (!(thisTab && thisTab.url && thisTab.id)) return
 
-        /**
-         * skip any broken sites
-         */
-        if (thisTab.site.isBroken) {
-            console.log('temporarily skip tracker blocking for site: ' +
-              utils.extractHostFromURL(thisTab.url) + '\n' +
-              'more info: https://github.com/duckduckgo/content-blocking-lists')
-            return
-        }
-
         // skip blocking on new tab and extension pages
         if (thisTab.site.specialDomainName) {
             return
@@ -154,9 +144,8 @@ function handleRequest (requestData) {
             }
 
             browserWrapper.notifyPopup({ updateTabData: true })
-
-            // Block the request if the site is not whitelisted
-            if (!thisTab.site.whitelisted && tracker.action.match(/block|redirect/)) {
+            // Block the request if the site is not allowlisted
+            if (!thisTab.site.isAllowlisted() && tracker.action.match(/block|redirect/)) {
                 // update badge icon for any requests that come in after
                 // the tab has finished loading
                 if (thisTab.status === 'complete') thisTab.updateBadgeIcon()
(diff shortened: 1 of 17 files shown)


shared/js/background/redirect.es6.js
```diff
@@ -154,9 +144,9 @@ function handleRequest (requestData) {
             }
 
             browserWrapper.notifyPopup({ updateTabData: true })
-
-            // Block the request if the site is not whitelisted
-            if (!thisTab.site.whitelisted && tracker.action.match(/block|redirect/)) {
+            const allowListed = thisTab.site.whitelisted || thisTab.site.isBroken;
```
sammacbeth, line 147: Can we not use `thisTab.site.isAllowlisted` directly here?

shared/js/ui/templates/shared/tracker-networks-text.es6.js
```diff
@@ -17,7 +17,7 @@ module.exports = function (site, isMajorNetworksCount) {
 
 function trackersBlockedOrFound (site, trackersCount) {
     let msg = ''
-    if (site && (site.isWhitelisted || trackersCount === 0)) {
+    if (site && site.isAllowlisted || trackersCount === 0) {
```
sammacbeth, line 20: So this means that when `site` is falsy we also say "Found", while before we said "Blocked", right? In which cases is `site` falsy? I guess one example might be when the extension reloads while a tab is open. Will we not be blocking requests for the tab in that case?

shared/js/ui/views/site.es6.js
```diff
@@ -17,6 +17,7 @@ function Site (ops) {
 
     // cache 'body' selector
     this.$body = window.$('body')
+    this.$protectionRow = window.$('body')
```
sammacbeth, line 20: This assignment seems to be redundant, `this.$protectionRow` is identical to `this.$body`

shared/scss/popup.scss
```diff
@@ -167,7 +167,7 @@ body {
             max-height: 19px;
             overflow: hidden;
 
-            .is-disabled & {
+            .is-disabled &, &.is-disabled {
```
sammacbeth, line 170: Why do we need both of these versions?

sammacbeth (changes requested):
A few comments, and you have some lint errors to fix.

shared/js/ui/templates/shared/tracker-networks-text.es6.js
jonathanKingston: We should be at least once the tds list is loaded. I think this has the potential to be the case after install, I'm not super familiar with when we trigger `site` calculation, I think it's on navigation from memory.

I've made the brackets match how they were anyway.

sammacbeth (approved):

PR #710, opened 2021-07-05
Fix audio buffer permission denied in Firefox.

### LINKED ISSUE #708: Permission denied in audio-protection.js
## Description
The call to `.reduce()` on `channelData` in `audio-protection.js` produces a permission denied error in Firefox 89. This prevents the audio anti-fingerprinting functionality from working on Firefox.

https://github.com/duckduckgo/duckduckgo-privacy-extension/blob/e871aeafcd311e42996e6aadb0e52de53e38dbd1/shared/js/content-scope/audio-protection.js#L11-L13

## Steps to Reproduce

1. Open a new Firefox profile with only DuckDuckGo Privacy Essentials installed.
2. Open the developer console and execute:
```js
var ab = new AudioBuffer({length: 64, numberOfChannels: 1, sampleRate: 48000});
var abb = ab.getChannelData(0);
abb.reduce((sum, v) => {return sum + v}, 0);
```
3.Notice that the sum is zero due to `transformArrayData` returning early due to a thrown exception.

**Expected behavior:**
4. Repeat the above in Chrome. Notice that the result is non-zero due to the noise added by anti-fingerprinting.

**Actual behavior:**
5. Adding a `try`/`catch` block to log the error will display the following in the console:
```
Error: Permission denied to access object
```

**Versions**

- Extension: 2021-06-02 and Git `develop` branch.
- Browser: Firefox 89
- OS: Linux

## Additional Information

Using an alternate method to sum the values appears to work in both Firefox and Chrome. For example:
```js
let cdSum = 0;
for (var k in channelData) {
    cdSum += channelData[k];
}
```
---
Unrelated note: adding noise to a new (all-zero) `AudioBuffer ` leaks data about whether the addon is installed. For example, when the result of this JavaScript is non-zero, it's clear that noise is being added to the audio buffer:
```js
var ab = new AudioBuffer({length: 64, numberOfChannels: 1, sampleRate: 48000});
var abb = ab.getChannelData(0);
if (abb.reduce((sum, v) => {return sum + v}, 0) != 0) {
  console.log("DDG installed");
}
```

jonathanKingston:
**Reviewer:** @sammacbeth 

## Description:
Fixes #708
This fixes the permission denied issue which is breaking the method and failing gracefully (like it should). For blank input we will now skip adding noise. However we should investigate further / switch to the computing of blank data to fully match canvas.

## Automated tests:
- [ ] Unit tests
- [ ] Integration tests

###### Reviewer Checklist:

###### PR Author Checklist:

### DIFF
--- a/shared/js/content-scope/audio-protection.js
+++ b/shared/js/content-scope/audio-protection.js
@@ -8,9 +8,14 @@ export function init (args) {
     function transformArrayData (channelData, domainKey, sessionKey, thisArg) {
         let { audioKey } = getCachedResponse(thisArg, args)
         if (!audioKey) {
-            const cdSum = channelData.reduce((sum, v) => {
-                return sum + v
-            }, 0)
+            let cdSum = 0
+            for (const k in channelData) {
+                cdSum += channelData[k]
+            }
+            // If the buffer is blank, skip adding data
+            if (cdSum === 0) {
+                return
+            }
             audioKey = getDataKeySync(sessionKey, domainKey, cdSum)
             setCache(thisArg, args, audioKey)
         }


sammacbeth (commented):
Looks good, you just have lint errors to fix :)

PR #718, opened 2021-07-06
Debugger panel

sammacbeth:
**Reviewer:**

## Description:

Adds a debugger page to the extension that shows requests blocked/ignored/redirected, cookies blocked and usaged of JS cookie and canvas APIs:

Screenshot 2021-07-06 at 12 51 34

It also allows protections and block/ignore rules to be toggled on/off, as well as switching between different versions of the blocklist.

There is also an editor that allows manual editing of blocklists and protection configuration files:

Screenshot 2021-07-06 at 12 53 59

## Automated tests:
- [ ] Unit tests
- [ ] Integration tests

###### Reviewer Checklist:

###### PR Author Checklist:


shared/js/background/channel.es6.js
```diff
@@ -0,0 +1,11 @@
+
+const RELEASE_EXTENSION_IDS = [
+    'caoacbimdbbljakfhgikoodekdnlcgpk', // edge store
+    'bkdgflcldnnnapblkhphbgpggdiikppg', // chrome store
+    'jid1-ZAdIEUB7XOzOJw@jetpack' // firefox
+]
+const IS_BETA = RELEASE_EXTENSION_IDS.indexOf(chrome.runtime.id) === -1
```
jdorweiler, line 7: 👍

shared/js/ui/views/hamburger-menu.es6.js
```diff
@@ -21,16 +22,22 @@ HamburgerMenu.prototype = window.$.extend({},
...
                 [this.$optionslink, 'click', this.openOptionsPage],
                 [this.$feedbacklink, 'click', this._handleFeedbackClick],
                 [this.$brokensitelink, 'click', this._handleBrokenSiteClick],
                 [this.model.store.subscribe, 'action:search', this._handleAction],
-                [this.model.store.subscribe, 'change:site', this._handleSiteUpdate]
+                [this.model.store.subscribe, 'change:site', this._handleSiteUpdate],
+                [this.$debuggerpanellink, 'click', this._handleDebuggerClick]
             ])
+            if (IS_BETA) {
+                console.log(this.$('#debugger-panel'))
```
jonathanKingston, line 38: This log isn't needed right? I see some others too that might be not relevant.

shared/html/devtools-panel.html
```diff
@@ -0,0 +1,176 @@
...
+    th,
+    td {
+      padding: 5px;
+      white-space: nowrap;
+      text-overflow: ellipsis;
+      overflow: hidden;
+    }
+
+    tr > td:nth-child(4) {
+      overflow: visible;
```
jonathanKingston, line 88: Allowing this to wrap works better on smaller screens without colliding with the next column. For the columns that use ellipsis, you could use title so it's still readable on hover perhaps.

```suggestion
word-wrap: anywhere;
overflow: auto;
white-space: break-spaces;
```

shared/js/content-scope/canvas-protection.js
```diff
@@ -1,13 +1,20 @@
...
 export function init (args) {
-    const { sessionKey, site } = args
+    const { sessionKey, site, debug } = args
     const domainKey = site.domain
 
     // Using proxies here to swallow calls to toString etc
     const getImageDataProxy = new DDGProxy(CanvasRenderingContext2D.prototype, 'getImageData', {
         apply (target, thisArg, args) {
+            if (debug) {
+                postDebugMessage('canvas', {
```
jonathanKingston, line 12: Could this be contained in `shouldExemptMethod` that way all implementations should get it and it allow us to toggle this too? Happy if we need to rename the method to explain the debugging.

The performance cost shouldn't be any different.

shared/js/devtools/panel.es6.js
```diff
@@ -0,0 +1,217 @@
...
+const refreshButton = document.getElementById('refresh')
+const protectionButton = document.getElementById('protection')
+const tabPicker = document.getElementById('tab-picker')
+const tdsOption = document.getElementById('tds')
+const displayFilters = document.querySelectorAll('#table-filter input')
+
+let tabId = chrome.devtools?.inspectedWindow?.tabId || parseInt(0 + new URL(document.location.href).searchParams.get('tabId'))
+const port = chrome.runtime.connect()
+const features = [
+    'canvas',
```
jonathanKingston, line 12: These will need to be renamed since brad's project has gone live. Perhaps there should be a mapping of short name to the long names?

Like this: https://github.com/duckduckgo/privacy-configuration/pull/1/files#diff-e727e4bdf3657fd1d798edcd6b099d6e092f8573cba266154583a746bba0f346R53-R65

jdorweiler:
@sammacbeth Could the devtools buttons update when you click them or is that complicated? Looks great either way.

sammacbeth:
> @sammacbeth Could the devtools buttons update when you click them or is that complicated? Looks great either way.

This is actually a bit tricky because with how protections are injected the change only takes effect after refreshing. Toggling could trigger the refresh automatically, but that may make it a bit cumbersome.

jonathanKingston:
> This is actually a bit tricky because with how protections are injected the change only takes effect after refreshing. Toggling could trigger the refresh automatically, but that may make it a bit cumbersome.

Maybe having some feedback that it would be disabled on refresh?
**Or** the toggle could reflect what a navigation/refresh would inject and the log could have when protections are injected/enabled in the page?

jdorweiler (approved):

shared/js/background/utils.es6.js
```diff
@@ -155,7 +155,14 @@ function getAsyncBlockingSupport () {
 */
 function isBroken (url) {
     if (!tdsStorage?.config.unprotectedTemporary) return
-    return isBrokenList(url, tdsStorage?.config.unprotectedTemporary)
+    return isBrokenList(url, tdsStorage?.config.unprotectedTemporary) !== -1
```
jonathanKingston, line 158: Can we rename `isBrokenList` to `brokenListIndex` or similar?

shared/js/background/storage/tds.es6.js
```diff
@@ -217,5 +225,27 @@ class TDSStorage {
 
         return versionParam
     }
+
+    /**
+     * Convert the given list into stringified form.
+     * @param {*} name
+     * @returns list in a fully serialisable format
+     */
+    getSerialisableList (name) {
```
jonathanKingston, line 234: We should probably default to US English.

docs/protection-debugger.md
```diff
@@ -0,0 +1,63 @@
+# Protection Debugging
```
jonathanKingston, line 1: 🔥 thanks for making this page.

jonathanKingston (changes requested):
2 nits and I think we're good to merge 👍🏻

jonathanKingston (approved):
🥇

PR #723, opened 2021-07-06
Remote Configuration

SlayterDev:

## Description:

Migrate the extension to the new remote config

## Steps to test this PR:
1. Go through all the tests in http://privacy-test-pages.glitch.me/privacy-protections/index.html
2. All tests should appear the same as develop branch

## Automated tests:
- [x] Unit tests
- [x] Integration tests

###### Reviewer Checklist:

###### PR Author Checklist:


shared/js/background/utils.es6.js
```diff
@@ -230,6 +236,30 @@ function isSafeListed (url) {
     return false
 }
 
+function isCookieExcluded (url) {
+    const domain = (new URL(url)).host
+    return isDomainCookieExcluded(domain)
+}
+
+function isDomainCookieExcluded (domain) {
+    const excludeList = configStorage.config.features?.cookieProtections.exceptions
```
jonathanKingston, line 245: Protections has "cookie" which will be needed still as protections.js lists out the features it expects (do-not-track is being removed). Perhaps combining both `cookie` and `cookieProtections`

shared/js/background/chrome-events.es6.js
```diff
@@ -167,12 +167,12 @@ const redirect = require('./redirect.es6')
...
 const pixel = require('./pixel.es6')
 const https = require('./https.es6')
-const cookieConfig = require('./../background/storage/cookies.es6')
+const configStorage = require('./../background/storage/config.es6')
 
 const requestListenerTypes = utils.getUpdatedRequestListenerTypes()
 
 function blockTrackingCookies () {
-    return true
+    return configStorage.config.features?.cookieProtections.state === 'enabled' || true
```
jonathanKingston, line 175: I think this will always resolve true.

SlayterDev: Yeah I was afraid of this. I'll fix this one.

shared/js/content-scope/protections.js
```diff
@@ -24,7 +24,7 @@ export async function loadProtections () {
         'audio',
         'battery',
         'canvas',
-        'cookie',
```
jonathanKingston, line 27: Can we use `-` dashes to make it consistent? Or change all the others to upper case?

jonathanKingston: Alternatively we could have the keys as uppercase and the files as dasherised. I'm indifferent so long as it's consistent though :).

shared/js/background/chrome-events.es6.js
```diff
@@ -863,13 +872,11 @@ chrome.alarms.onAlarm.addListener(alarmEvent => {
...
             .catch(e => console.log(e))
+        configStorage.updateConfigData()
     } else if (alarmEvent.name === 'clearExpiredHTTPSServiceCache') {
         httpsService.clearExpiredCache()
     } else if (alarmEvent.name === 'updateUserAgentData') {
-        settings.ready()
-            .then(() => {
-                cookieConfig.updateCookieData()
-            }).catch(e => console.log(e))
+        // TODO: Reinstate user agent rotation
```
jonathanKingston, line 879: Is this to be added after or in this PR?

shared/js/background/storage/config.es6.js
```diff
@@ -0,0 +1,74 @@
+// Babel includes for async/await
+import 'regenerator-runtime/runtime'
+
+const load = require('./../load.es6')
+const constants = require('../../../data/constants')
+const settings = require('./../settings.es6')
+
+/**
+ *  Manage local cookie exclusion data.
```
jonathanKingston, line 9: Perhaps?

```suggestion
 *  Manage local config exclusion data.
```

shared/js/background/storage/config.es6.js
```diff
@@ -0,0 +1,74 @@
...
+ *  Data is updated regularly with the latest
+ *  cookie exclusions, and stored here for fast retrieval and usage
```
jonathanKingston, line 11:
```suggestion
 *  exclusions, and stored here for fast retrieval and usage
```

shared/js/background/storage/config.es6.js
```diff
@@ -0,0 +1,74 @@
...
+ *  when checking which cookies to block.
+ **/
+class ConfigStorage {
+    constructor () {
+        this.config = {}
+    }
+
+    /**
+     * Retrieve the latest lists of excluded cookie domains. Store as needed.
```
jonathanKingston, line 20:
```suggestion
     * Retrieve the latest config. Store as needed.
```

shared/js/background/storage/config.es6.js
```diff
@@ -0,0 +1,74 @@
...
+    updateConfigData () {
+        console.log('Config: Getting config data')
+        const extensionUrlData = constants.ExtensionConfig
+        const source = extensionUrlData.source || 'external'
+        const configName = extensionUrlData.name
+        const etag = settings.getSetting(`${configName}-etag`) || ''
+        load.loadExtensionFile({ url: extensionUrlData.url, etag: etag, returnType: extensionUrlData.format, source, timeout: 60000 })
+            .then(response => {
+                if (response && response.status === 200) {
+                    // New cookie data to process.
```
jonathanKingston, line 31:
```suggestion
                    // New data to process.
```

shared/js/background/chrome-events.es6.js
SlayterDev: On second thought this isn't needed. If we bring the user agent stuff back we'll put it in the config which is already loaded elsewhere.

shared/js/background/utils.es6.js
```diff
@@ -154,32 +155,32 @@ function getAsyncBlockingSupport () {
...
 }
 
 function getBrokenFeaturesAboutBlank (url) {
-    if (!tdsStorage?.protections) return
+    if (!configStorage.config.features) return
     const brokenFeatures = []
-    for (const feature in tdsStorage.protections) {
-        if (tdsStorage.protections[feature]?.aboutBlankEnabled === false) {
+    for (const feature in configStorage.config.features) {
+        if (configStorage.config.features[feature]?.aboutBlankEnabled === 'disabled') {
```
jonathanKingston, line 166: Maybe this should be changed to `aboutBlankState` for parity.

shared/js/background/chrome-events.es6.js
```diff
@@ -167,11 +167,15 @@ const redirect = require('./redirect.es6')
...
 const pixel = require('./pixel.es6')
 const https = require('./https.es6')
-const cookieConfig = require('./../background/storage/cookies.es6')
+const configStorage = require('./../background/storage/config.es6')
 
 const requestListenerTypes = utils.getUpdatedRequestListenerTypes()
 
 function blockTrackingCookies () {
+    if (configStorage.config.features) {
+        return configStorage.config.features.trackingCookies.state === 'enabled'
```
sammacbeth, line 176: Should we trust the config format to always have the expected structure under `features`? If the format changed, or a broken version of the list got shipped then this line could throw, effectively disabling cookie protection. Can we instead fail to a less broken state?

shared/js/background/storage/config.es6.js
```diff
@@ -0,0 +1,74 @@
...
+const constants = require('../../../data/constants')
+const settings = require('./../settings.es6')
+
+/**
+ *  Manage local cookie exclusion data.
+ *  Data is updated regularly with the latest
+ *  cookie exclusions, and stored here for fast retrieval and usage
+ *  when checking which cookies to block.
+ **/
+class ConfigStorage {
```
sammacbeth, line 14: What's the reasoning behind creating a new class and loading logic for this file when tdsStorage is already able to load and persist files to disk (including protections.json which this replaces)?

shared/js/background/storage/config.es6.js
```diff
@@ -0,0 +1,74 @@
...
+    /**
+     * Retrieve the latest lists of excluded cookie domains. Store as needed.
+     */
+    updateConfigData () {
+        console.log('Config: Getting config data')
+        const extensionUrlData = constants.ExtensionConfig
+        const source = extensionUrlData.source || 'external'
+        const configName = extensionUrlData.name
+        const etag = settings.getSetting(`${configName}-etag`) || ''
+        load.loadExtensionFile({ url: extensionUrlData.url, etag: etag, returnType: extensionUrlData.format, source, timeout: 60000 })
```
sammacbeth, line 28: How does this update differ to tds `getLists`? Could we share/reuse some code here?

shared/js/background/storage/config.es6.js
```diff
@@ -0,0 +1,74 @@
...
+
+/**
+ *  Manage local cookie exclusion data.
+ *  Data is updated regularly with the latest
+ *  cookie exclusions, and stored here for fast retrieval and usage
+ *  when checking which cookies to block.
+ **/
+class ConfigStorage {
+    constructor () {
+        this.config = {}
```
sammacbeth, line 16: We could use a `Proxy` here to provide a default structure to `config.features` (like a `defaultdict` in Python) when the config is not loaded, or if an attribute is missing from the config. This could guarantee safe access to `config.feature.myFeature.state` at all times.

jonathanKingston (approved):

jonathanKingston:
Nice thanks for cleaning this all up!

shared/js/background/chrome-events.es6.js
```diff
@@ -167,11 +167,14 @@ const redirect = require('./redirect.es6')
 const tabManager = require('./tab-manager.es6')
 const pixel = require('./pixel.es6')
 const https = require('./https.es6')
-const cookieConfig = require('./../background/storage/cookies.es6')
 
 const requestListenerTypes = utils.getUpdatedRequestListenerTypes()
 
 function blockTrackingCookies () {
+    if (tdsStorage.config.features) {
```
kdzwinel, line 174: There was no way to disable third party cookies before?

shared/js/background/chrome-events.es6.js
```diff
@@ -716,9 +723,10 @@ if (chrome.webRequest.OnBeforeSendHeadersOptions.EXTRA_HEADERS) {
 chrome.webRequest.onBeforeSendHeaders.addListener(
     request => {
         const GPCHeader = GPC.getHeader()
+        const GPCEnabled = tdsStorage.config.features?.gpc.state === 'enabled'
```
kdzwinel, line 726: We are doing this check quite a lot (`tdsStorage.config.features?.gpc.state`) - can we move it to a helper function e.g.`isFeatureEnabled('name') => bool` ?

shared/js/background/chrome-events.es6.js
```diff
@@ -716,9 +723,10 @@ if (chrome.webRequest.OnBeforeSendHeadersOptions.EXTRA_HEADERS) {
...
 
         let requestHeaders = request.requestHeaders
-        if (GPCHeader) {
+        if (GPCHeader && GPCEnabled) {
```
kdzwinel, line 729: ditto, there was no way to disable GPC before?

shared/js/background/chrome-events.es6.js
```diff
@@ -850,10 +858,7 @@ chrome.alarms.onAlarm.addListener(alarmEvent => {
     } else if (alarmEvent.name === 'clearExpiredHTTPSServiceCache') {
         httpsService.clearExpiredCache()
     } else if (alarmEvent.name === 'updateUserAgentData') {
```
kdzwinel, line 860: I don't know when/if user agent rotation is coming back, can we drop this if and corresponding chrome.alarm ?

shared/js/background/redirect.es6.js
```diff
@@ -124,7 +126,7 @@ function handleRequest (requestData) {
         }
 
         // count and block trackers. Skip things that matched in the trackersWhitelist unless they're first party
-        if (tracker && !(tracker.action === 'ignore' && tracker.reason !== 'first party')) {
```
kdzwinel, line 127: have we tested what will happen when blocking is disabled? Is this going to trigger "this page is temporary disabled" banner in the popup? Or will it look like it's blocking stuff or will it say that there are no trackers? I think we may need to understand that if we ever have to disable it.

shared/js/background/storage/tds.es6.js
```diff
@@ -12,8 +12,81 @@ class TDSStorage {
         })
         this.tds = { entities: {}, trackers: {}, domains: {}, cnames: {} }
         this.surrogates = ''
-        this.brokenSiteList = []
-        this.protections = {}
+        this.config = {
```
kdzwinel, line 15: why we decided to have a built in fallback? protections didn't have one

shared/js/background/tracker-utils.js
```diff
@@ -32,12 +32,13 @@ function isTracker (url) {
...
 /**
- * Determine if the social entity should be blockedon this URL. returns True if so.
+ * Determine if the social entity should be blocked on this URL. returns True if so.
  */
 function shouldBlockSocialNetwork (entity, url) {
+    const ctpEnabled = tdsStorage.config.features?.clickToPlay.state === 'enabled'
     const domain = tldts.parse(url).domain
     const excludeData = getDomainsToExludeByNetwork()
-    return excludeData.filter(e => e.domain === domain && e.entity === entity).length === 0
+    return ctpEnabled && excludeData.filter(e => e.domain === domain && e.entity === entity).length === 0
```
kdzwinel, line 41: Why do we need ctpEnabled if there is clickToLoadIsActive? Should we reuse `clickToLoadIsActive` in here instead?

shared/js/background/tracker-utils.js
```diff
@@ -167,8 +172,8 @@ function truncateReferrer (referrer, target) {
         return undefined
     }
 
-    if (tdsStorage.ReferrerExcludeList && tdsStorage.ReferrerExcludeList.excludedReferrers) {
-        const excludedDomains = tdsStorage.ReferrerExcludeList.excludedReferrers.map(e => e.domain)
+    if (tdsStorage.config.features.referrer && tdsStorage.config.features.referrer.exceptions) {
+        const excludedDomains = tdsStorage.config.features.referrer.exceptions.map(e => e.domain)
```
kdzwinel, line 176: it looks like we are doing exceptions differently here than everywhere else? we are looking for exact match instead of doing wildcard matching e.g. if "test.example.com" will be on the exception list, this exception will never match because we are doing `tldts.parse(referrer).domain` wich will return "example.com". Seems like a bug.

Should we use `isBrokenList` here instead?

shared/js/background/utils.es6.js
```diff
@@ -193,7 +193,12 @@ function isBrokenList (url, lists) {
     // If root domain in temp unprotected list, return true
     return lists.some((brokenSiteDomain) => {
         if (brokenSiteDomain) {
-            return hostname.match(new RegExp(brokenSiteDomain + '$'))
+            // TODO: Remove string check after config migration
+            if (brokenSiteDomain instanceof String) {
+                return hostname.match(new RegExp(brokenSiteDomain + '$'))
+            } else {
+                return hostname.match(new RegExp(brokenSiteDomain.domain + '$'))
```
kdzwinel, line 200: is this correct? Isn't this going to match "testexample.com" to "example.com"?

kdzwinel (commented):
Sorry for coming late, but I left a bunch of questions. Additionally:
1. Are we cleaning up old cache entries?
2. Are we sure that when updating old cache entries will not be read and cause trouble?

shared/js/background/chrome-events.es6.js
jonathanKingston: It was just the DOM signal before

shared/js/background/chrome-events.es6.js
jonathanKingston: Not globally only per domain.

shared/js/background/utils.es6.js
SlayterDev: Yes. But that raises so many questions 👀

PR #728, opened 2021-07-08
Remote Configuration Follow Up

SlayterDev:
**Reviewer:** @jonathanKingston @kdzwinel 

## Description:

Follow up changes based on comments in #723 

## Steps to test this PR:
1.  Go through all the tests in http://privacy-test-pages.glitch.me/privacy-protections/index.html
2.  All tests should appear the same as develop branch

## Automated tests:
- [x] Unit tests
- [x] Integration tests

###### Reviewer Checklist:

###### PR Author Checklist:


jonathanKingston (approved):
LGTM as this was @kdzwinel's feedback please wait for his before merging.

shared/js/background/utils.es6.js
```diff
@@ -306,6 +306,23 @@ function isSameTopLevelDomain (url1, url2) {
...
+ *
+ * @param {String} featureName - the name of the feature
+ * @param {String} customState - An optional custom state to check for
+ * @returns {bool} - if feature is enabled/matches specified state
+ */
+function isFeatureEnabled (featureName, customState) {
+    const feature = tdsStorage.config.features[featureName]
+    if (!feature) {
+        return false
+    }
```
sammacbeth, line 321: The default state (when the config for the feature is missing) is disabled? This means that if a feature name changed (e.g. due to refactoring), all old versions of the extension would disable that feature. I think a case for both defaults could be made, but maybe worth enumerating pros and cons.

shared/js/background/chrome-events.es6.js
```diff
@@ -172,7 +172,7 @@ const requestListenerTypes = utils.getUpdatedRequestListenerTypes()
 
 function blockTrackingCookies () {
     if (tdsStorage.config.features) {
-        return tdsStorage.config.features.trackingCookies.state === 'enabled'
+        return utils.isFeatureEnabled('trackingCookies')
```
kdzwinel, line 175: ❤️

shared/js/background/utils.es6.js
kdzwinel: We have a default configuration embedded into the extension, maybe we should merge it with the remote one (prioritizing the remote one)?

shared/js/background/chrome-events.es6.js
```diff
@@ -857,8 +857,6 @@ chrome.alarms.onAlarm.addListener(alarmEvent => {
             .catch(e => console.log(e))
     } else if (alarmEvent.name === 'clearExpiredHTTPSServiceCache') {
         httpsService.clearExpiredCache()
-    } else if (alarmEvent.name === 'updateUserAgentData') {
-        // TODO: Reinstate user agent rotation
```
kdzwinel, line 861: we also need to remove corresponding alarm initialization - https://github.com/duckduckgo/duckduckgo-privacy-extension/blob/46c81ba7b4fe57c520af29b9a77cea6dd22f1655/shared/js/background/chrome-events.es6.js#L835

kdzwinel (commented):
Nicely done! Looks like couple of nits are not resolved yet:

1. clean up old cache entries
1. question here - https://github.com/duckduckgo/duckduckgo-privacy-extension/pull/723#discussion_r666069559
1. bug in referrer domain matching - https://github.com/duckduckgo/duckduckgo-privacy-extension/pull/723#discussion_r666077792  (or have we moved it to another task?)
1. https://github.com/duckduckgo/duckduckgo-privacy-extension/pull/723#discussion_r666073025
1. how disabling blocking will look like in the extension? I see you've done some changes.

shared/js/background/utils.es6.js
SlayterDev: So are we saying when we get the remote config to merge it in to the default config which would:
- Overwrite existing keys
- Keep keys in the default config that don't exist in the remote config?

SlayterDev:
1. Old versions stored with their etag? I'm not sure of the purpose and I don't see us doing that with other lists.
2. Not sure why we didn't with the old file but a fallback will be good for race conditions on launch and other network issues so that we're not throwing errors.
3. I know we're tracking this in another task but I don't know where its supposed to be fixed. Will sync with @jonathanKingston 
4. I'll look in to this one.
5. Disabling `contentBlocking` will appear the same as a temp unprotected site.

shared/js/background/tracker-utils.js
```diff
@@ -172,17 +172,9 @@ function truncateReferrer (referrer, target) {
...
-            if (excludedDomains.includes(tldts.parse(referrer).domain) ||
-                excludedDomains.includes(tldts.parse(target).domain)) {
-                // referrer or target is in the Referrer safe list
-                return undefined
-            }
-        } catch (e) {
-            // if we can't parse the domains for any reason, assume it's not exluded.
-        }
+    const exceptionList = tdsStorage.config.features.referrer.exceptions
+    if (utils.isBrokenList(referrer, exceptionList) || utils.isBrokenList(target, exceptionList)) {
```
jonathanKingston, line 176: This will need a rebase as @sammacbeth has renamed it.

shared/js/content-scope/protections.js
```diff
@@ -34,7 +35,7 @@ export async function loadProtections () {
     ]
 
     for (const protectionName of protectionNames) {
-        const protection = import(`./${protectionName}-protection.js`).then(({ init, load, update }) => {
+        const protection = import(`./${protectionName}Protection.js`).then(({ init, load, update }) => {
```
kdzwinel, line 38: Might be worth catching the error here.

shared/js/background/chrome-events.es6.js
```diff
@@ -647,7 +643,7 @@ function getArgumentsObject (tabId, sender, documentUrl) {
     const site = Object.assign({}, tab?.site || {})
     const referrer = tab?.referrer || ''
 
-    const firstPartyCookiePolicy = tdsStorage.config.features?.trackingCookies.settings.firstPartyCookiePolicy || {
+    const firstPartyCookiePolicy = utils.getFeatureSettings('trackingCookies1p').firstPartyTrackerCookiePolicy || {
```
sammacbeth, line 646: `utils.getFeatureSettings('trackingCookies1p')` can return `undefined` which here would throw an exception instead of using the default value.

shared/js/background/utils.es6.js
```diff
@@ -313,6 +313,38 @@ function isSameTopLevelDomain (url1, url2) {
...
 
+/**
+ * Checks the config to see if a feature is enabled. You can optionally pass a second "customState"
+ * parameter to check if the state is equeal to other states (i.e. state === 'beta').
+ *
+ * @param {String} featureName - the name of the feature
+ * @param {String} customState - An optional custom state to check for
+ * @returns {bool} - if feature is enabled/matches specified state
+ */
+function isFeatureEnabled (featureName, customState) {
```
sammacbeth, line 324: What is `customState` used for? I don't see it being used.

shared/js/content-scope/protections.js
sammacbeth: I would _much_ prefer to have dash-separated-lowercase filenames here. Why should these files be the only camelCased files in the whole repo?

jonathanKingston:
Can we drop unit-test/data/protections.json as this has been since removed. Do we need an equiv?

shared/js/background/utils.es6.js
SlayterDev: It is future proofing. If we have a feature that uses a `beta` state for example we can check for it using `customState` instead of just checking if its equal to `enabled`.

sammacbeth (changes requested):
- There a conflict with the `develop` branch to resolve
- There is still an issue with the default value of feature settings
- Some issues with error handling.

shared/js/background/storage/tds.es6.js
```diff
@@ -28,6 +28,10 @@ class TDSStorage {
                     },
                     exceptions: []
                 },
+                trackingCookies3p: {
+                    state: 'enabled',
+                    exceptions: []
```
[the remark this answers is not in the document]
SlayterDev: Sorry I don't understand? There is now a separate trackingCookies1p that may have its own exceptions list to act on the first party.

shared/js/background/chrome-events.es6.js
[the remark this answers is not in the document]
SlayterDev: Wouldn't it if `firstPartyTrackerCookiePolicy` doesn't exist?

shared/js/content-scope/protections.js
```diff
@@ -34,13 +35,18 @@ export async function loadProtections () {
...
+            const protection = import(`./${filename}-protection.js`).then(({ init, load, update }) => {
+                if (load) {
+                    load()
+                }
+                return { protectionName, init, update }
+            })
+            protections.push(protection)
+        } catch (e) {
+            console.error(`Error loading ${protectionName}: ${e}`)
+        }
```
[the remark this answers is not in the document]
SlayterDev: @kdzwinel requested a try/catch here. If it failed it would just continue on to the next protection.

shared/js/background/chrome-events.es6.js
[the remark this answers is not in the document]
SlayterDev: We're attempting to get the settings object. Not the setting itself.

shared/js/content-scope/protections.js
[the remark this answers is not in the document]
jonathanKingston: We could split this out into another issue perhaps.

I'd prefer we failed hard at build time given that the hardcoded list of features are above and the code is all in the repo. We actually rely on the rollup dynamic import to turn this into a normal function call and bundle it all together; we could make a PR/fork that to fail hard with the desired configuration options specified to rollup.

As for catching errors that's a whole discussion:
- For loading the protections, I agree we should probably fail silently to minimise breakage in unexpected scenarios.
- Internally in the protections themselves we should be throwing as the API's do to prevent easy detection of our code.

I'm unsure if the cookie code which loads in the `load()` phase needs to throw exceptions like this; perhaps you both can answer that.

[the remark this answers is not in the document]
kdzwinel: Sorry @SlayterDev for confusion, maybe I've put my comment in a wrong place. +1 to what Sam and Jonathan are saying - we should fail hard if there are no external variables here, but when there are external variables (remote config) that can potentially have unknown/missing entries we should fail gracefully only disabling features that we don't have a config for.

shared/js/background/storage/tds.es6.js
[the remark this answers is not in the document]
sammacbeth: For all features in the config, if `example.com` is in `exceptions` it means: this feature is disabled when the site url is `example.com`.

However, `trackerCookies3p`, works differently. In this case on _any_ site, 3rd party cookies to `example.com`. I'm saying that
 1. Because the semantics of this exception differ to how `exceptions` are handled in all other cases, the name of this list in the config should be different to prevent confusion. e.g. `3pCookiesAllowed`.
 2. Perhaps we should also support the same kind of site exception for this feature too (i.e. disabling 3p cookie blocking on `example.com`) under the `exceptions` key for consistency.

sammacbeth (commented):
The 3p cookie `exceptions` are still confusing and should be named differently.

shared/js/background/chrome-events.es6.js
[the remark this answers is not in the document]
sammacbeth: Ah right - sorry I missed that.

shared/js/background/storage/tds.es6.js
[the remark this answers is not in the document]
SlayterDev: But we ARE disabling 3p cookie blocking on sites in this exceptions list. Changing the name of this list would be highly inconsistent in the config and cause maintainability headaches.

shared/js/background/utils.es6.js
jonathanKingston: Should that not be checked inside the function rather than passed in?

Aka when the state is `beta` we check that the users channel is `beta` / they've enabled some custom mode somehow.

As we're not currently using this I think we should drop it; the format supports different states which is a great improvement.

jonathanKingston (changes requested):
`brokenListIndex(url, tdsStorage.config.features[feature].aboutBlankSites || [])` and `tdsStorage.config.features[key]?.scripts || []`

These are array lists of domains that are passed into the content script; should they be arrays of objects?

sammacbeth (approved):
Thanks for making these changes!

PR #753, opened 2021-07-22
Perform tab check later and safer

SlayterDev:
**Reviewer:** @GioSensation @sammacbeth 

## Description:

`thisTab` was being checked in cases where it didn't exist. This PR moves it a bit later and performs an extra check for the tab.

## Steps to test this PR:
1. Go to https://www.wga.com/users/ron-ratto
2. Click on the Pinnacle logo
3. The page opens in a new tab
4. Open the popup → 2 trackers found

## Automated tests:
- [ ] Unit tests
- [ ] Integration tests

###### Reviewer Checklist:

###### PR Author Checklist:

### DIFF
--- a/shared/js/background/redirect.es6.js
+++ b/shared/js/background/redirect.es6.js
@@ -48,17 +48,11 @@ function handleRequest (requestData) {
 
     let thisTab = tabManager.get(requestData)
 
-    let blockingEnabled = utils.isFeatureEnabled('contentBlocking')
-
     // control access to web accessible resources
     if (requestData.url.startsWith(browserWrapper.getExtensionURL('/web_accessible_resources'))) {
         if (!thisTab || !thisTab.hasWebResourceAccess(requestData.url)) {
             return { cancel: true }
         }
-    }
-
-    if (thisTab) {
-        blockingEnabled = !utils.isFeatureBrokenForURL(thisTab.url, 'contentBlocking')
     }
 
     // For main_frame requests: create a new tab instance whenever we either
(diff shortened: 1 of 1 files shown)


shared/js/background/redirect.es6.js
```diff
@@ -48,7 +48,7 @@ function handleRequest (requestData) {
 
     let thisTab = tabManager.get(requestData)
 
-    const blockingEnabled = utils.isFeatureEnabled('contentBlocking') && !utils.isFeatureBrokenForURL(thisTab.url, 'contentBlocking')
+    let blockingEnabled = utils.isFeatureEnabled('contentBlocking')
```
sammacbeth, line 51: We don't use this variable until the `else` that starts on line 77. We can just move the whole line down there after the `thisTab` check that is done at the start of that block.

sammacbeth (changes requested):

shared/js/background/redirect.es6.js
```diff
@@ -57,6 +57,10 @@ function handleRequest (requestData) {
         }
     }
 
+    if (thisTab) {
+        blockingEnabled = !utils.isFeatureBrokenForURL(thisTab.url, 'contentBlocking')
```
jonathanKingston, line 61: This should be an `&&` as before.

shared/js/background/redirect.es6.js
```diff
@@ -48,7 +48,7 @@ function handleRequest (requestData) {
 
     let thisTab = tabManager.get(requestData)
```
jonathanKingston, line 49: Perhaps we should actually address why _blank is triggering the tab not being loaded into the tabManager as this seems a bigger concern to me.

sammacbeth: I think the error happens on the `main_frame` request, which further down should create the tab object.

sammacbeth (approved):

shared/js/background/redirect.es6.js
jonathanKingston: Yeah just seen that actually thanks :)

PR #762, opened 2021-07-28
Add stack and arguments to panel.

jonathanKingston:
Fix checking the correct feature name.
Move checks into DDG proxy to simplify methods.

**Reviewer:**

## Automated tests:
- [ ] Unit tests
- [ ] Integration tests

###### Reviewer Checklist:

###### PR Author Checklist:


shared/js/content-scope/utils.js
```diff
@@ -130,19 +133,43 @@ export function defineProperty (object, propertyName, descriptor) {
...
+        return letter.toUpperCase()
+    })
+}
+
 export class DDGProxy {
-    constructor (objectScope, property, proxyObject) {
+    constructor (featureName, objectScope, property, proxyObject) {
         this.objectScope = objectScope
         this.property = property
+        this.featureName = featureName
```
sammacbeth, line 146: `featureName`'s only use seems to be for `postDebugMessage` where it has to be camelCased. Why not just set it in camelCase originally?

sammacbeth (commented):
Looks good. Just wondering why we don't just use camelCase for `featureName` from the start?

shared/js/content-scope/utils.js
jonathanKingston: `shouldExemptMethod` works on the raw name instead which works just on the passed raw names.

jonathanKingston: We could instead camelcase on injection into the content script but we'd just move the problem I think as the inclusion of the files requires the kabab case.

sammacbeth (approved):

PR #785, opened 2021-08-08
Migrate to stable node

jonathanKingston:
**Reviewer:** @kdzwinel 
**CC:** @moollaza 

## Description:

The rationale to keeping LTS node rather than latest appears to be because of Fibers. This was added purely for a speed improvement. This only appears to be the case for async sass plugins which as far as I know we aren't using.

I've also removed grunt-contrib-compas as it appears we've never used it and it has security notices that aren't fixable.

We should audit the other Grunt plugins as I suspect there are others there that aren't needed like concat and watch.

## Steps to test this PR:

1. Remove both node_modules and build dirs
2. nvm use
3. npm install
4. npm run dev-firefox

Check that everything works the same as before.


kdzwinel (approved):
LGTM, also seems that build is faster now?

PR #788, opened 2021-08-09
Create CODE_OF_CONDUCT.md

jonathanKingston:
**Reviewer:** @kdzwinel 

## Description:

1. Because we aren't savages :)
2. This is just a standard document for addressing internet harassment 

Should we add additional / alternative emails to this document?


kdzwinel:
@jonathanKingston  let's wait with this till we wrap up a broader discussion about which COC we want (and how to put it on all our repos).

PR #797, opened 2021-08-10
Correctly calculate partyness for about:blank frames

sammacbeth:
**Reviewer:**

## Description:

For third-party cookie blocking we block the cookie getter/setter for iframes where the iframe origin is a tracker domain, which is owned by a different entity to the top-level frame domain.

In order to do this partyness check, we need the document url of the iframe, and the URL of the tab the frame is in. For the former we were using `sender.url`, which in Firefox can be `about:blank` in some circumstances. When that case happens, the check would always return that `about:blank` was third-party to whatever the tab URL was. Because the check for whether the frame document URL was a tracker used the `documentUrl` parameter sent by the content script (which would not be `about:blank`), it was possible to block cookies in a iframe from a tracker domain when it was embedded on a page of the same origin.

This PR fixes the issue by using `documentUrl` for the third party check, which should always correctly reflect the iframe document's URL.

## Automated tests:
- [ ] Unit tests
- [ ] Integration tests

###### Reviewer Checklist:

###### PR Author Checklist:

### DIFF
--- a/shared/js/background/events.es6.js
+++ b/shared/js/background/events.es6.js
@@ -637,7 +637,7 @@ function getArgumentsObject (tabId, sender, documentUrl) {
         return null
     }
     // Clone site so we don't retain any site changes
-    const site = Object.assign({}, tab?.site || {})
+    const site = Object.assign({}, tab.site || {})
     const referrer = tab?.referrer || ''
 
     const firstPartyCookiePolicy = utils.getFeatureSettings('trackingCookies1p').firstPartyTrackerCookiePolicy || {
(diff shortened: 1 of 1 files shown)


SlayterDev (approved):
LGTM! 👍

shared/js/background/events.es6.js
```diff
@@ -656,7 +656,7 @@ function getArgumentsObject (tabId, sender, documentUrl) {
         site.brokenFeatures = site.brokenFeatures.concat(utils.getBrokenFeaturesAboutBlank(tab.url))
     }
 
-    if (tab?.site.isFeatureEnabled('trackingCookies3p')) {
+    if (tab?.site.isFeatureEnabled('trackingCookies3p') || tab?.site.isFeatureEnabled('trackingCookies1p')) {
         // determine the register domain of the sending tab
         const tabUrl = tab ? tab.url : sender.tab.url
```
jonathanKingston, line 661: I'm interested when tab is blank here, do you know?

Should we change the about blank line on 656 to match?

sammacbeth: Good spot - given the check on like 633 `tab` cannot be falsy here.

I'll tidy up the optional chaining here as it seems to be used in many cases where it is not required.

jonathanKingston: Can you remove the assignment of `tabUrl` as it's not needed also? Optionally move it to 640 also but make it:
```
const tabUrl = tab.url
```

jonathanKingston (approved):

PR #799, opened 2021-08-11
Tests for 3rd party JS cookie blocking

sammacbeth:
**Reviewer:**

## Description:
Tests for the fix in #797 . The issue however was only happening in Firefox, so the test doesn't actually surface the issue when the patch is removed. However the tests do show the steps to reproduce the issue in Firefox. There does seem to be a different periodic failure related to 3rd party JS cookie blocking on chrome though: https://github.com/duckduckgo/duckduckgo-privacy-extension/runs/3302776725?check_suite_focus=true

## Automated tests:
- [ ] Unit tests
- [ ] Integration tests

###### Reviewer Checklist:

###### PR Author Checklist:

### DIFF
--- a/integration-test/background/storage.js
+++ b/integration-test/background/storage.js
@@ -122,7 +122,7 @@ describe('Storage blocking Tests', () => {
             const results = JSON.parse(await page.evaluate('JSON.stringify(results);'))
             const savedResult = results.results.find(({ id }) => id === 'memory').value
             const sameEntityiFrameResult = results.results.find(({ id }) => id === 'tracking third party iframe')?.value.find(({ test }) => test === 'JS cookie')?.result
-            console.log(savedResult, sameEntityiFrameResult)
+            // console.log(savedResult, sameEntityiFrameResult)
             expect(sameEntityiFrameResult).toBeTruthy()
             expect(sameEntityiFrameResult).not.toEqual(savedResult)
             page.close()
--- a/integration-test/background/storage.js
+++ b/integration-test/background/storage.js
@@ -44,7 +44,9 @@ describe('Storage blocking Tests', () => {
                 // Load the test pages home first to give some time for the extension background to start
                 // and register the content-script-message handler
                 await page.goto(`https://${testPageDomain}/`, { waitUntil: 'networkidle0' })
+                await page.bringToFront()
                 await page.goto(`https://${testPageDomain}/privacy-protections/storage-blocking/?store`, { waitUntil: 'networkidle0' })
+                await page.bringToFront()
                 // eslint-disable-next-line no-unmodified-loop-condition
                 while (!iframeFullyLoaded) {
                     await wait.ms(100)
@@ -116,6 +118,7 @@ describe('Storage blocking Tests', () => {
                 }
             }
             await page.goto(`https://${thirdPartyTracker}/privacy-protections/storage-blocking/?store`, { waitUntil: 'networkidle0' })
+            await page.bringToFront()
             await waitForAllResults()
             await page.click('#retrive')
             await waitForAllResults()
--- a/integration-test/background/storage.js
+++ b/integration-test/background/storage.js
@@ -1,5 +1,4 @@
 const harness = require('../helpers/harness')
-const wait = require('../helpers/wait')
 
 const testPageDomain = 'privacy-test-pages.glitch.me'
 const thirdPartyDomain = 'good.third-party.site'
@@ -24,30 +23,15 @@ describe('Storage blocking Tests', () => {
         let cookies = []
 
         beforeAll(async () => {
-            let iframeFullyLoaded = false
             const { browser, page } = await setup()
             try {
-                page.on('requestfinished', (req) => {
-                    // once we see this url, we can consider the test completed
-                    if (req.url().startsWith(`https://${thirdPartyTracker}/set-cookie`)) {
-                        iframeFullyLoaded = true
-                    }
-                })
                 // Load the test pages home first to give some time for the extension background to start
                 // and register the content-script-message handler
                 await page.goto(`https://${testPageDomain}/`, { waitUntil: 'networkidle0' })
                 await page.bringToFront()
-                await page.goto(`https://${testPageDomain}/privacy-protections/storage-blocking/?store`, { waitUntil: 'networkidle0' })
-                await page.bringToFront()
-                // eslint-disable-next-line no-unmodified-loop-condition
-                while (!iframeFullyLoaded) {
-                    await wait.ms(100)
-                }
+                await page.goto(`https://${testPageDomain}/privacy-protections/storage-blocking/?store`, { waitUntil: 'networkidle2' })
                 // collect all browser cookies
-                do {
-                    await wait.ms(1000) // allow cookies to be set
-                    cookies = (await page._client.send('Network.getAllCookies')).cookies
-                } while (cookies.length === 0)
+                cookies = (await page._client.send('Network.getAllCookies')).cookies
             } finally {
                 await page.close()
                 await harness.teardown(browser)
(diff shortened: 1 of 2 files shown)


SlayterDev (approved):

shared/js/content-scope/tracking-cookies-3p-protection.js
```diff
@@ -1,18 +1,39 @@
...
+        }
     })
 }
 
 export function init (args) {
     args.cookie.debug = args.debug
     if (window.top !== window && args.cookie.isTrackerFrame && args.cookie.shouldBlock && args.cookie.isThirdParty) {
-        // overrides expiry policy with blocking - only in subframes
-        blockCookies()
+        // overrides expiry policy with blocking - only in subframesw
```
jonathanKingston, line 36:
```suggestion
        // overrides expiry policy with blocking - only in subframes
```

integration-test/background/storage.js
```diff
@@ -19,73 +17,111 @@ describe(`On https://${testPageDomain}/privacy-protections/storage-blocking/`, (
...
+                    }
+                })
+                // Load the test pages home first to give some time for the extension background to start
+                // and register the content-script-message handler
+                await page.goto(`https://${testPageDomain}/`, { waitUntil: 'networkidle0' })
+                await page.bringToFront()
+                await page.goto(`https://${testPageDomain}/privacy-protections/storage-blocking/?store`, { waitUntil: 'networkidle0' })
+                await page.bringToFront()
+                // eslint-disable-next-line no-unmodified-loop-condition
+                while (!iframeFullyLoaded) {
```
jonathanKingston, line 43: `page.waitForRequest` or `page.waitForResponse` will likely be much less flakey and make the code simpler.

integration-test/background/storage.js
```diff
@@ -19,73 +17,111 @@ describe(`On https://${testPageDomain}/privacy-protections/storage-blocking/`, (
...
-            // Load the test pages home first to give some time for the extension background to start
-            // and register the content-script-message handler
-            await page.goto(`https://${testPageDomain}/`, { waitUntil: 'networkidle0' })
-            await page.goto(`https://${testPageDomain}/privacy-protections/storage-blocking/?store`, { waitUntil: 'networkidle0' })
-            // eslint-disable-next-line no-unmodified-loop-condition
-            while (!iframeFullyLoaded) {
-                await wait.ms(100)
+                // collect all browser cookies
+                do {
+                    await wait.ms(1000) // allow cookies to be set
```
kdzwinel, line 48: If the page itself was reporting somehow that it's "done" setting everything, would it be enough to get rid of those waits? One thing that I think I messed up with privacy test pages is not providing a clear signal from the page itself that the test is "done".

integration-test/background/storage.js
```diff
@@ -19,73 +17,111 @@ describe(`On https://${testPageDomain}/privacy-protections/storage-blocking/`, (
...
+                    cookies = (await page._client.send('Network.getAllCookies')).cookies
+                } while (cookies.length === 0)
```
kdzwinel, line 50: do we want some safeguard like do this "max 5 times"?

integration-test/background/storage.js
```diff
@@ -19,73 +17,111 @@ describe(`On https://${testPageDomain}/privacy-protections/storage-blocking/`, (
...
+            const { browser, page } = await setup()
+            async function waitForAllResults () {
+                while ((await page.$$('#tests-details > li > span > ul')).length < 2) {
+                    await new Promise(resolve => setTimeout(resolve, 100))
+                }
+            }
+            await page.goto(`https://${thirdPartyTracker}/privacy-protections/storage-blocking/?store`, { waitUntil: 'networkidle0' })
+            await page.bringToFront()
+            await waitForAllResults()
+            await page.click('#retrive')
```
kdzwinel, line 116: BTW we also have `https://privacy-test-pages.glitch.me/privacy-protections/storage-blocking/?retrive`, but maybe clicking on a button is faster here.

integration-test/background/storage.js
```diff
@@ -19,73 +17,111 @@ describe(`On https://${testPageDomain}/privacy-protections/storage-blocking/`, (
...
-        expect(jsCookie.expires).toBeLessThan(Date.now() / 1000 + 864000) // 10 days in the future
+    describe(`On https://${thirdPartyTracker}/privacy-protections/storage-blocking/`, () => {
+        it('does not block iFrame tracker cookies from same entity', async () => {
+            const { browser, page } = await setup()
+            async function waitForAllResults () {
+                while ((await page.$$('#tests-details > li > span > ul')).length < 2) {
+                    await new Promise(resolve => setTimeout(resolve, 100))
+                }
+            }
+            await page.goto(`https://${thirdPartyTracker}/privacy-protections/storage-blocking/?store`, { waitUntil: 'networkidle0' })
```
kdzwinel, line 113: Why aren't we using testPageDomain here? Isn't this changing outcome of test that we are using `broken.third-party.site` here?

shared/js/content-scope/tracking-cookies-3p-protection.js
```diff
@@ -1,18 +1,39 @@
...
-        get: () => ''
+        set: function (value) {
+            if (debug) {
+                postDebugMessage('jscookie', {
+                    action: 'block',
+                    reason: 'tracker frame',
+                    documentUrl: document.location.href,
+                    scriptOrigins: [],
+                    value: value
+                })
```
kdzwinel, line 15: this is just debug for manual debugging? Or are we depending on those messages in tests somehow?

kdzwinel (commented):
Thanks @sammacbeth for putting it together, I left a couple of questions.

integration-test/background/storage.js
```diff
@@ -19,73 +17,111 @@ describe(`On https://${testPageDomain}/privacy-protections/storage-blocking/`, (
...
+                page.on('requestfinished', (req) => {
+                    // once we see this url, we can consider the test completed
+                    if (req.url().startsWith(`https://${thirdPartyTracker}/set-cookie`)) {
+                        iframeFullyLoaded = true
+                    }
+                })
+                // Load the test pages home first to give some time for the extension background to start
+                // and register the content-script-message handler
+                await page.goto(`https://${testPageDomain}/`, { waitUntil: 'networkidle0' })
+                await page.bringToFront()
```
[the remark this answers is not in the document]
sammacbeth: For me this resolved issues when running the test locally, and was also convenient to see the test progress rather than just staring at the postinstall page until the tests failed :)

integration-test/background/storage.js
sammacbeth: In this case I'm not 100% sure - in this case we're waiting until we can read the cookies set from the browser's cookie DB, and I think there would still be a chance of a race if we called `Network.getAllCookies` in the same tick as the page set the cookies - I'm not sure how quickly the values get flushed to the DB.

integration-test/background/storage.js
sammacbeth: The test timeout is ~20s anyway, so we'd hit that cap not long after 5 times.

integration-test/background/storage.js
sammacbeth: That's the purpose of this test - we're loading a tracker page and checking that the cookies for an iframe on the same origin (i.e. entity) are not blocked. We're testing that blocking works differently to the other block of tests where the iframe is 3rd party.

shared/js/content-scope/tracking-cookies-3p-protection.js
sammacbeth: These are for the debugger panel, yes.

integration-test/background/storage.js
sammacbeth: It seems to now work using only `waitUntil: 'networkidle2'`.

kdzwinel (approved):

PR #827, opened 2021-09-01
Update email protection in settings

GioSensation:
**Asana:** https://app.asana.com/0/312629933896096/1200412250706923/f

## Description:
Changes Email Protection text in the settings and adds a visible Autofill Disabled state (in previous version, when autofill was disabled we just didn't show it).

## Steps to test this PR:
Go to settings and check that the Email Protection section is visible both when logged in and logged out.


shared/js/ui/models/user-data.es6.js
```diff
@@ -13,7 +13,7 @@ UserData.prototype = window.$.extend({},
 
         logout () {
             this.fetch({ logout: true })
-                .then(() => this.set('loggingOut', true))
+                .then(() => this.set('userName', null))
```
GioSensation, line 16: In the previous iteration this had a transient state (`loggingOut`), but now we either have the `userName` or we don't.

jonathanKingston (approved):

PR #837, opened 2021-09-09
Improve breakage reporting UX

SlayterDev:
**Reviewer:**

## Automated tests:
- [ ] Unit tests
- [ ] Integration tests

###### Reviewer Checklist:

###### PR Author Checklist:


shared/js/ui/base/ui-wrapper.es6.js
```diff
@@ -58,6 +58,23 @@ const closePopup = () => {
...
 
+const startBreakageFlow = (tabId, flowStyle, cb) => {
+    window.chrome.tabs.reload(tabId, { bypassCache: true }, () => {
+        setTimeout(() => {
+            window.chrome.tabs.executeScript(tabId, {
+                code: `window.breakageModel = '${flowStyle}'`
+            }, () => {
+                console.log('execute 2')
+                window.chrome.tabs.executeScript(tabId, {
+                    file: 'public/js/content-scripts/breakage-flow.js'
```
jonathanKingston, line 69: Rather than injecting in a different script and also needing a window global, can you make this a `content-scope` script and use the `update` method as cookies use to send the `flowStyle`?

This will have the following advantages:
- Using remote config is built in by default
- There isn't an exposed global on the page that the page could potentially attack
- One unified script, dynamic script execution is potentially going soon with manifest v3

Let me know if you need some pointers on this code as I know it's a little more complex.

shared/js/content-scripts/breakage-flow.js
```diff
@@ -0,0 +1,403 @@
...
+
+        modal.appendChild(modalContent)
+        modalContainer.appendChild(modal)
+
+        return modalContainer
+    }
+
+    function confirmSiteFixed (modalContainer, modalContent) {
+        // Remove modal content
+        while (modalContent.childNodes.length > 1) {
```
jonathanKingston, line 316: Can you make this a method? Alternatively just rebuild the modal DOM via a function? It would probably better to delete everything when it's not required.

shared/js/background/events.es6.js
```diff
@@ -499,6 +499,44 @@ chrome.runtime.onMessage.addListener((req, sender, res) => {
...
+        chrome.tabs.create({ url: url })
+
+        return true
+    } else if (req.openBreakageReport) {
+        chrome.tabs.create({ url: req.url })
+
+        return true
+    } else if (req.breakageFlow) {
+        chrome.tabs.reload(req.tabId, { bypassCache: true }, () => {
+            setTimeout(() => {
```
jonathanKingston, line 527: We should probably wait for the page navigation to complete right?

We should store the modal event in a global/storage and trigger it on _chrome.webNavigation.onDOMContentLoaded_ or similar.

So perhaps:
```
    } else if (req.breakageFlow) {
        browserWrapper.syncToStorage({ breakageModalLoad: req.tabId })
        chrome.tabs.reload(req.tabId, { bypassCache: true })
```

And then:
```
chrome.webNavigation.onDOMContentLoaded.addListener((load) => {
    browserWrapper.getFromStorage('breakageModalLoad', (modalId) => {
      if (moadId && load.tabId === moadId) {
        // Inject modal
      }
    })
}, {...})
```

shared/js/background/events.es6.js
```diff
@@ -499,6 +499,44 @@ chrome.runtime.onMessage.addListener((req, sender, res) => {
...
+                chrome.tabs.executeScript(req.tabId, {
+                    code: `window.breakageModel = '${req.breakageFlow}'`
+                }, () => {
+                    console.log('execute 2')
+                    chrome.tabs.executeScript(req.tabId, {
+                        file: 'public/js/content-scripts/breakage-flow.js'
```
jonathanKingston, line 533: If we aren't using the content-scope bundling code, I *think* we can use params here instead:
```suggestion
                        file: `public/js/content-scripts/breakage-flow.js?breakageModel=${encodeURIComponent(req.breakageFlow)}`
```

Then in the script check: `document.currentScript` params.

sammacbeth:
Closing as stale.

PR #843, opened 2021-09-14
Async shim for background pages

jonathanKingston:
**Reviewer:**

## Automated tests:
- [ ] Unit tests
- [ ] Integration tests

###### Reviewer Checklist:

###### PR Author Checklist:


shared/js/content-scripts/click-to-load.js
```diff
@@ -1,4 +1,7 @@
 (function clickToLoad () {
+    function sendMessage (messageType, options, callback) {
+        chrome.runtime.sendMessage({ messageType, options }, callback)
```
kdzwinel, line 3: no browser / promises in here? I guess we don't want to inject the polyfil as a content script, right?

shared/js/content-scripts/fb-surrogate-xray.js
```diff
@@ -16,7 +16,8 @@
     let sdkPatterns = [/connect.facebook.net\/[a-zA-Z_]+\/(sdk|all).js/]
     // Request list of rules to capture from the extension
     chrome.runtime.sendMessage({
-        getSocialSurrogateRules: entity
+        messageType: 'getSocialSurrogateRules',
+        options: entity
```
kdzwinel, line 20: why change of a format?

shared/js/background/devtools.es6.js
```diff
@@ -24,7 +24,7 @@ function connected (port) {
             const matchedTracker = trackers.getTrackerData(requestData.url, siteUrl, requestData)
             if (tracker.matchedRule) {
                 // find the rule for this url
-                const ruleIndex = matchedTracker.tracker.rules.findIndex((r) => r.rule.toString() === tracker.matchedRule)
+                const ruleIndex = matchedTracker.tracker.rules.findIndex((r) => r.rule?.toString() === tracker.matchedRule)
```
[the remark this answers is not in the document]
jonathanKingston: Yeah we can land this separate, there's another PR for this and I can rebase this away.

shared/js/content-scripts/click-to-load.js
jonathanKingston: Not yet at least, I'm just changing the `messageType` shape so it's easier to process in the onMessage handler.

shared/js/content-scripts/fb-surrogate-xray.js
jonathanKingston: Same as above, it eases the handling in the `onMessage` handler without it we would have to check *all* the properties of the message and see if it aligns with an `export` name which seemed very fragile.

I moved to the `export` because having an `async` handler when sometimes there isn't a response meant it might never complete which is obviously non ideal. The change wasn't required but the alternative made the same kinda size of change in the `onMessage` handler at least.

For example some messages look like: `{setListContents: true, value: {}}`

If we added a `export function value() {}` we'd end up calling the other method.

package.json
```diff
@@ -99,7 +100,11 @@
             [
                 "@babel/preset-env",
                 {
-                    "debug": true
+                    "debug": true,
+                    "targets": {
```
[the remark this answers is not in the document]
jonathanKingston: It prevents the need for the regenerator runtime for any file that needs `await` this is very bloated for platforms that we don't support.

shared/js/background/redirect.es6.js
```diff
@@ -58,7 +58,7 @@ function handleRequest (requestData) {
 
     // For main_frame requests: create a new tab instance whenever we either
     // don't have a tab instance for this tabId or this is a new requestId.
-    if (requestData.type === 'main_frame' && window.chrome) {
```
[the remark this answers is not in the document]
jonathanKingston: This I believe is called also through a unit test and in that case it's not available. I'm not really sure why we took this approach over loading in a shim of the APIs but I didn't want to change even more things :P

PR #852, opened 2021-09-24
Move to contextMenu.onClicked as onclick isn't accessible in an event…

jonathanKingston:
… page

**Reviewer:** @GioSensation 

## Description:

`onclick` doesn't work inside an event page / service worker and the onClicked event listener is identical from what I can tell.

## Automated tests:
- [ ] Unit tests
- [ ] Integration tests

###### Reviewer Checklist:

###### PR Author Checklist:

### DIFF
--- a/shared/js/background/email-utils.es6.js
+++ b/shared/js/background/email-utils.es6.js
@@ -50,15 +50,15 @@ const createAutofillContextMenuItem = () => {
         id: MENU_ITEM_ID,
         title: 'Use Duck Address',
         contexts: ['editable'],
-        visible: false,
-        onclick: (info, tab) => {
-            const userData = getSetting('userData')
-            if (userData.nextAlias) {
-                browser.tabs.sendMessage(tab.id, {
-                    type: 'contextualAutofill',
-                    alias: userData.nextAlias
-                })
-            }
+        visible: false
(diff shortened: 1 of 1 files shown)


shared/js/background/email-utils.es6.js
```diff
@@ -6,7 +6,7 @@ let attempts = 1
 
 const fetchAlias = () => {
     // if another fetch was previously scheduled, clear that and execute now
-    browser.alarms.get(REFETCH_ALIAS_ALARM, () => browser.alarms.clear(REFETCH_ALIAS_ALARM))
+    browser.alarms.clear(REFETCH_ALIAS_ALARM)
```
GioSensation, line 9: Had to change this, otherwise the new code threw. Looks like the polyfill doesn't support callbacks. Anyway, it turns out you don't need to call `.get` before you call `.clear`. The call just returns `false` if the alarm was not previously set.

GioSensation (approved):
LGTM.

shared/js/background/email-utils.es6.js
jonathanKingston: Ah yeah sorry this was done elsewhere, it should have used `await` instead of the callback.

PR #853, opened 2021-09-24
Migrate from using localStorage

jonathanKingston:
**Reviewer:**

## Description:

Removes the dependency on window.localStorage which isn't accessible in a worker scope.

## Automated tests:
- [ ] Unit tests
- [ ] Integration tests

###### Reviewer Checklist:

###### PR Author Checklist:

### DIFF
--- a/shared/js/background/events.es6.js
+++ b/shared/js/background/events.es6.js
@@ -287,7 +287,7 @@ chrome.omnibox.onInputEntered.addListener(function (text) {
         active: true
     }, function (tabs) {
         chrome.tabs.update(tabs[0].id, {
-            url: 'https://duckduckgo.com/?q=' + encodeURIComponent(text) + '&bext=' + localStorage.os + 'cl'
+            url: 'https://duckduckgo.com/?q=' + encodeURIComponent(text) + '&bext=' + utils.getOsName() + 'cl'
         })
     })
 })
--- a/shared/js/background/background.es6.js
+++ b/shared/js/background/background.es6.js
@@ -23,12 +23,5 @@ settings.ready().then(() => {
     // clearing last search on browser startup
     settings.updateSetting('last_search', '')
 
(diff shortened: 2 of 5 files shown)


shared/js/background/events.es6.js
```diff
@@ -287,7 +287,7 @@ chrome.omnibox.onInputEntered.addListener(function (text) {
         active: true
     }, function (tabs) {
         chrome.tabs.update(tabs[0].id, {
-            url: 'https://duckduckgo.com/?q=' + encodeURIComponent(text) + '&bext=' + localStorage.os + 'cl'
+            url: 'https://duckduckgo.com/?q=' + encodeURIComponent(text) + '&bext=' + utils.getOsName() + 'cl'
         })
     })
```
kdzwinel, line 292: This whole feature is not working. I'm not sure why we have it in the first place. To test this you have to type "ddg" in the omnibox, click TAB, and then search for something. It won't work and it didn't work before this PR. Let's remove the whole thing (it's this one listener and note in manifest.json)?

kdzwinel: Hm, wait, it's working on production, something weird is going on because that `tabs[0].id` definitely shouldn't work.

kdzwinel: Yeah, it thows errors but it works somehow. `update` defaults to current tab and this first param is optional, so lets remove it.

jonathanKingston: This works for me (in Chrome). Perhaps it's chrome only?

kdzwinel: Yeah, my bad. Bug only exists when debugging

kdzwinel (approved):
Can't tell why we were using localStorage there, thanks for removing it! Tested both places request is created, bext param is correct.

PR #863, opened 2021-09-27
Sync session key via session storage

jonathanKingston:
**Reviewer:** @kzar 

## Description:

https://app.asana.com/0/312629933896096/1201066533025796/f

Changes to synced storage for the session key sent to the content script (used for psudo-randomisation)

## Steps to test this PR:

1.  Open [several canvas draw pages](https://good.third-party.site/features/canvas-draw.html)
2. Click 'draw pattern'
3. Ensure the output is the same across the same page.
4. Open a different first party domain: https://bad.third-party.site/features/canvas-draw.html
5. Click 'draw pattern'
6. Ensure the output is different to the previous hashes.

Open the extension debugger and run _dbg.utils.resetSessionKey()_ and verify the tabs have different hashes; this simulates session key expiry via the alarm.

## Automated tests:
- [ ] Unit tests
- [ ] Integration tests

###### Reviewer Checklist:

###### PR Author Checklist:

### DIFF
--- a/shared/js/background/debug.es6.js
+++ b/shared/js/background/debug.es6.js
@@ -9,6 +9,7 @@ const https = require('./https.es6')
 const tds = require('./storage/tds.es6')
 const startup = require('./startup.es6')
 const browserWrapper = require('./wrapper.es6')
+const utils = require('./utils.es6')
 
 self.dbg = {
     settings,
@@ -16,7 +17,8 @@ self.dbg = {
     tabManager,
     atb,
     https,
-    tds
+    tds,
+    utils
 }
 
 // mark this as a dev build
--- a/shared/js/background/events.es6.js
+++ b/shared/js/background/events.es6.js
@@ -19,24 +19,9 @@ const tdsStorage = require('./storage/tds.es6')
 const browserWrapper = require('./wrapper.es6')
 const limitReferrerData = require('./events/referrer-trimming')
 const { dropTracking3pCookiesFromResponse, dropTracking3pCookiesFromRequest } = require('./events/3p-tracking-cookie-blocking')
-const getArgumentsObject = require('./helpers/arguments-object')
 const startup = require('./startup.es6')
 
-const sha1 = require('../shared-utils/sha1')
-
 const manifestVersion = browserWrapper.getManifestVersion()
-
-/**
- * Produce a random float, same output as Math.random()
- * @returns {number}
- */
-function getFloat () {
-    return crypto.getRandomValues(new Uint32Array(1))[0] / 2 ** 32
-}
-
-function getHash () {
-    return sha1(getFloat().toString())
-}
 
 async function onInstalled (details) {
     tdsStorage.initOnInstall()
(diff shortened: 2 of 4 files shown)


shared/js/background/events.es6.js
```diff
@@ -352,8 +366,13 @@ browser.runtime.onMessage.addListener((req, sender) => {
 /**
  * Fingerprint Protection
  */
-// TODO fix for manifest v3
-let sessionKey = getHash()
+let sessionKey
```
kzar, line 369: How about instead we make the `browser.runtime.onMessage` listener `async` and do `await getSessionKey()` where we need the session key? Then we could avoid having the local variable + init function.

I don't think this code is susceptible to race conditions, but doing it that way would also be setting a better example of how to manage state with MV3 (e.g. using session storage and waiting for it to be ready, rather than using local variables). That might help others avoid foot-guns in the future?

shared/js/background/events.es6.js
```diff
@@ -33,7 +33,21 @@ function getFloat () {
     return crypto.getRandomValues(new Uint32Array(1))[0] / 2 ** 32
 }
 
-function getHash () {
+async function getHash () {
+    let hash = await browserWrapper.getFromSessionStorage('sessionKey')
+    if (!hash) {
+        hash = setNewHash()
```
kzar, line 39: I'd prefer to make `setNewHash` `async`, have it `await` on `browserWrapper.setToSessionStorage` and then `await` on the `setNewHash` call here. I don't think a race condition (where we call `getHash` again before finished writing the hash to storage) is likely though to be fair.

kzar (changes requested):

shared/js/background/events.es6.js
```diff
@@ -33,7 +33,21 @@ function getFloat () {
...
+    }
+    return hash
+}
+
+function setNewHash () {
+    const hash = getNewHash()
```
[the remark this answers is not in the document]
jonathanKingston: I was hoping this could be reused if we need it again but yeah ok.

shared/js/background/events.es6.js
```diff
@@ -33,7 +33,21 @@ function getFloat () {
     return crypto.getRandomValues(new Uint32Array(1))[0] / 2 ** 32
 }
 
-function getHash () {
+async function getHash () {
```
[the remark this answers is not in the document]
jonathanKingston: Yeah wfm

shared/js/background/events.es6.js
jonathanKingston: Making the onMessage handler be async is a very large task, we should avoid that here (mostly why I've avoided this feedback as didn't have chance to fix it).

I've moved the one handler into the message-handlers file instead and made this async; I think this matches the spirit of your feedback and removes the local var.

shared/js/background/debug.es6.js
```diff
@@ -9,14 +9,16 @@ const https = require('./https.es6')
...
 
 self.dbg = {
     settings,
     startup,
     tabManager,
     atb,
     https,
-    tds
+    tds,
+    utils
```
jonathanKingston, line 21: This allows us to run _dbg.utils.resetSessionKey()_ in devtools windows to verify that the session key is reset.

shared/js/background/utils.es6.js
```diff
@@ -1,12 +1,35 @@
...
 import tdsStorage from './storage/tds.es6'
 import settings from './settings.es6'
 import load from './load.es6'
 import * as tldts from 'tldts'
 import parseUserAgentString from '../shared-utils/parse-user-agent-string.es6'
+import sha1 from '../shared-utils/sha1'
 const browserInfo = parseUserAgentString()
 
+/**
+ * Produce a random float, same output as Math.random()
```
kzar, line 12: I think this comment should explain the purpose or appropriate use of this function a bit more. The reader might ask "If it's the same output as `Math.random()`, then why not just use `Math.random()`?"

shared/js/background/events.es6.js
```diff
@@ -539,8 +490,7 @@ browser.alarms.onAlarm.addListener(async alarmEvent => {
     } else if (alarmEvent.name === 'clearExpiredHTTPSServiceCache') {
         httpsService.clearExpiredCache()
     } else if (alarmEvent.name === 'rotateSessionKey') {
-        // TODO fix for manifest v3
-        sessionKey = getHash()
+        utils.resetSessionKey()
```
kzar, line 493: I think we should add an `await` to this call.

jonathanKingston: That doesn't give us the storage guarantees you're expecting as I mentioned in the meeting. I can add it, but it might confuse another contributor expecting they can rely on the API so long as it's always awaited and I don't think this is the case especially with the event listeners.

I think we should address the reliability of the session writing elsewhere and for this the chance of a race condition is if a session can't be written every 24 hours we might risk overwriting it... If we can't rely on that; we have much bigger issues.

kzar: I think we should add the `await`, but also maybe add a comment to the start of the (already `async`) function to explain that. We already use `await` elsewhere in the function, so I figure we should use `await` consistently, but also take care to explain the potential downfalls to folks.

kzar (approved):
LGTM

PR #869, opened 2021-09-28
Resolve reinstall caching previous bundled config values

jonathanKingston:
**Reviewer:**

## Description:

Manual testing is the only way I've found to resolve this issue mentioned in: https://app.asana.com/0/1200223097357040/1201035302493499/f

## Steps to test this PR:

For this ensure you're on an old copy of the extension copy in privacy-configuration module (you can checkout an old commit and then run `npm run generate-etags`)

1. Load the extension with web-ext
2. Inspect the storage to see if there's what is stored in the indexedDB
3. Check network requests to see if a 304 occurred
4. Update the code (web ext should trigger an update)
5. Check the storage again to verify it's still the remote copy

## Automated tests:
- [ ] Unit tests
- [ ] Integration tests

###### Reviewer Checklist:

###### PR Author Checklist:

### DIFF
--- a/shared/js/background/storage/tds.es6.js
+++ b/shared/js/background/storage/tds.es6.js
@@ -16,14 +16,15 @@ class TDSStorage {
         this.surrogates = ''
         this.ClickToLoadConfig = {}
         this.config = {}
-        // If initOnInstall is not called before getList, just resolve.
-        this._installing = Promise.resolve(false)
+        this.isInstalling = false
 
         this.removeLegacyLists()
     }
 
-    initOnInstall () {
-        this._installing = this._internalInitOnInstall()
+    async initOnInstall () {
+        this.isInstalling = true
+        this._installingPromise = await this._internalInitOnInstall()
+        this.isInstalling = false
     }
 
     async _internalInitOnInstall () {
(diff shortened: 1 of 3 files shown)


shared/js/background/storage/tds.es6.js
```diff
@@ -14,17 +15,36 @@ class TDSStorage {
...
+    }
+
     getLists () {
         return Promise.all(constants.tdsLists.map(list => this.getList(list)))
     }
 
-    getList (list) {
+    async getList (list) {
+        // If initOnInstall was called, await the updating from the local bundles before fetching
+        await this._installing
```
GioSensation, line 47: I find this pattern rather confusing. We're setting `this._installing` to a promise resolving to false, then to a function call that doesn't return anything (`undefined`), and here we are not calling anything. Isn't there a clearer pattern we can use? Like calling a function that actually changes `_installing` to something meaningful, like `true`/`false` once the promise resolves?

GioSensation (approved):
This looks good. I've left a comment for something that I find confusing.

shared/js/background/storage/tds.es6.js
jonathanKingston: It's a weird pattern I agree, we do use it elsewhere for settings for example (it also sets `isReady` which is handy when we can choose to ignore if that it's ready, however in this case we must block to prevent the race).

PR #874, opened 2021-09-29
Fetch remote config instead of using the git module

jonathanKingston:
**Reviewer:**

## Description:

Moves from the bundled config to loading it remotely. We can easily expand this now to all the tds configs too.

## Automated tests:
- [ ] Unit tests
- [ ] Integration tests

###### Reviewer Checklist:

###### PR Author Checklist:


scripts/bundleConfig.js
```diff
@@ -0,0 +1,27 @@
+const fs = require('fs')
+const config = require('../shared/data/constants.js')
+const fetch = (...args) => import('node-fetch').then(({ default: fetch }) => fetch(...args))
+
+const outputEtagFile = './shared/data/etags.json'
+const bundlePath = './shared/data/bundled/'
+const configProps = config.tdsLists.find((c) => c.name == 'config');
```
GioSensation, line 7: Should this be `===`? Also removed the `;`.

```suggestion
const configProps = config.tdsLists.find((c) => c.name === 'config')
```

scripts/bundleConfig.js
```diff
@@ -0,0 +1,27 @@
...
+    etags[`${name}-etag`] = response.headers.get('etag')
+    const fileText = await response.text()
+    const urlObject = new URL(url)
+    const fileName = urlObject.pathname.split('/').pop()
+    fs.writeFileSync(`${bundlePath}/${fileName}`, fileText)
+}
+
+async function fetchConfigs() {
+    await getConfig('config', configProps.url)
+    fs.writeFileSync(outputEtagFile, JSON.stringify(etags))
```
GioSensation, line 24: What about logging a success message here?

```suggestion
    fs.writeFileSync(outputEtagFile, JSON.stringify(etags))
    console.log('Config imported successfully')
```

scripts/bundleConfig.js
```diff
@@ -0,0 +1,27 @@
+const fs = require('fs')
+const config = require('../shared/data/constants.js')
+const fetch = (...args) => import('node-fetch').then(({ default: fetch }) => fetch(...args))
+
+const outputEtagFile = './shared/data/etags.json'
+const bundlePath = './shared/data/bundled/'
+const configProps = config.tdsLists.find((c) => c.name == 'config');
+const etags = {};
```
GioSensation, line 8:
```suggestion
const etags = {}
```

scripts/bundleConfig.js
```diff
@@ -0,0 +1,27 @@
...
+
+async function getConfig(name, url) {
+    let response = await fetch(url)
```
GioSensation, line 11: My IDE is suggesting all these linting improvements, sorry for being nitty 🤓.

```suggestion
async function getConfig (name, url) {
    const response = await fetch(url)
```

scripts/bundleConfig.js
```diff
@@ -0,0 +1,27 @@
...
+        throw new Error(`Failed to load tds list ${name}: ${url}`)
+    }
+    etags[`${name}-etag`] = response.headers.get('etag')
+    const fileText = await response.text()
+    const urlObject = new URL(url)
+    const fileName = urlObject.pathname.split('/').pop()
+    fs.writeFileSync(`${bundlePath}/${fileName}`, fileText)
+}
+
+async function fetchConfigs() {
```
GioSensation, line 22:
```suggestion
async function fetchConfigs () {
```

GioSensation (approved):
Thanks for this, left a few minor nits.

scripts/bundleConfig.js
kdzwinel: Good catch Emanuele - something is clearly wrong with linter setup though. Can you double check if that file is linted @jonathanKingston ?

scripts/bundleConfig.js
```diff
@@ -0,0 +1,27 @@
...
+async function getConfig(name, url) {
+    let response = await fetch(url)
+    if (!response.ok) {
+        throw new Error(`Failed to load tds list ${name}: ${url}`)
+    }
+    etags[`${name}-etag`] = response.headers.get('etag')
+    const fileText = await response.text()
+    const urlObject = new URL(url)
+    const fileName = urlObject.pathname.split('/').pop()
+    fs.writeFileSync(`${bundlePath}/${fileName}`, fileText)
```
kdzwinel, line 19: can we do some sanity check before writting? Like verify if this is valid JSON and throw if not? Throwing will fail the build process right?

scripts/bundleConfig.js
```diff
@@ -0,0 +1,27 @@
+const fs = require('fs')
+const config = require('../shared/data/constants.js')
+const fetch = (...args) => import('node-fetch').then(({ default: fetch }) => fetch(...args))
```
kdzwinel, line 3: OMG what happened here? Why are we importing node-fetch only after fetch is called? Can we make it a bit less interesting?

scripts/bundleConfig.js
jonathanKingston: scripts/ are explicitly not linted because most of them are very broken lint-wise. I'll add this for now to the lint step and fix it up. I don't want to muddy this PR with lots of syntax changes... I'll make that a follow up.

scripts/bundleConfig.js
jonathanKingston: This is the standard way to import it now when not in a mjs file: https://www.npmjs.com/package/node-fetch I changed to .mjs which still equally feels weird to me.

scripts/bundleConfig.js
jonathanKingston: Yeah it will fail it, for now I've just added JSON parse which is a good enough sanity check. We could additionally run the tests on the file but that's a much bigger change.

PR #899, opened 2021-10-15
Improve release automation with Asana magic

GioSensation:
**Reviewers:** @jonathanKingston @MariagraziaAlastra @sammacbeth 
**Asana:** https://app.asana.com/0/0/1200750505563921/1201183123265449/f

## Description:
Automate the Asana paperwork!

- Run the workflow, with manual input of version (as the current stage.yml)
- Run the script and commit the changes
- Generate a new Release draft (prerelease: true)
- Use the list of commits from the previous tag to the current to populate the release notes
- Build zip files and upload them to the release ← We already have most of this logic in publish.yml
- Asana Script:
  - Get list of tasks in the Extension release project under the Release section
  - Duplicate the release template, paste in the list of tasks in the appropriate section
  - Upload the zip files to Asana

## Steps to test this PR:
Give it a first pass, then we'll merge and try to cut test releases to see how it works. You can see what the Asana script spits out in https://app.asana.com/0/312629933896096/1201202709634481/f.

### DIFF
--- a/.github/workflows/stage.yml
+++ b/.github/workflows/stage.yml
@@ -20,6 +20,15 @@ jobs:
         uses: actions/setup-node@v1
         with:
           node-version: 16.x
+
+      - name: Restore npm artifact from cache
+        uses: actions/cache@v2
+        with:
+          path: ~/.npm
+          key: ${{ runner.os }}-node-${{ hashFiles('**/package-lock.json') }}
+          restore-keys: |
+            ${{ runner.os }}-node-
+
       - name: Install dependencies
         run: |
           npm i -g web-ext
@@ -41,12 +50,6 @@ jobs:
         with:
           myToken: ${{ secrets.GITHUB_TOKEN }}
 
-      - uses: actions/cache@v2
-        with:
-          path: ~/.npm
-          key: ${{ runner.os }}-node-${{ hashFiles('**/package-lock.json') }}
-          restore-keys: |
-            ${{ runner.os }}-node-
         - name: Build
           run: |
             npm run bundle-config
--- a/.github/workflows/stage.yml
+++ b/.github/workflows/stage.yml
@@ -10,25 +10,75 @@ on:
 jobs:
   release_pr:
     runs-on: ubuntu-20.04
+    strategy:
+      matrix:
+        target: [ release-chrome, release-firefox ]
 
     steps:
-    - uses: actions/checkout@v2
-    - name: Use Node.js 16
-      uses: actions/setup-node@v1
-      with:
-        node-version: 16.x
-    - name: Install dependencies
-      run: |
-        npm i -g web-ext
-        npm run install-ci
-    - name: Fetch config and update version
-      run: |
-        npm run bundle-config
-        node scripts/bumpVersion.js ${{ github.event.inputs.version }}
-    - name: Create Pull Request
-      uses: peter-evans/create-pull-request@v3
-      with:
-        commit-message: ${{ github.event.inputs.version }}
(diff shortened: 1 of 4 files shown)


.github/workflows/stage.yml
```diff
@@ -10,25 +10,75 @@ on:
...
+      - uses: actions/cache@v2
+        with:
+          path: ~/.npm
+          key: ${{ runner.os }}-node-${{ hashFiles('**/package-lock.json') }}
+          restore-keys: |
+            ${{ runner.os }}-node-
+        - name: Install dependencies
+          run: |
+            npm ci
+            npm i -g web-ext
```
sammacbeth, line 53: It looks like we're running `Install dependencies` twice in the same pipeline. Is that on purpose?

.github/workflows/stage.yml
```diff
@@ -10,25 +10,75 @@ on:
...
+              ./build/firefox/release/web-ext-artifacts/*.zip
+              ./build/chrome/release/*.zip
+
+        - name: Asana Workflow
+            env:
+              ASANA_ACCESS_TOKEN: ${{ secrets.ASANA_ACCESS_TOKEN }}
+              VERSION: ${{ github.event.release.tag_name }}
+              RELEASE_URL: ${{ github.event.release.html_url }}
+            run: |
+              npm install
```
sammacbeth, line 83: Way do we need to run `npm install` again?

sammacbeth (commented):
Looks good. It seems we're running npm install too often, but otherwise this is great! Also needs a rebase, but then feel free to merge.

.github/workflows/stage.yml
GioSensation: Just oversight. I'm removing the extra installs. While checking, I noticed that we also have an `install-ci` command in package.json (see [here](https://github.com/duckduckgo/duckduckgo-privacy-extension/blob/2e0473ba17a9314c63fdfd48aac728b26da9d57a/package.json#L10)). Should we use that instead? It rebuilds puppeteer (which we don't need) but I'm not sure about the rest. cc @jonathanKingston

sammacbeth: Yes I think we should use `install-ci` as this prevents a potential exploit of the build via npm install scripts.

.github/workflows/stage.yml
```diff
@@ -10,25 +10,70 @@ on:
...
+        uses: metcalfc/changelog-generator@c3f61b2a3db543ceee83c759ac8e48419a1a2211
+        with:
+          myToken: ${{ secrets.GITHUB_TOKEN }}
+
+      - uses: actions/cache@v2
+        with:
+          path: ~/.npm
+          key: ${{ runner.os }}-node-${{ hashFiles('**/package-lock.json') }}
+          restore-keys: |
+            ${{ runner.os }}-node-
```
sammacbeth, line 49: This should be moved to before the `install-ci` step (as in https://github.com/duckduckgo/duckduckgo-privacy-extension/blob/develop/.github/workflows/tests.yml#L14-L20). This is step restores a cached `.npm` folder for so the `npm install` artifacts don't need to be fetched.

See https://github.com/actions/cache/blob/main/examples.md#node---npm for more on how this cache step works.

sammacbeth (changes requested):

.github/workflows/stage.yml
GioSensation: Thanks for noticing. I've pushed the change. Do you mind checking again if I got it right this time 🙂?

sammacbeth (approved):

PR #947, opened 2021-11-24
GPC reference tests

kdzwinel:
**Reviewer:** @dharb 

## Description:
This was supposed to be only GPC reference tests integration, but I also did a bit of a cleanup:
- I moved reference-tests submodule from `/unit-test/background` because this is the main folder with tests, and that submodule is actually data and should be in `/unit-test/data`
- I moved mock-browser-api.js from background, because again, it's not a test. Instead it's now in `/unit-test/helpers`
- Finally, I grouped all reference tests implementation files (actual tests running reference-tests data) in one folder

## Steps to test this PR:
1. Run the tests locally! Or trust that Github Actions is doing it well :) 
2. Since content-scope script was touched, we should also test that content script privacy features are correctly enabled/disabled when page is on the unprotected list or allowlisted by the user

## Automated tests:
- [ ] Unit tests
- [ ] Integration tests

###### Reviewer Checklist:

###### PR Author Checklist:


Gruntfile.js
```diff
@@ -42,7 +42,7 @@ module.exports = function (grunt) {
             '<%= dirs.public.css %>/autofill.css': ['<%= ddgAutofill %>/autofill.css']
         },
         unitTest: {
-            '<%= dirs.unitTest.build %>/background.js': ['<%= dirs.unitTest.background %>/**/*.js', '!<%= dirs.unitTest.background %>/reference-tests/**/*.js'],
```
kdzwinel, line 45: we no longer need to exclude reference-tests repo, because it's not in the tests folder

unit-test/background/reference-tests/tracker-allowlist-tests.js
```diff
@@ -0,0 +1,31 @@
+const tdsStorageStub = require('../../helpers/tds.es6')
```
kdzwinel, line 1: this file was moved and paths were updated, not sure why it's considered new

dharb (approved):
Code looks good, tests are all passing locally. Thank you for doing this!

PR #964, opened 2021-12-09
Promisify message sending in the Click to Play code

kzar:
**Reviewer**: @jonathanKingston
**Cc**: @Charlie-belmer, @ladamski

## Description:

Small change to promisify the `sendMessage` function in the Click to Play code. I split this out from my project work as it changes the indentation level of quite a lot of code and was going to make reviewing my changes a pain.

## Steps to test this PR:
1. Ensure the Facebook Click to Load functionality still works OK on the test page https://privacy-test-pages.glitch.me/privacy-protections/click-to-load/

## Automated tests:
- [x] Unit tests
- [x] Integration tests

###### Reviewer Checklist:

###### PR Author Checklist:

### DIFF
--- a/shared/js/content-scripts/click-to-load.js
+++ b/shared/js/content-scripts/click-to-load.js
@@ -1,6 +1,14 @@
 (function clickToLoad () {
-    function sendMessage (messageType, options, callback) {
-        chrome.runtime.sendMessage({ messageType, options }, callback)
+    function sendMessage (messageType, options = {}) {
+        return new Promise((resolve, reject) => {
+            chrome.runtime.sendMessage({ messageType, options }, response => {
+                if (chrome.runtime.lastError) {
+                    reject(new Error(chrome.runtime.lastError.message))
+                } else {
+                    resolve(response)
+                }
+            })
+        })
     }
     let appID
     const loadingImages = {
@@ -597,7 +605,7 @@
      * @param {Object} widgetData - a single entry from elementData
      * @param {Element} originalElement - the element on the page we are replacing
      */
-    function createReplacementWidget (entity, widgetData, originalElement) {
+    async function createReplacementWidget (entity, widgetData, originalElement) {
         // Construct the widget based on data in the original element
         const widget = new DuckWidget(widgetData, originalElement, entity)
         const parent = originalElement.parentNode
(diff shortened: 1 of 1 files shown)


shared/js/content-scripts/click-to-load.js
```diff
@@ -1,6 +1,8 @@
 (function clickToLoad () {
-    function sendMessage (messageType, options, callback) {
-        chrome.runtime.sendMessage({ messageType, options }, callback)
+    function sendMessage (messageType, options = {}) {
+        return new Promise((resolve, reject) => {
+            chrome.runtime.sendMessage({ messageType, options }, resolve)
```
jonathanKingston, line 4: This doesn't handle .lastError like the mozilla shim does that we load elsewhere. Can we bundle that here instead?

shared/js/content-scripts/click-to-load.js
```diff
@@ -673,20 +673,17 @@
     })
 
     // Fetch reusable assets
-    sendMessage('getLoadingImage', 'light',
-        function (response) {
-            loadingImages.lightMode = response
-        })
+    sendMessage('getLoadingImage', 'light').then(response => {
```
jonathanKingston, line 676: When looking at all this code before it all looks like it races, is that the case?

Can we file a follow up issue (not a priority) to consolidate these messages too as it's not needed anymore iiuc too.

shared/js/content-scripts/click-to-load.js
kzar: > This doesn't handle .lastError...

Yea, I nearly left a comment about this but decided against. My thinking was that we don't handle errors with the existing code either, I was trying to avoid changing the behaviour.

> the mozilla shim...

Could you point me at the existing shim you'd like me to include?

shared/js/content-scripts/click-to-load.js
kzar: Yea, I think you're right that this code races. Agree that we should we should consolidate (or if possible just remove) these messages in the future.

shared/js/content-scripts/click-to-load.js
kzar: I've tweaked `sendMessage` to handle lastError without the shim, does that look OK to you?

shared/js/content-scripts/click-to-load.js
jonathanKingston: Can you file one or two follow up issues here?

shared/js/content-scripts/click-to-load.js
jonathanKingston: https://github.com/duckduckgo/duckduckgo-privacy-extension/blob/8d236939180663b07876d8f494cbbbf6c27f0b79/shared/js/background/wrapper.es6.js is an example that uses the shim. I don't think we have anything that exists in the content of the page currently and I think we can push that to a follow up as it probably requires build script wrangling that we should avoid here :)


Again can you file the follow up to clean this up into using the shim?

jonathanKingston:
LGTM will approve once the follow up issues have been filed :)

shared/js/content-scripts/click-to-load.js
kzar: Done.

shared/js/content-scripts/click-to-load.js
kzar: Done.

jonathanKingston (approved):

PR #976, opened 2021-12-16
Don't display Click to Play placeholders for first-party domains

kzar:
Click to Play is intended to protect the user from tracking via
embedded third-party content. The placeholder elements should never be
displayed for first-party embedded content.

**CC:** @Charlie-belmer, @ladamski, @kdzwinel 

## Steps to test this PR:
1. Ensure Facebook Click to Load still works correctly: https://privacy-test-pages.glitch.me/privacy-protections/click-to-load/

## Automated tests:
- [ ] Unit tests
- [ ] Integration tests

###### Reviewer Checklist:

###### PR Author Checklist:

### DIFF
--- a/shared/js/background/message-handlers.js
+++ b/shared/js/background/message-handlers.js
@@ -65,6 +65,13 @@ export async function initClickToLoad (unused, sender) {
     const tab = tabManager.get({ tabId: sender.tab.id })
     const config = { ...tdsStorage.ClickToLoadConfig }
 
+    // Remove first-party entries.
+    const siteUrlSplit = tab.site.domain.split('.')
+    const websiteOwner = trackers.findWebsiteOwner({ siteUrlSplit })
+    if (websiteOwner) {
+        delete config[websiteOwner]
+    }
+
     // Determine whether to show one time messages or simplified messages
     for (const [entity] of Object.entries(config)) {
         const clickToLoadClicks = settings.getSetting('clickToLoadClicks') || {}


shared/js/background/message-handlers.js
```diff
@@ -65,6 +65,13 @@ export async function initClickToLoad (unused, sender) {
     const tab = tabManager.get({ tabId: sender.tab.id })
     const config = { ...tdsStorage.ClickToLoadConfig }
 
+    // Remove first-party entries.
+    const siteUrlSplit = utils.extractHostFromURL(sender.url).split('.')
+    const websiteOwner = trackers.findWebsiteOwner({ siteUrlSplit })
+    if (websiteOwner) {
```
jonathanKingston, line 71: Isn't this just checking if the frame that sent the message is on the entity list? Shouldn't we be comparing the tab and sender url for the same entity?

jonathanKingston (changes requested):
Please see comment, I don't think this will work as expected.

shared/js/background/message-handlers.js
kzar: Yes, good point. So far, this is checking if the sending frame's domain belongs to the entity in question.

I suppose the most thorough approach would be to compare the sending frame's domain, then traverse up the frame hierarchy comparing each frame's domain in turn until we reached the top-level frame. 

That said, the logic below this that checks if for an `excludedDomains` match seems to just check the top-level frame's domain for a match: https://github.com/duckduckgo/duckduckgo-privacy-extension/blob/34671acc8288c9fd37ec35021c7de00d3c43ed7a/shared/js/background/message-handlers.js#L84-L89 

If I understand it right, the logic to decide weather to block a CTP request also just checks the top-level frame's  domain: 

https://github.com/duckduckgo/duckduckgo-privacy-extension/blob/34671acc8288c9fd37ec35021c7de00d3c43ed7a/shared/js/background/redirect.es6.js#L102-L103

So perhaps we should just check the top-level frame's domain here too? I suppose the main thing  is that we're consistent, since we really want to avoid the situation where we display a placeholder for an unblocked frame, or block a frame which we then don't display a placeholder for.

What do you think?

jonathanKingston: I think it's the comment that threw me in all honesty. It's certainly not a first party check happening here.

My understanding your intent here is so that for example facebook.com doesn't show click to load buttons. This should be checking the tab URL only?

Traversing the frame stack is used for things like CSP where all parties are involved in what can frame another, in this case we just want to treat the website as Facebook and not break it.

kzar: OK cool, sounds like we're agreed that just checking the top-level frame's domain is the expected behaviour.

kzar: Done.

jonathanKingston (approved):
LGTM I tested on privacy test pages and couldn't see any breakages.

I couldn't find a test case to manually verify we are fixing though.

PR #977, opened 2021-12-16
Check excludedDomains before blocking/redirecting Click to Play requests

kzar:
So far, the excludedDomains option in the Click to Play configuration
stops the Click to Play placeholders from being shown, but not the
corresponding requests from being blocked or redirected. Let's correct
that here.

**CC:** @Charlie-belmer, @ladamski, @kdzwinel 

## Steps to test this PR:
1. Check that Click to Play is still working OK for Facebook https://privacy-test-pages.glitch.me/privacy-protections/click-to-load/

## Automated tests:
- [ ] Unit tests
- [ ] Integration tests

###### Reviewer Checklist:

###### PR Author Checklist:

### DIFF
--- a/shared/js/background/redirect.es6.js
+++ b/shared/js/background/redirect.es6.js
@@ -102,6 +102,15 @@ function handleRequest (requestData) {
             if (tracker && socialTracker && trackerutils.shouldBlockSocialNetwork(socialTracker.entity, thisTab.site.url)) {
                 if (!trackerutils.isSameEntity(requestData.url, thisTab.site.url) && // first party
                     !thisTab.site.clickToLoad.includes(socialTracker.entity)) {
+                    // Ensure this website isn't in the excluded domains list.
+                    if (socialTracker.data && socialTracker.data.excludedDomains) {
+                        for (const { domain: excludedDomain } of socialTracker.data.excludedDomains) {
+                            if (excludedDomain === thisTab.site.domain) {
+                                return
+                            }
+                        }
+                    }
+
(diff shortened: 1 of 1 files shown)


shared/js/background/redirect.es6.js
```diff
@@ -102,6 +102,15 @@ function handleRequest (requestData) {
             if (tracker && socialTracker && trackerutils.shouldBlockSocialNetwork(socialTracker.entity, thisTab.site.url)) {
                 if (!trackerutils.isSameEntity(requestData.url, thisTab.site.url) && // first party
                     !thisTab.site.clickToLoad.includes(socialTracker.entity)) {
+                    // Ensure this website isn't in the excluded domains list.
+                    if (socialTracker.data && socialTracker.data.excludedDomains) {
```
jonathanKingston, line 106: Can you outline the difference between the check above `trackerutils.shouldBlockSocialNetwork`?
This uses the cache which allows the user to exempt the load later.

Could you explain why this is needed or if a change to the mentioned method would be better. Perhaps another trackerutils method would clean this up if it is needed?

kzar: I've looked into this again today and to be honest I can't see why this change is necessary. Like you point out, `trackerutils.shouldBlockSocialNetwork` should take care of this and running that function through a debugger it seems to work OK. Since I wrote this before Christmas I might be forgetting some details, but I'm going to close this for now and will re-open if it turns out there's something I'm forgetting.

PR #983, opened 2021-12-20
fix(integration-tests): create and cleanup temp profile directories f…

shakyShane:
…or puppeteer

**Reviewer: @jonathanKingston**

## Description:
Running integration tests locally causes `temp-profile-*` folders to be created that sometimes cannot be cleaned up.

image

This PR exposes a `teardown` function on the return value of `harness.setup()` that when called will clean up the temp directory. This is an important step for us to get correct since every run of integration tests can create approx 18mb of data.

The `tempy` package is used to generate a valid temp dir on any platform.

## Automated tests:
- [ ] ~Unit tests~
- [x] Integration tests

###### Reviewer Checklist:

###### PR Author Checklist:

### DIFF
--- a/integration-test/helpers/harness.js
+++ b/integration-test/helpers/harness.js
@@ -1,15 +1,18 @@
 const puppeteer = require('puppeteer')
-const execSync = require('child_process').execSync
+const spawnSync = require('child_process').spawnSync
+const tempy = require('tempy')
 
 jasmine.DEFAULT_TIMEOUT_INTERVAL = 20000
 if (process.env.KEEP_OPEN) {
     jasmine.DEFAULT_TIMEOUT_INTERVAL = 20000 * 1000
 }
 
+const DATA_DIR_PREFIX = 'ddg-temp'
+
 const setup = async (ops) => {
     ops = ops || {}
 
-    const dataDir = process.env.CI ? `/tmp/profile${Math.random()}` : `temp-profile-${Math.random()}`
+    const dataDir = tempy.directory({ prefix: DATA_DIR_PREFIX })
     const puppeteerOps = {
         args: [
             '--disable-extensions-except=build/chrome/dev',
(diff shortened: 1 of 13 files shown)


integration-test/helpers/harness.js
```diff
@@ -63,35 +66,34 @@ const setup = async (ops) => {
 
     bgPage.on('request', (req) => { requests.push(req.url()) })
 
-    return { browser, bgPage, requests }
+    return { browser, bgPage, requests, teardown: teardown.bind(null, browser, dataDir) }
```
jonathanKingston, line 69: Just so I understand, is this to prevent multiple tests at the same time running?

Can we use a closure instead, largely we avoid `bind` calls.

integration-test/helpers/harness.js
```diff
@@ -1,15 +1,18 @@
 const puppeteer = require('puppeteer')
-const execSync = require('child_process').execSync
+const spawnSync = require('child_process').spawnSync
+const tempy = require('tempy')
```
jonathanKingston, line 3: How is this different to mkdtempSync?

integration-test/helpers/harness.js
shakyShane: It's mostly just to bind a teardown method for each call to `setup()` - this allows us to cleanly dispose of temporary data without having to introduce another param that would need to be given to 'teardown'.

For example, to correctly cleanup after a test run, without the introduced closure/bound function, we'd end up with:

**before**
```js
// in setup 
({browser, dataDir}) = await harness.setup();

// later in cleanup
await harness.teardown(browser, dataDir)
```

Instead, I suggest we switch to the well-used pattern of the `setup()` call returning it's own teardown logic. This gives us scope in the future to increase/decrease the work done in teardown without it affecting the interface.

**after**
```js
// in setup 
({browser, teardown}) = await harness.setup();

// later in cleanup
await teardown()
```

In terms of avoiding the `bind` call, yeah we could just move the teardown function into the setup (now that it's not exposed manually) 👍🏻

integration-test/helpers/harness.js
```diff
@@ -63,35 +66,34 @@ const setup = async (ops) => {
 
     bgPage.on('request', (req) => { requests.push(req.url()) })
 
-    return { browser, bgPage, requests }
-}
+    async function teardown (browser, dataDir) {
```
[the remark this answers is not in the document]
shakyShane: Yep!

jonathanKingston (approved):

PR #1004, opened 2022-01-04
run eslint on more JS files + re-use argument list in package.json

shakyShane:

## Description:
run eslint on more/all JS files + re-use argument list in package.json (the `--fix` variant was not equal to the regular one

This also ensures that `.js` files in the root are picked up too (like `Gruntfile` etc)


package.json
```diff
@@ -15,8 +15,8 @@
         "dev": "echo 'Try npm run dev-firefox or npm run dev-chrome instead' && exit 0",
         "release": "grunt build",
         "selenium": "cd selenium-test && node test.js",
-        "lint": "eslint shared/js unit-test integration-test scripts",
-        "lint-fix": "eslint shared/js unit-test integration-test --fix",
+        "lint": "eslint '*.js' shared/js shared/data unit-test integration-test scripts",
+        "lint-fix": "npm run lint -- --fix",
```
shakyShane, line 19: with this, the list of files to `--fix` cannot become out of sync 👌🏻

jonathanKingston (approved):

PR #1006, opened 2022-01-07
Remove unused pixel tracking

jonathanKingston:
**Reviewer:** @kdzwinel 

## Description:

Removes all tracking pixels other than the site report as it was all unused.

I renamed the class and hard coded the pixel name so it can't be used again without reimplementing it as it breaks addon policy.

https://app.asana.com/0/892838074342800/1200747843240318/f

## Steps to test this PR:

1. Open up the browser inspector
2. Report a broken site
3. Ensure the report is firing correctly

## Automated tests:
- [ ] Unit tests
- [ ] Integration tests

###### Reviewer Checklist:

###### PR Author Checklist:

### DIFF
--- a/shared/js/background/message-handlers.js
+++ b/shared/js/background/message-handlers.js
@@ -7,7 +7,7 @@ const trackerutils = require('./tracker-utils')
 const trackers = require('./trackers.es6')
 const constants = require('../../data/constants')
 const Companies = require('./companies.es6')
-const pixel = require('./pixel.es6')
+const brokenSiteReport = require('./broken-site-report')
 const browserName = utils.getBrowserName()
 const devtools = require('./devtools.es6')
 const browserWrapper = require('./wrapper.es6')
@@ -33,11 +33,8 @@ export function getBrowser () {
     return browserName
 }
 
-export function firePixel (fireArgs) {
-    if (fireArgs.constructor !== Array) {
-        fireArgs = [fireArgs]
-    }
-    return pixel.fire.apply(null, fireArgs)
+export function submitBrokenSiteReport (brokenSiteArgs) {
+    return brokenSiteReport.fire.apply(null, brokenSiteArgs)
 }
 
 export function getTab (tabId) {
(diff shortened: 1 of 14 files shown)


shared/js/background/message-handlers.js
```diff
@@ -33,11 +33,8 @@ export function getBrowser () {
     return browserName
 }
 
-export function firePixel (fireArgs) {
-    if (fireArgs.constructor !== Array) {
-        fireArgs = [fireArgs]
-    }
-    return pixel.fire.apply(null, fireArgs)
+export function submitBrokenSiteReport (brokenSiteArgs) {
+    return brokenSiteReport.fire.apply(null, brokenSiteArgs)
```
jonathanKingston, line 37: This is needed to proxy from the UI as a message handler into the broken site report class.

jonathanKingston: There's only one call site and it's always an array so there's no point supporting the other case now.

shared/js/background/message-handlers.js
```diff
@@ -33,11 +33,8 @@ export function getBrowser () {
     return browserName
 }
 
-export function firePixel (fireArgs) {
-    if (fireArgs.constructor !== Array) {
```
kdzwinel, line 37: any idea why we had this check before?

unit-test/background/pixel.es6.js
```diff
@@ -1,5 +1,5 @@
```
kdzwinel, line 1: hm, since we changed the original file being testsed here, I think we should also rename the test file

kdzwinel (commented):
Thanks for doing this! I left two comments, but other than that it looks and works great.

unit-test/background/pixel.es6.js
jonathanKingston: Yeah I was in two minds about this (as it's only testing pixels and not the whole thing), I think I agree :)

shared/js/background/message-handlers.js
jonathanKingston: `firePixel('eph')` was used in a few places for example and it also supported the `firePixel(['eph', ...])`

kdzwinel (approved):
Nicely done!

PR #1014, opened 2022-01-12
Reference tests update and surrogate reporting fix

kdzwinel:
**Reviewer:** @sammacbeth 

## Description:
This PR updates reference tests after latest changes to HTTPS tests and small adjustment to broken site reporting tests - https://github.com/duckduckgo/privacy-reference-tests/pull/24 . It also fixes a bug in how surrogates are reported in the breakage reports (previously surrogated hosts went into both `blockedTrackers` and `surrogates` fields, they should have gone only to `surrogates`).

## Steps to test this PR:
1. Send a breakage report from a site that uses a surrogate (e.g. https://privacy-test-pages.glitch.me/tracker-reporting/1major-with-surrogate.html) and double check that report being sent only mentions surrogate in the `surrogates` field and not the `blockedTrackers` field of the report
2. Double check that blocked trackers are still correctly sent with `blockedTrackers`

## Automated tests:
- [ ] Unit tests
- [ ] Integration tests

###### Reviewer Checklist:

###### PR Author Checklist:

### DIFF
--- a/unit-test/background/reference-tests/broken-site-reporting-tests.js
+++ b/unit-test/background/reference-tests/broken-site-reporting-tests.js
@@ -64,7 +64,7 @@ for (const setName of Object.keys(testSets)) {
 
                 if (test.expectReportURLParams) {
                     test.expectReportURLParams.forEach(param => {
-                        expect(requestURLString).toContain(`${param.name}=${param.value}`)
+                        expect(requestURLString).toMatch(new RegExp(`[?&]${param.name}=${param.value}[&$]`))
                     })
                 }
             })
--- a/unit-test/background/reference-tests/broken-site-reporting-tests.js
+++ b/unit-test/background/reference-tests/broken-site-reporting-tests.js
@@ -63,8 +63,12 @@ for (const setName of Object.keys(testSets)) {
                 }
 
                 if (test.expectReportURLParams) {
+                    const requestUrl = new URL(requestURLString)
+                    // we can't use searchParams because those are automatically decoded
+                    const searchItems = requestUrl.search.split('&')
+
                     test.expectReportURLParams.forEach(param => {
-                        expect(requestURLString).toMatch(new RegExp(`[?&]${param.name}=${param.value}[&$]`))
+                        expect(searchItems).toContain(`${param.name}=${param.value}`)
                     })
                 }
             })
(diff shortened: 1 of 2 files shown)


unit-test/background/reference-tests/broken-site-reporting-tests.js
```diff
@@ -64,7 +64,7 @@ for (const setName of Object.keys(testSets)) {
 
                 if (test.expectReportURLParams) {
                     test.expectReportURLParams.forEach(param => {
-                        expect(requestURLString).toContain(`${param.name}=${param.value}`)
+                        expect(requestURLString).toMatch(new RegExp(`[?&]${param.name}=${param.value}[&$]`))
```
kdzwinel, line 67: tests were not catching the surrogates/blockedTrackers issue because I used too simple method of matching params, regex matching fixes it

jonathanKingston, line 67: Can we use the URL constructor here?

Untested but something like this:
```suggestion
                        let url = new URL(requestURLString)
                        expect(url.searchParams.get(param.name)).equals(param.value)
```

kdzwinel: I used that initially but URL object automatically decodes URL params and I get test failures:

> Expected 'https://example.test/' to be 'https%3A%2F%2Fexample.test%2F'

I wanted tests to check if parameters are properly url encoded, so I don't want to adjust tests. I guess I could encode those parameters that I will get back again, but this is not great either.

jonathanKingston: Ah makes sense, I still think I'd prefer using encodeURIComponent if that works than RegExp.

sammacbeth: `param.value` is already URLencoded here right? I wonder how safe it is to just drop this into a Regex unsanitized? Are there any Regex special characters that could come through URL encoding? Are we OK allowing the reference test repo to run any regex here if it wants to?

kdzwinel: Yep, param value is already encoded in tests. I don't think there is any security risk here (we control all inputs, this is test code), but I agree that it can happen that we will write a test that has some special chars and end up with unexpected results. I'll fix 👍

kdzwinel: Bleh, encodeURIComponent behaves differently than encodeURI and, unexpectedly, encodes `,`. I'll have to find a different way.

sammacbeth (approved):
LGTM!

PR #1016, opened 2022-01-13
Split apart createReplacementWidget function and add events

kzar:
Split out the DuckWidget creation and element replacing logic from
createReplacementWidget function. Tidy up some of the function and
variable names to make the code easier to understand. Begin emitting
events when a tracking element is replaced, so that the surrogate
scripts can tie tracking elements to their corresponding placeholder.

**CC:** @ladamski, @kdzwinel, @Charlie-belmer 

## Description:
See above.

## Steps to test this PR:
1. Check that the Facebook CTP still works as expected: https://privacy-test-pages.glitch.me/privacy-protections/click-to-load/

## Automated tests:
- [ ] Unit tests
- [ ] Integration tests

###### Reviewer Checklist:

###### PR Author Checklist:

### DIFF
--- a/shared/js/content-scripts/click-to-load.js
+++ b/shared/js/content-scripts/click-to-load.js
@@ -615,41 +615,48 @@
         replaceClickToLoadElements(extensionResponseData)
     }
 
-    /**
-     * Creates a safe element and replaces the original tracking element with it.
-     * @param {string} entity
-     *   The entity (e.g. company) associated with the tracking element.
-     * @param {Object} widgetData
-     *   A single entry from the CTP configuration file.
-     * @param {Element} originalElement
-     *   The tracking element on the page which we're replacing with a placeholder.
-     */
-    async function createReplacementWidget (entity, widgetData, originalElement) {
-        // Construct the widget based on data in the original element
-        const widget = new DuckWidget(widgetData, originalElement, entity)
-        const parent = originalElement.parentNode
+    function replaceTrackingElement (widget, trackingElement, placeholderElement) {
+        widget.dispatchEvent(trackingElement, 'ddg-ctp-tracking-element')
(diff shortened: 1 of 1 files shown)


shared/js/content-scripts/click-to-load.js
```diff
@@ -663,10 +670,11 @@
...
         for (const entity of Object.keys(config)) {
-            for (const widget of Object.values(config[entity].elementData)) {
-                const els = document.querySelectorAll(widget.selectors.join())
-                for (const el of els) {
-                    createReplacementWidget(entity, widget, el)
+            for (const widgetData of Object.values(config[entity].elementData)) {
+                const trackingElements = document.querySelectorAll(widgetData.selectors.join())
+                for (const trackingElement of trackingElements) {
+                    const widget = new DuckWidget(widgetData, trackingElement, entity)
+                    createPlaceholderElementAndReplace(widget, trackingElement)
```
sammacbeth, line 677: This is an async function, but you're not assigning the promise returned by it. I wonder if this could introduce a race condition if a subsequent call should await the placeholders being created?

sammacbeth (approved):
Mostly looks fine to me (most of the diff just seems to be variable renaming). Just one minor question, but you can merge if that is not an issue.

shared/js/content-scripts/click-to-load.js
kzar: Yea, you're completely right. I debated `await`ing here, but decided against. I figured we don't need (or want) to perform these calls in series and we don't (yet!) need to wait until `replaceClickToLoadElements` has finished. With my next change I will make `replaceClickToLoadElements` wait for those calls  so that we can dispatch a `"ddg-ctp-ready"` event reliably.

kzar:
Thanks for the review @sammacbeth :+1:

PR #1037, opened 2022-01-27
Enable browsingTopics protection

SlayterDev:
**Reviewer:** @jonathanKingston, @sammacbeth 

## Description:

https://app.asana.com/0/1163321984198618/1201719300245228/f

`browsingTopics` is effectively replacing `floc`. This PR turns our `floc` protection into a `browsingTopics` protection.

## Steps to test this PR:
Browsing topics is not yet deployed in current versions of chrome and we're basing this protection on currently available documentation ([see here](https://blog.google/products/chrome/get-know-new-topics-api-privacy-sandbox/)). The content script can be verified by adding a log line to ensure the script is properly injected.

## Automated tests:
- [ ] Unit tests
- [ ] Integration tests

###### Reviewer Checklist:

###### PR Author Checklist:


README.md
```diff
@@ -18,7 +18,7 @@ The extension provides users with an improved private browsing experience whilst
 - [Announces to a website they must not track or sell on your data](https://globalprivacycontrol.org/)
 - Trims referrer header data
 - Reduced fingerprintability of various browser features
-- Removes FLoC from Chromium based browsers
+- Removes FLoC/browsingTopics from Chromium based browsers
```
kdzwinel, line 21: I think we can drop "FLoC" now that browsing-topics effectively replaced it.

shared/data/bundled/extension-config.json
```diff
@@ -1,6 +1,6 @@
 {
     "readme": "https://github.com/duckduckgo/privacy-configuration",
-    "version": 1642335495631,
+    "version": 1643197961938,
```
kdzwinel, line 3: IMO we should not mix config updates with unrelated PRs, but rather have one PR that updates the config just before the release.

shared/js/background/events.es6.js
```diff
@@ -185,8 +185,8 @@ const extraInfoSpec = ['blocking', 'responseHeaders']
 if (browser.webRequest.OnHeadersReceivedOptions.EXTRA_HEADERS) {
     extraInfoSpec.push(browser.webRequest.OnHeadersReceivedOptions.EXTRA_HEADERS)
 }
-// we determine if FLoC is enabled by testing for availability of its JS API
-const isFlocEnabled = ('interestCohort' in document)
+// we determine if browsingTopics is enabled by testing for availability of its JS API
+const isTopicsEnabled = ('browsingTopics' in document) && utils.isFeatureEnabled('browsingTopics')
```
kdzwinel, line 189: oh, we overlooked isFeatureEnabled before - does that mean that there was no way to disable it via config?

kdzwinel (commented):
Nice work @SlayterDev ! I left couple of comments. Also I understand that this is blocked until we merge https://github.com/duckduckgo/content-scope-scripts/pull/17/files ? We should also wrap up internal discussion and figure out how to test it before we merge.

shared/js/background/events.es6.js
jonathanKingston: Only for the header yeah.

shared/data/bundled/extension-config.json
jonathanKingston: I think we'd want this tbf as the key has changed. We could probably mark this file as a binary as it's not really that meaningful anyway.

SlayterDev: Agreed. Using our current flow for testing config changes this updated the embedded config when I removed the test version. When we're ready to ship we'll use the version that includes this feature.

shared/js/background/events.es6.js
SlayterDev: Specifically the header piece of this. The Content scope script would've been disabled as expected.

jonathanKingston (changes requested):
I think we're missing some changes downstream but everything appears well with the approach.

I think we should not point to the commit in PRs like we do elsewhere, happy to discuss but I think it will make PRs easier etc and just state to the reviewer they need to use `npm link @duckduckgo/content-scope-scripts` for example.

shared/data/etags.json
```diff
@@ -1 +1 @@
-{"config-etag":"W/\"553f1008cd87e248350713bfbdc89868\""}
```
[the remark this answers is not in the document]
SlayterDev: I checked out the version from develop and for some reason it still added a newline.

jonathanKingston:
Just going to wait until we can cut a tag on content scope scripts but other than that we should be good here to merge :)

shared/data/etags.json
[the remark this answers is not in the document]
jonathanKingston: Can you just run:
```
git checkout origin/develop -- shared/data/etags.json
git checkout origin/develop -- package-lock.json
```

Neither should be needed here and we rely on just updating etags for the release as it changes so often.

jonathanKingston (approved):

PR #978, opened 2021-12-16
Implement YouTube Click to Load

kzar:
Add privacy protection from embedded YouTube videos and
playlists. Block third-party YouTube requests and display a
placeholder for the corresponding element that allows users to
unblock the content again. Where possible, also upgrade embedded
videos to use youtube-nocookie.com.

**Reviewer:** @jonathanKingston 

**CC:** @ladamski, @kdzwinel, @Charlie-belmer, @sammacbeth  

## Steps to test this PR:
1. Add the following entry to your `social_ctp_configuration.json` file: https://gist.github.com/kzar/89bc3b8971daef09334da5a0799589c0
2. Navigate to the YouTube test page and follow the instructions: https://privacy-test-pages.glitch.me/privacy-protections/youtube-click-to-load/

## Automated tests:
- [ ] Unit tests
- [x] Integration tests

###### Reviewer Checklist:
- [ ] **Ensure the PR solves the problem**
- [ ] **Review every line of code**
- [ ] **Ensure the PR does no harm by testing the changes thoroughly**
- [ ] **Get help if you're uncomfortable with any of the above!**
- [ ] Determine if there are any quick wins that improve the implementation

###### PR Author Checklist:
- [ ] Get advice or leverage existing code
- [ ] Agree on technical approach with reviewer (if the changes are nuanced)
- [ ] Ensure that there is a testing strategy (and documented non-automated tests)
- [ ] Ensure there is a documented monitoring strategy (if necessary)
- [ ] Consider systems implications

PR #1039, opened 2022-01-28
Improve Facebook Click to Load integration tests 

kzar:
Improve Facebook Click to Load integration tests

Improvements to the Facebook Click to Load integration tests,
including:

 1. Reduce flaky failures by waiting until all the required extension
    configuration has been loaded.
 2. Log requests using Puppeteer instead of relying on the test page
    to accurately report the number of allowed/blocked requests.
 3. Test requests to the SDK are redirected properly.
 4. Test that unblocking works correctly and that reloading the page
    restores blocking.

This change also:

 - Renames and expands on the background waiting integration test
   helper module. That way, we can avoid duplicating configuration
   waiting logic.
 - Tweaks the request logging helper module to avoid logging in-flight
   requests that started before a navigation but finished
   afterwards. That race condition was the cause of some test flakiness.
 - Tweaks the YouTube CTL tests to wait for "networkidle0" instead of
   "networkidle2". That means waiting for all requests to complete,
   rather than all but two.

**CC:** @ladamski, @Charlie-belmer, @sammacbeth, @kdzwinel 
**Depends on:** https://github.com/duckduckgo/duckduckgo-privacy-extension/pull/978

## Description:
This change makes use of the YouTube Click to Load work to improve the Facebook Click to Load integration tests.

## Steps to test this PR:
1. Run the integration tests (to do that locally type `npm run test-int`.

## Automated tests:
- [ ] Unit tests
- [x] Integration tests

###### Reviewer Checklist:

###### PR Author Checklist:


integration-test/background/click-to-load-facebook.js
```diff
@@ -1,23 +1,67 @@
 const harness = require('../helpers/harness')
+const { logPageRequests } = require('../helpers/requests')
 
 const testSite = 'https://privacy-test-pages.glitch.me/privacy-protections/click-to-load/'
+const facebookDomains = new Set(['facebook.com', 'facebook.net'])
```
kdzwinel, line 5: I'd also consider fbcdn.net (https://slayterdev.github.io/tracker-radar-wiki/entities/Facebook,%20Inc..html) it loads quite a bit of stuff on the privacy test page.

integration-test/background/click-to-load-facebook.js
```diff
@@ -1,23 +1,67 @@
...
 
+function summariseFacebookRequests (requests) {
+    const facebookSDKRedirect = { checked: false, alwaysRedirected: true }
+    // Note: Requests to the SDK are not included in counts.
+    let requestCount = 0
+    let allowCount = 0
+    let blockCount = 0
+
+    for (const request of requests) {
+        if (facebookDomains.has(request.url.hostname.split('.').slice(-2).join('.'))) {
```
kdzwinel, line 19: this line is way too smart, how about we use tldts to extract the eTLD+1?

integration-test/background/click-to-load-facebook.js
```diff
@@ -1,23 +1,67 @@
...
+            if (request.url.pathname === '/en_US/sdk.js') {
+                facebookSDKRedirect.alwaysRedirected = (
+                    facebookSDKRedirect.alwaysRedirected &&
```
kdzwinel, line 22: smart, probably too smart, drop a comment that this sdk is loading multiple times on the page so that's why we need this check

integration-test/background/click-to-load-facebook.js
```diff
@@ -1,23 +1,67 @@
...
-            { polling: 100, timeout: 10000 }
+            () => (
+                window.dbg &&
+                window.dbg.tds &&
+                !window.dbg.tds.isInstalling &&
+                window.dbg.tds.ClickToLoadConfig &&
+                window.dbg.tds.tds.trackers &&
+                window.dbg.tds.tds &&
+                window.dbg.tds.tds.domains &&
+                window.dbg.tds.tds.domains['facebook.com'] === 'Facebook, Inc.'
```
kdzwinel, line 61: oh boy, that's a lot, do we need all those checks? Can we compress it with some `?.` magic?

integration-test/background/click-to-load-facebook.js
```diff
@@ -26,61 +70,74 @@ describe('Test Facebook Click To Load', () => {
...
-                expectedCalls: document.getElementById('facebook_iFrames').innerHTML
-            }
+        // Once the user clicks to load the Facebook content, the SDK should be
+        // loaded and the content should be unblocked.
+        pageRequests.length = 0
+        const buttonCount = await page.evaluate(() => {
+            window.buttons =
+                Array.from(document.querySelectorAll('body > div'))
+                    .map(div => div.shadowRoot && div.shadowRoot.querySelector('button'))
+                    .filter(button => button)
```
kdzwinel, line 103:
```suggestion
                    .filter(div => div.shadowRoot && div.shadowRoot.querySelector('button'))
```

I *think* that should work?

integration-test/background/click-to-load-facebook.js
```diff
@@ -26,61 +70,74 @@ describe('Test Facebook Click To Load', () => {
...
+            return window.buttons.length
```
kdzwinel, line 104: hm, why are we returning count and not references to buttons? can we return array of "handles" and then perform clicks?

integration-test/background/click-to-load-facebook.js
```diff
@@ -26,61 +70,74 @@ describe('Test Facebook Click To Load', () => {
...
-        const page = await browser.newPage()
-
-        try {
-            await page.goto(testSite, { waitUntil: 'networkidle2', timeout: 10000 })
-        } catch (e) {
-            // timed out waiting for page to load, let's try running the test anyway
-            console.log(`Timed out: ${e}`)
+        for (let i = 0; i < buttonCount; i++) {
+            const button = await page.evaluateHandle(i => window.buttons[i], i)
+            await button.click()
```
kdzwinel, line 108: do we care if those are clicked one by one? WHy not click all at once?

integration-test/background/click-to-load-facebook.js
```diff
@@ -26,61 +70,74 @@ describe('Test Facebook Click To Load', () => {
...
+        await page.waitForNetworkIdle({ idleTime: 1000 })
+        {
+            const {
+                requestCount, blockCount, allowCount, facebookSDKRedirect
+            } = summariseFacebookRequests(pageRequests)
+
+            expect(facebookSDKRedirect.checked).toBeTrue()
+            // FIXME - It seems that requests to the SDK are not reliably
+            //         redirected after the page is reloaded.
+            // expect(facebookSDKRedirect.alwaysRedirected).toBeTrue()
```
kdzwinel, line 135: so weird, we don't see that in the regular browser, right?

kdzwinel (commented):
Nicely done! I left some comments, mostly nits 👀

integration-test/background/click-to-load-facebook.js
kzar: Well, no I don't think we care if they are clicked one-by-one, or what order they're clicked. In practice the buttons are clicked very fast however and I didn't see the point in optimising it. The main time taken by these tests is waiting for requests. Happy to rework to use `Promise.all` if you like?

integration-test/background/click-to-load-facebook.js
kzar: I suspect that we would, I spent quite a while debugging the tests to make sure it wasn't just a bug in the tests themselves. I filed an Asana task to follow up.

integration-test/background/click-to-load-facebook.js
kzar: Not waiting for all the configuration to load turned out to be the cause of a lot of the test flakes. To help reduce the flakes we had an arbitrary X second wait, but that also meant that the tests ran slower than necessary. I agree it's kind of ugly checking so many things here, but it makes the tests run reliably and quickly so it seems worth it.

I will see if I can split this logic out into a helper at least though.

integration-test/background/click-to-load-facebook.js
kzar: Gotya, Done.

integration-test/background/click-to-load-facebook.js
kzar: I think that this is made pretty clear by the variable names `facebookSDKRedirect.alwaysRedirected` and `facebookSDKRedirect.checked`. Our tests need to know that the SDK request was made at least one and if requests to the SDK were redirected every time it was made. I figure that knowing just if the last request to the SDK was redirected doesn't tell us much from a privacy respect.

integration-test/background/click-to-load-facebook.js
kzar: That would filter the elements correctly, but it would result in an array of the div elements instead of an array of the button elements.

integration-test/background/click-to-load-facebook.js
kzar: (I moved this logic out into the harness, that has already saved duplication and will help for future tests too.)

integration-test/background/click-to-load-facebook.js
kdzwinel: fair enough 👍

integration-test/background/click-to-load-facebook.js
kdzwinel: yup, my bad, thanks!

integration-test/background/click-to-load-facebook.js
kdzwinel: so what would happen if you used evaluateHandle and returned window.buttons in this case? Wouldn't that be a handle to the buttons array?

integration-test/background/click-to-load-facebook.js
kdzwinel: If there is no time difference, I don't mind keeping current solution 👍

integration-test/background/click-to-load-facebook.js
kdzwinel: OK, fair enough, we can figure this out in the follow up.

integration-test/helpers/harness.js
```diff
@@ -90,6 +90,23 @@ const setup = async (ops) => {
...
+        await bgPage.waitForFunction(
+            () => (
+                window.dbg &&
+                window.dbg.tds &&
+                !window.dbg.tds.isInstalling &&
+                window.dbg.tds.ClickToLoadConfig &&
+                window.dbg.tds.tds.trackers &&
+                window.dbg.tds.tds &&
+                window.dbg.tds.tds.domains &&
+                window.dbg.tds.tds.domains['facebook.com'] === 'Facebook, Inc.'
```
kdzwinel, line 104: is this something that we can generalize? will this entry always be in the configuration? I'm thinking that some tests might be mocking TDS and not having this entry.

kzar: Uh, maybe? I'm open to suggestions. For now, I just put this behind the `waitForConfiguration` option so that we don't break the existing tests. At least that way there's a way to wait for all configuration to load for tests that want to do that.

integration-test/background/click-to-load-facebook.js
kzar: That would return a `JSHandle` of the array, instead of an `ElementHandle` of one of the buttons. I think the `JSHandle` can be passed to a function that you're evaluating in the page, but I don't think it can be used directly. For example, I just tried this:

```javascript
        for (let button of await page.evaluateHandle(() => window.buttons)) {
            await button.click()
        }
```

and that resulted in an exception `TypeError: (intermediate value) is not iterable`

integration-test/helpers/harness.js
kzar: I've improved this logic now:
- I noticed that the `integration-tests/helpers/wait.js` module already took care of some aspects of background-page waiting, so I expanded on that and renamed the module to `backgroundWait.js`.
- Instead of hard-coding Facebook, I now check the first entity in the list.
- I made use of the fancy `?` syntantic sugar.
- I added some further checks for other aspects of the extension configuration.

integration-test/background/click-to-load-facebook.js
kdzwinel: Gotcha, thanks for double-checking!

kdzwinel:
Thanks! LGTM.

kdzwinel (approved):

PR #1041, opened 2022-02-02
Add API schema tests for Click to Load surrogate scripts

kzar:
The Click to Load feature redirects requests to certain scripts in
order to replace them with our "surrogate" scripts. It's important we
know when the original script has changed materially, so that we can
update the corresponding surrogate script. To that end, let's add
integration tests that check the schema of important Objects provided
by those scripts against the schema that we expect.

**Reviewer:** @kdzwinel 

**CC:** @jonathanKingston, @sammacbeth, @Charlie-belmer, @ladamski 
**Depends on:** https://github.com/duckduckgo/duckduckgo-privacy-extension/pull/978, https://github.com/duckduckgo/duckduckgo-privacy-extension/pull/1039

## Description:
Adds integration tests to alert us when the Facebook/YouTube scripts change, so that we have a chance to fix the corresponding surrogate scripts.

_Note: This is ready, but marking as WIP since this depends on other changes and I want to avoid an accidental squash._

## Steps to test this PR:
1. Run integration tests.

## Automated tests:
- [ ] Unit tests
- [x] Integration tests

###### Reviewer Checklist:

###### PR Author Checklist:


integration-test/background/click-to-load-youtube.js
```diff
@@ -0,0 +1,335 @@
+const fs = require('fs')
+const path = require('path')
+const harness = require('../helpers/harness')
+const { logPageRequests } = require('../helpers/requests')
+const { loadTestConfig, unloadTestConfig } = require('../helpers/testConfig')
+const { getObjectSchema } = require('../helpers/objectSchema')
+
+const testSite = 'https://privacy-test-pages.glitch.me/privacy-protections/youtube-click-to-load/'
+const testConfig = {
+    'dbg.tds.tds.trackers.youtube\\.com': {
```
kdzwinel, line 10: why only last `.` is escaped?
should we move that config to external file that we load here (like `youtube-iframe-api.json`)?

integration-test/background/click-to-load-youtube.js
```diff
@@ -0,0 +1,335 @@
...
+    'dbg.tds.tds.trackers.youtube-nocookie\\.com': {
+        owner: {
+            name: 'Google LLC',
+            displayName: 'YouTube',
+            privacyPolicy: 'https://policies.google.com/privacy?hl=en',
+            url: 'http://google.com'
+        },
+        default: 'ignore'
+    },
+    'dbg.tds.ClickToLoadConfig.Google LLC': {
```
kdzwinel, line 28: this looks weird, is that correct?

integration-test/background/click-to-load-youtube.js
```diff
@@ -0,0 +1,335 @@
...
+        }
+
+        let stats
+
+        if (youTubeStandardDomains.has(request.url.hostname)) {
+            stats = youTubeStandard
+        } else if (youTubeNocookieDomains.has(request.url.hostname)) {
+            stats = youTubeNocookie
+        } else {
+            continue
```
kdzwinel, line 112: we know that only other expected domain is glitch.me, should we (both here and in FB test) make some noise if we see something unexpected during the test? This could help us catch widget changes

integration-test/background/click-to-load-youtube.js
```diff
@@ -0,0 +1,335 @@
...
+        await bgPage.waitForFunction(
+            () => (
+                window.dbg &&
+                window.dbg.tds &&
+                !window.dbg.tds.isInstalling &&
+                window.dbg.tds.ClickToLoadConfig &&
+                window.dbg.tds.tds.trackers &&
+                window.dbg.tds.tds &&
+                window.dbg.tds.tds.domains &&
+                window.dbg.tds.tds.domains['youtube.com'] === 'Google LLC'
```
kdzwinel, line 143: ditto about `?.`

integration-test/background/click-to-load-youtube.js
```diff
@@ -0,0 +1,335 @@
...
+            expect(youTubeNocookie.blocked).toEqual(youTubeNocookie.total)
+            expect(youTubeNocookie.allowed).toEqual(0)
+        }
+
+        // Once the user clicks to load a video, the iframe_api should be loaded
+        // and the video should be unblocked.
+        pageRequests.length = 0
+        const button = await page.evaluateHandle(
+            'document.querySelector("div:nth-child(2) > div")' +
+            '.shadowRoot.querySelector("button")'
```
kdzwinel, line 190: do we have some line limits set up in the linter?

integration-test/background/click-to-load-youtube.js
```diff
@@ -0,0 +1,335 @@
...
+        )
+        await button.click()
```
kdzwinel, line 192: for facebook we are clicking all buttons, here we are clicking one, what's the logic behind this?

integration-test/background/click-to-load-youtube.js
```diff
@@ -0,0 +1,335 @@
...
+            expect(youTubeIframeApi.alwaysRedirected).toBeTrue()
+            expect(youTubeStandard.total).toBeGreaterThan(5)
+            expect(youTubeStandard.blocked).toEqual(youTubeStandard.total)
+            expect(youTubeStandard.allowed).toEqual(0)
+            expect(youTubeNocookie.blocked).toEqual(youTubeNocookie.total)
+            expect(youTubeNocookie.allowed).toEqual(0)
+        }
+
+        // The header button should also unblock YouTube.
+        pageRequests.length = 0
```
kdzwinel, line 224: should this be a separate `it` test?

integration-test/background/click-to-load-youtube.js
```diff
@@ -0,0 +1,335 @@
...
+                page.waitForFunction(expectedBorder => (
+                    document.getElementById('existing-video')
+                        .style.border.split(' ').pop() === expectedBorder
+                ),
+                { polling: 10 }, expectedBorder)
+
+            await waitForExpectedBorder('')
+
+            const button = await page.evaluateHandle(
+                'document.querySelector("div:nth-child(7) > div")' +
```
kdzwinel, line 263: that selector looks fragile, can we change something on the test page so that we can use something more solid like ID CSS selector here?

integration-test/background/click-to-load-youtube.js
```diff
@@ -0,0 +1,335 @@
...
+        const page = await browser.newPage()
+        await page.goto(testSite, { waitUntil: 'networkidle2' })
+
+        // Test the Iframe API controls and events function correctly, even when
+        // used with an existing video.
+        {
+            const waitForExpectedBorder = expectedBorder =>
+                page.waitForFunction(expectedBorder => (
+                    document.getElementById('existing-video')
+                        .style.border.split(' ').pop() === expectedBorder
```
kdzwinel, line 256: why border?

integration-test/background/click-to-load-youtube.js
```diff
@@ -0,0 +1,335 @@
...
+                    }
+                    return document.getElementById('spherical-video-roll')
+                        .value === expectedRoll
+                },
+                { polling: 10 }, expectedRoll, clickFlip)
+
+            await waitForExpectedRoll('')
+
+            const button = await page.evaluateHandle(
+                'document.querySelector("div:nth-child(8) > div")' +
```
kdzwinel, line 291: ditto re selector

integration-test/helpers/objectSchema.js
```diff
@@ -0,0 +1,47 @@
...
+        const descriptors = Object.getOwnPropertyDescriptors(targetObject)
+        for (const propertyName of Object.keys(descriptors).sort()) {
+            const descriptor = descriptors[propertyName]
+            result[propertyName] = Object.create(null)
+            for (const key of Object.keys(descriptor).sort()) {
+                const value = descriptor[key]
+                if (typeof value === 'function') {
+                    result[propertyName][key] = 'function(' + value.length + ')'
+                } else if (typeof value === 'object' && value !== null) {
+                    result[propertyName][key] = getObjectSchema(value)
```
kdzwinel, line 34: looped references are possible, should we add some kind of failsafe here? e.g. if depth > 10 => stop / error out

integration-test/helpers/objectSchema.js
```diff
@@ -0,0 +1,47 @@
...
+                } else {
+                    result[propertyName][key] = JSON.stringify(value)
```
kdzwinel, line 36: FYI stringify will also fail if there is a circular reference (try `JSON.stringify(window)`)

integration-test/helpers/requests.js
```diff
@@ -0,0 +1,72 @@
...
+ *   The Puppeteer page to log requests for.
+ * @param {array} requests
+ *   An array which request detals (@see <loggedRequestDetails>) for the page
+ *   will be appended.
+ *   Note: The requests array is mutated by this function.
+ */
+function logPageRequests (page, requests) {
+    const requestDetailsByRequestId = new Map()
+
+    page._client.on('Network.requestWillBeSent', ({
```
kdzwinel, line 20: isn't `_client` an interanal API? Is there a public getter?

integration-test/helpers/testConfig.js
```diff
@@ -0,0 +1,90 @@
+function parsePath (path) {
```
kdzwinel, line 1: oh boy, a parser? you'll have to help me understand what's going on :)

kdzwinel (commented):
OK, I did a first pass. Haven't reviewed youtube-iframe-api and click-to-load because those seem to be handled by a different PR, correct?

integration-test/background/click-to-load-youtube.js
kzar: This change already landed, but the last `.` is escaped because that's part of the property's name. In other words the Object we're referring to is `window['dbg']['tds']['tds']['trackers']['youtube.com']`.

integration-test/helpers/objectSchema.js
kzar: Ah, good point! Done.

integration-test/helpers/objectSchema.js
kzar: Ack, I think this is OK now.

integration-test/helpers/objectSchema.js
```diff
@@ -0,0 +1,65 @@
...
+    return await page.evaluate(function getObjectSchema (targetObject, previouslySeenObjects) {
+        const result = Object.create(null)
+        const descriptors = Object.getOwnPropertyDescriptors(targetObject)
+
+        // Keep track of Object values that we've already seen in order to avoid
+        // following cyclic Object values.
+        previouslySeenObjects = new Set(previouslySeenObjects)
+        previouslySeenObjects.add(targetObject)
+
+        const seenObjects = new Set(previouslySeenObjects)
```
kdzwinel, line 33: hm, I don't quite follow this code. why are we overriding previouslySeenObjects just to copy them right after?

BTW this code is getting sufficiently complex to have its own test set :) I don't want to derail you PR completely but have you considered using a library instead? https://www.npmjs.com/package/stringify-object

integration-test/data/api_schemas/facebook-sdk.json
```diff
@@ -0,0 +1,536 @@
+{
+    "FB": {
```
kdzwinel, line 2: this seems to include EVERYTHING, including properties that we don't  care about (e.g. click to load will still work fine). Can we be more precise and only report issues when something related to CTL changes? What I'm worried about is constant alerts from those tests whenever FB updates their library and some internal API changes (e.g. `__globalCallbacks`) or new props are added that we don't care about.

kdzwinel (commented):
I made some waves in comments, but to be clear, I think this is a good idea. I just worry that:

- those tests will be too noisy because they panic whenever anything in FB/YT library changes
- those tests will only run on PR while those libraries can change unexpectedly between PRs and break CTL
- getObjectSchema starts to get complex and might be potentially replaced by an existing lib

integration-test/data/api_schemas/facebook-sdk.json
kzar: My thinking is that it will be less work to tweak the JSON file occasionally to get the test passing, than it will be to attempt to debug breakage on website that our out of date surrogate scripts start breaking. Also personally I would like to know as a point of professional pride if the surrogate scripts I write are broken, or if they are just out of date.

Yea, I agree the included `FB` SDK seems quite large. I don't know which parts of it matter, I'm not sure that is knowable either. But I suppose a simple option might be to only test the YouTube SDK for now?

kzar: (Maybe also worth noting that @Charlie-belmer currently has a weekly recurring task in Asana to manually check the Facebook SDK for changes. Doing that manually seems like a PITA, when we could automate it like this instead.)

integration-test/helpers/objectSchema.js
kdzwinel: yeah, fair enough, I also avoid adding 3p libraries where possible.

One more question about the code - I understand the purpose you mentioned, but what I don't understand is making a copy of the set twice:

> previouslySeenObjects = new Set(previouslySeenObjects)
> const seenObjects = new Set(previouslySeenObjects)

integration-test/data/api_schemas/facebook-sdk.json
kdzwinel: Gotcha, yeah, I didn't realize we checked that manually. Automating it, even if it's not perfect, does seem better 👍

integration-test/helpers/objectSchema.js
kzar: Ah, good point! We don't need to do that, fixed. While at it, I added some more comprehensive comments.

kdzwinel (approved):
Nice work 👏 Let's merge this!

PR #1056, opened 2022-02-18
Don't create Tab objects for initial tabs until config is loaded.

sammacbeth:
**Reviewer:**

## Description:

Fixes an issue where we incorrectly showed the message "temporarily disabled Privacy Protection" for all tabs which are open on extension start.

## Automated tests:
- [ ] Unit tests
- [ ] Integration tests

###### Reviewer Checklist:

###### PR Author Checklist:

### DIFF
--- a/shared/js/background/events.es6.js
+++ b/shared/js/background/events.es6.js
@@ -489,15 +489,6 @@ browser.alarms.onAlarm.addListener(async alarmEvent => {
  * on start up
  */
 const onStartup = async () => {
-    const savedTabs = await browser.tabs.query({ currentWindow: true, status: 'complete' })
-    for (let i = 0; i < savedTabs.length; i++) {
-        const tab = savedTabs[i]
-
-        if (tab.url) {
-            tabManager.create(tab)
-        }
-    }
-
     await settings.ready()
     experiment.setActiveExperiment()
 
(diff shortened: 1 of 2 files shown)


kdzwinel:
Thanks for jumping on it Sam! Two things:

1. Can you explain why moving this code fixes the message?
2. I can see that the message is gone and I can see that stuff gets blocked on restored tabs, but popup doesn't seem to update with blocked content:

- load interia.pl
- don't approve the consent popup
- reload the extension
- approve the consent popup

what happens is that a bunch of trackers try to load and are blocked, but popup is showing 0
blocked

I think we can merge the message thing first and deal with that later if it will turn out to be a non-obvious to fix.

There is also one nit: extension icon doesn't update it shows Dax for me until I reload / navigate away. This might be related to 2, but might be separate. This is least important.

sammacbeth:
Thanks for the review Konrad.

1/ When we create a new `Site` object we get the enabled features for it based on the config:

https://github.com/duckduckgo/duckduckgo-privacy-extension/blob/2082d0841bc277794d63cf4e85ac90293e02e15d/shared/js/background/utils.es6.js#L192-L201

This requires that we've already loaded the tds config, otherwise the feature list is always empty. If the list is empty then we consider the site 'broken' in the UI and show the warning:

https://github.com/duckduckgo/duckduckgo-privacy-extension/blob/2082d0841bc277794d63cf4e85ac90293e02e15d/shared/js/ui/models/site.es6.js#L251

This change moves the tab initialization code after we've done `await tdsStorage.getLists()` which will load the config, so we are guaranteed to have features loaded when we build the enabled feature list.

2/ This looks like another bug with tabs created on browser start. I just pushed another commit to fix that.

Usually, a `Tab` object is created from a `main_frame` request. However, when we create `Tab`s on startup, we don't have that, so use a `chrome.tab` object instead. In our blocking code we were only adding blocking info when the tab has `statusCode = 200` though, which would never be the case for Tabs created from a `chrome.tab` object (because that object does not include a status code), so we would never add tracker info for tabs created in this way.

kdzwinel (approved):
LGTM, thanks so much Sam!

PR #1081, opened 2022-03-04
Add screenshot integration tests for Facebook Click to Load

kzar:
Let's add some integration tests for the Facebook Click to Load
feature that compare observed UI to expected screenshots. When those
tests fail,  store the offending screenshots as artifacts and log the
file paths.

**Reviewer:** @kdzwinel 
**CC:** @ladamski 

## Automated tests:
- [ ] Unit tests
- [x] Integration tests

###### Reviewer Checklist:

###### PR Author Checklist:


kzar:
Thanks for giving it a try @ladamski!

> Unfortunately the rendering didn't match exactly on MacOS.

Dang, thanks for confirming. It's good to know that this patch needs more work and can't be landed as-is. I don't have an Apple test device yet, but will figure something out and keep going with this when I get a chance.

> The artifacts directory shown below is empty so not sure how to recover those image diffs?

Weird! The files listed in the console output should be there, I'm not sure what happened. Again I'll have to keep working on this.

jonathanKingston:
- Are we able to add a % jitter perhaps?
- Alternatively we could exclude them or provide different data for macos?
- We care most about CI working right?

ladamski:
I was just running into this with the privacy dashboard tests, which use Jest's `toMatchImageSnapshot`.   I suggest we should use Jest or something similar for pixel matching tests, so then we can do something similar to https://github.com/more-duckduckgo-org/duckduckgo-privacy-dashboard/compare/la/update-e2e-thresholds

jonathanKingston:
@shakyShane is there another tool you've used in the past for screenshotting and handling differences like this?

shakyShane:
@jonathanKingston it's always used something chromium-based in the past.

an idea could be to have a higher threshold on platforms that require it. 

So for example, keep it at `0.1` for CI, but allow a higher number for `macOS`? You'd need to verify that it can still detect changes that you expect, but worth a shot

ladamski:
I think the issue right now is that there's 0% margin for error, since the comparison function is hand spun.  If we use the same approach as I linked above in the privacy dashboard (Jest), at least we would have some consistency there?

kzar:
@shakyShane Am I right in saying you've set up some screenshot tests for the extension now? I vaguely remember someone talking about that at the team meetup, but when I checked the extension's integration tests I couldn't find them. (Just trying to figure out if I should close this PR.)

shakyShane:
This popped into my feed just now, sorry for not seeing it originally! All I can say is that we have playwright now, which supports screenshots

PR #1092, opened 2022-03-15
AMP Protections

SlayterDev:
**Reviewer:**

## Description:

https://app.asana.com/0/72649045549333/1201724156469593/f

## Automated tests:
- [ ] Unit tests
- [ ] Integration tests

###### Reviewer Checklist:

###### PR Author Checklist:


shared/js/background/amp-protection.es6.js
```diff
@@ -0,0 +1,207 @@
...
+        !tdsStorage.config.features ||
+        !tdsStorage.config.features.ampLinks ||
+        !tdsStorage.config.features.ampLinks.settings ||
+        !tdsStorage.config.features.ampLinks.settings.linkFormats ||
+        !tdsStorage.config.features.ampLinks.settings.keywords) {
+        return false
+    }
+
+    regexPatterns = tdsStorage.config.features.ampLinks.settings.linkFormats
+    keywords = tdsStorage.config.features.ampLinks.settings.keywords
```
jonathanKingston, line 29: Can you just make this a single var?
```suggestion
    ampSettings = tdsStorage.config.features.ampLinks.settings
```

That way checking and unsetting will be a little less error prone.

shared/js/background/redirect.es6.js
```diff
@@ -72,7 +74,43 @@ function handleRequest (requestData) {
...
         }
 
-        const mainFrameRequestURL = new URL(requestData.url)
+        let mainFrameRequestURL = new URL(requestData.url)
+
+        // AMP protection
+        const canonUrl = ampProtection.extractAMPURL(thisTab.site, mainFrameRequestURL.href)
+        if (canonUrl) {
+            thisTab.setAmpUrl(mainFrameRequestURL.href)
+            thisTab.cleanAmpUrl = canonUrl
```
jonathanKingston, line 83: Mixing accessor function and assignment is a bit of a smell.

jonathanKingston (changes requested):
I think we should simplify the redirect method and also be handling timeouts from the AMP servers, leaving a user on about:blank even if an edge case should be avoided.

shared/js/background/amp-protection.es6.js
```diff
@@ -0,0 +1,212 @@
...
+    const canonicalLinks = doc.querySelectorAll('[rel="canonical"]')
+
+    if (canonicalLinks.length > 0) {
+        const newSite = new Site(canonicalLinks[0].href)
+
+        if (newSite.specialDomainName || !newSite.isFeatureEnabled(featureName)) {
+            return null
+        }
+
+        return canonicalLinks[0].href
```
jonathanKingston, line 162: Dropping 'All' from the method and renaming the variable reduces the overhead here a little:
```suggestion
    const firstCanonicalLink = doc.querySelector('[rel="canonical"]')

    if (firstCanonicalLink) {
        const newSite = new Site(firstCanonicalLink.href)

        if (newSite.specialDomainName || !newSite.isFeatureEnabled(featureName)) {
            return null
        }

        return firstCanonicalLink.href
```

shared/js/background/amp-protection.es6.js
```diff
@@ -0,0 +1,212 @@
...
+    if (!text) {
+        return null
+    }
+
+    const doc = new DOMParser().parseFromString(text, 'text/html')
+    const errorNode = doc.querySelector('parsererror')
+    if (errorNode) {
+        // DOMNParser failed to parse the document
+        return null
+    }
```
jonathanKingston, line 151: This appears to only ever be for XML, I tried hard to make it create an invalid doc and it doesn't. Instead you get a blank HTML file for invalid HTML.

For example: `const doc = new DOMParser().parseFromString('</d>', 'text/html')` just results in a blank doc.

I'd just drop this:
```suggestion
```

shared/js/background/amp-protection.es6.js
```diff
@@ -0,0 +1,212 @@
...
+    const currentTab = tabManager.get({ tabId: thisTab.id })
+    if (currentTab) {
+        // Set clean url to the canonical url or the original url if no canonical url is found
+        currentTab.cleanAmpUrl = canonUrl || url
+    }
+
+    if (canonUrl) {
+        return { redirectUrl: canonUrl }
+    }
+}
```
jonathanKingston, line 189: No longer used right? Drop this and the export.
```suggestion
```

shared/js/background/redirect.es6.js
```diff
@@ -36,6 +38,40 @@ function buildResponse (url, requestData, tab, isMainFrame) {
...
+    }
+}
+
+async function handleAmpDelayedUpdate (thisTab, url) {
+    const canonicalUrl = await ampProtection.fetchAMPURL(thisTab.site, url)
+    const currentTab = tabManager.get({ tabId: thisTab.id })
+    const newUrl = canonicalUrl || url
+    updateTabCleanAmpUrl(currentTab, canonicalUrl, url)
+
+    browser.tabs.update(thisTab.id, { url: newUrl })
```
jonathanKingston, line 62: I think in the `url` case when `canonicalUrl` is null we're going to make an infinite loop no?

jonathanKingston (changes requested):
Some more cleanup. I still need to look at the tests yet but everything is looking great.

I'm concerned about the comment about infinite looping... let's discuss over MM / Zoom :)

shared/js/background/amp-protection.es6.js
```diff
@@ -0,0 +1,183 @@
...
+        return null
+    }
+
+    if (site.specialDomainName || !site.isFeatureEnabled(featureName) || site.url === 'about:blank') {
+        return null
+    }
+
+    let data = null
+    const timeoutController = new AbortController()
+    setTimeout(() => timeoutController.abort(), 1500)
```
jonathanKingston, line 128: Can we make timeout constant this a config setting as it could be a source of breakage / need tweaking?

shared/js/background/amp-protection.es6.js
```diff
@@ -0,0 +1,183 @@
...
+/**
+ * Check if the given request is a suspected 1st party AMP URL. If it is, then we will attempt to fetch the canonical URL
+ *
+ * @param {object} requestData - request data
+ * @param {object} thisTab - current tab
+ * @param {object} mainFrameRequestURL - main frame request url
+ * @returns true if the request is a suspected 1st party AMP url
+ */
+function tabNeedsDeepExtraction (requestData, thisTab, mainFrameRequestURL) {
+    if (utils.getBrowserName() !== 'moz') {
```
jonathanKingston, line 74: Can you file a follow up to investigate if there's other solutions here and remove the dead code pretty swiftly if not?

jonathanKingston (changes requested):
Can we make the timeout configurable? Alternatively/Additionally disabling deep extraction.

shared/js/background/amp-protection.es6.js
SlayterDev: Follow up was planned as part of postmortem process 👍

jonathanKingston (approved):

PR #1141, opened 2022-04-22
Remove xray code for click to play

jonathanKingston:
**Reviewer:** @Charlie-belmer  or @kzar 

## Description:

This was fixed by https://bugzilla.mozilla.org/show_bug.cgi?id=1694679 so the regular code path works as intended.
https://app.asana.com/0/1176047645786601/1202155353194348/f

I've been testing on many sites and struggling to find the original breakage for this.

## Automated tests:
- [ ] Unit tests
- [ ] Integration tests

###### Reviewer Checklist:

###### PR Author Checklist:


integration-test/background/click-to-load-facebook.js
```diff
@@ -159,8 +159,7 @@ describe('Facebook SDK schema', () => {
         //       to match
         //       /integration-test/artifacts/api_schemas/facebook-sdk.json
         //       and make any corresponding changes required to the surrogate
-        //       scripts /shared/data/web_accessible_resources/facebook-sdk.js
-        //       and /shared/js/content-scripts/fb-surrogate-xray.js.
+        //       script /shared/data/web_accessible_resources/facebook-sdk.js
```
kzar, line 162: Nit: Missing full-stop.

kzar (commented):
Thanks, great patch! I'm always happy to remove code like this!

Only thing is, [that Firefox bug was fixed with Firefox 89](https://bugzilla.mozilla.org/show_bug.cgi?id=1694679), but we so far support Firefox >= 65.0. I think this change also needs to disable CTP entirely for Firefox < 89 (or drop support for Firefox < 89).

FWIW my vote would be to drop support for Firefox < 89, but that might take some time to get approved :/.

kzar:
@jonathanKingston Oh, one other thing. It would be nice to also remove the references to `fb-surrogate-xray.js` from `integration-test/data/configs/click-to-load-facebook.json` and `unit-test/data/clickToLoadConfig.json`.

jonathanKingston:
> Only thing is, [that Firefox bug was fixed with Firefox 89](https://bugzilla.mozilla.org/show_bug.cgi?id=1694679), but we so far support Firefox >= 65.0. I think this change also needs to disable CTP entirely for Firefox < 89 (or drop support for Firefox < 89).

Right but they will still work in some contexts. IIRC it's only when the tag is marked with "crossorigin" and even then I think we'd basically just be blocking rather than having a semi working experience.

> Oh, one other thing. It would be nice to also remove the references to fb-surrogate-xray.js from integration-test/data/configs/click-to-load-facebook.json and unit-test/data/clickToLoadConfig.json.

That was actually intended to ensure that the code keeps working with the downloaded format. I think we should clear this up once we land this and then remove it from the remote file. Happy to remove before though.

kzar:
> > Only thing is, [that Firefox bug was fixed with Firefox 89](https://bugzilla.mozilla.org/show_bug.cgi?id=1694679)...
> 
> Right but they will still work in some contexts. IIRC it's only when the tag is marked with "crossorigin" and even then I think we'd basically just be blocking rather than having a semi working experience.

Yea, I see what you mean. Fair enough, sounds good to me, sound OK to you @Charlie-belmer?
 
> > It would be nice to also remove the references to fb-surrogate-xray.js...
> 
> That was actually intended to ensure that the code keeps working with the downloaded format. I think we should clear this up once we land this and then remove it from the remote file. Happy to remove before though.

Ah right, got it. Hmm, I think my preference is still to remove those now.

kzar (approved):
LGTM :1st_place_medal:

jonathanKingston:
> Ah right, got it. Hmm, I think my preference is still to remove those now.
Raised https://app.asana.com/0/1201720254973470/1202164530144430 to track the removal from the config

Charlie-belmer (approved):
LGTM

PR #1229, opened 2022-06-06
Use `globalThis` instead of `window` in integration tests

kzar:
With Chrome Manifest V3, the extension's background page is replaced
with a background ServiceWorker that no longer has access to a
`window` Object. Let's replace (where possible) references to the
`window` Object (and implicit references to `window.dbg`) with
references to `globalThis`. While this isn't enough to get the
integration tests passing for Chrome MV3, it removes one more reason
for the tests to fail!

Note: Some references to `window` would have worked, since they are in
      a page instead of the background ServiceWorker. `self` would
      have worked too. Let's replace those anyway, to set a clear
      example about which global Object to use.

**CC:** @ladamski 

## Steps to test this PR:
1. Ensure integration tests are still passing.

## Automated tests:
- [ ] Unit tests
- [x] Integration tests

###### Reviewer Checklist:

###### PR Author Checklist:

### DIFF
--- a/integration-test/background/atb.js
+++ b/integration-test/background/atb.js
@@ -1,4 +1,3 @@
-/* global dbg:false */
 const harness = require('../helpers/harness')
 const backgroundWait = require('../helpers/backgroundWait')
 const pageWait = require('../helpers/pageWait')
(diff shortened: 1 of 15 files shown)


jonathanKingston:
Elsewhere we've migrated to globalThis why self here? The changes look fine otherwise. globalThis seems better to set as the default as it also works in nodeJS contexts too.

jonathanKingston (approved):

PR #1236, opened 2022-06-10
Properly handle requests initiated by ServiceWorkers

kzar:
Requests initiated by ServiceWorkers are given a tabId of -1, but we
still want to block tracking requests that they might make. So let's:
 - Stop ignoring requests with a tabId of -1.
 - Create an opaque Tab Object for those requests, but skip storing it
   for later lookup (since by definition we aren't sure which tab the
   request originated from).
 - Consider the site URL for such requests to be the request
   initiator, when available. Otherwise, the request will always be
   considered to have originated from its own URL and will therefore
   always be considered a first party request.

**Reviewer:** @kdzwinel 

## Steps to test this PR:
1. Check that ServiceWorker initiated requests can be blocked, e.g. by using the test page https://privacy-test-pages.glitch.me/privacy-protections/request-blocking/

## Automated tests:
- [ ] Unit tests
- [ ] Integration tests

###### Reviewer Checklist:

###### PR Author Checklist:

### DIFF
--- a/shared/js/background/classes/tab.es6.js
+++ b/shared/js/background/classes/tab.es6.js
@@ -48,7 +48,17 @@ class Tab {
         this.cleanAmpUrl = null
         this.requestId = tabData.requestId
         this.status = tabData.status
-        this.site = new Site(this.url)
+
+        // Consider the site URL to be the request initiator for requests fired
+        // inside opaque 'tabs' (e.g. via ServiceWorkers). Otherwise, consider
+        // the site URL to be the request URL.
+        const requestInitiator = tabData.initiator || tabData.originUrl
+        if (this.id === -1 && requestInitiator) {
+            this.site = new Site(requestInitiator)
+        } else {
+            this.site = new Site(this.url)
+        }
+
         this.httpsRedirects = new HttpsRedirects()
         this.statusCode = null // statusCode is set when headers are recieved in tabManager.js
         this.stopwatch = {
@@ -56,7 +66,13 @@ class Tab {
             end: null,
             completeMs: null
         }
-        this.resetBadgeIcon()
+
+        // Set the default extension icon for the new tab, assuming it's really
+        // a tab (e.g. not a ServiceWorker).
+        if (this.id !== -1) {
+            this.resetBadgeIcon()
+        }
+
         this.webResourceAccess = []
         this.surrogates = {}
     };
--- a/shared/js/background/tab-manager.es6.js
+++ b/shared/js/background/tab-manager.es6.js
@@ -8,6 +8,11 @@ class TabManager {
         this.tabContainer = {}
     };
 
+    _createInternal (tabData) {
+        const normalizedData = browserWrapper.normalizeTabData(tabData)
+        return new Tab(normalizedData)
+    }
+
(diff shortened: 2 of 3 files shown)


shared/js/background/tab-manager.es6.js
```diff
@@ -36,6 +40,12 @@ class TabManager {
      * get({tabId: ###});
      */
     get (tabData) {
+        // Opaque 'tab' (e.g. ServiceWorker). Create the Tab Object, but don't
+        // store it for later lookup.
+        if (tabData.tabId === -1) {
```
kdzwinel, line 45: So normal tabs go through "create" and then can be fetched via "get", but SW tabs don't go through "create"?

shared/js/background/classes/tab.es6.js
```diff
@@ -48,15 +48,31 @@ class Tab {
         this.cleanAmpUrl = null
         this.requestId = tabData.requestId
         this.status = tabData.status
-        this.site = new Site(this.url)
+
+        // Consider the site URL to be the request initiator for requests fired
+        // inside opaque 'tabs' (e.g. via ServiceWorkers). Otherwise, consider
+        // the site URL to be the request URL.
+        const requestInitiator = tabData.initiator || tabData.originUrl
```
kdzwinel, line 55: Internally service workers always have a clear origin AFAIK - is our method of determining it reliable? In what cases initiator or originUrl are empty?

sw

kdzwinel (commented):
I left some questions in code + I wonder what happens when there are multiple service workers at play - is same tabID messing things up (e.g. overwriting them)?

shared/js/background/classes/tab.es6.js
kzar: I don''t think it is safe to assume the initiator URL is available, because in my experience it isn't always available. For example, perhaps requests initiated by other places will have a tab ID of -1 but no initiator URL. I figure it's better not to throw an exception in those cases, or to convert `undefined` to a URL of `"undefined"` by accident.

shared/js/background/tab-manager.es6.js
kzar: Yea, to be honest I ran the code through a debugger to figure out what happens and what needs to happen. [At the start of handleRequest, we call tabManager.get()](https://github.com/duckduckgo/duckduckgo-privacy-extension/blob/e3ebc63/shared/js/background/redirect.es6.js#L89) to fetch the Tab Object for the request. For requests initiated by ServiceWorkers, there won't be a Tab Object and so we need to return a temporary opaque Tab Object if we want tracker blocking to work.

In other words there were three things stopping tracker blocking working for such requests:
1. We explicitly ignored requests with tab ID of -1
2. We had no Tab Object for the "tab" (because its not a tab)
3. When creating an temporary Tab Object, the site URL and tab URL was the same and so the request was always considered first-party

kzar:
> I wonder what happens when there are multiple service workers at play - is same tabID messing things up (e.g. overwriting them)?

Hopefully not, since we create a temporary Tab Object and don't save it for later lookup if the tab ID is -1.

sammacbeth (approved):

PR #1249, opened 2022-06-15
Include update timing information in TDS + HTTPS storage classes

kzar:
The TDS and HTTPS storage classes handle both storing and periodically
updating the extension configuration files. In order to know when an
update is required, the first step is to know both when a
configuration file was last updated and how frequently the
configuration file should be updated.

For that, let's:
- Move the update period into the respective classes.
- Take note of the lastUpdate time each time a configuration is
  fetched (even if there weren't any changes).

Note: It is suboptimal how we are saving the lastUpdate field in
      localStorage via the settings abstraction, instead of in
      IndexedDB along with the rest of the configuration data. We're
      doing it this way for two reasons.
      1. Firstly, we need to update the lastUpdate time even if there
         was no changes to the configuration. It seems wasteful to
         rewrite the entire configuration, when it hasn't changed,
         just to update the lastUpdate time.
      2. The entity tags for configurations are already stored in this
         way anyway, so it seems pragmatic to do the same with these
         changes.

**Reviewer:** @sammacbeth 
**CC:** @jonathanKingston 

## Steps to test this PR:
1. Hard to test, but you could try calling `chrome.alarms.getAll(console.log)` in the background console to ensure the alarms are being set with the correct period.
2. Can also call `dbg.settings.getSetting('config-lastUpdate')` in the background console to inspect the stored lastUpdate time.

## Automated tests:
- [ ] Unit tests
- [ ] Integration tests

###### Reviewer Checklist:

###### PR Author Checklist:

### DIFF
--- a/shared/js/background/storage/https.es6.js
+++ b/shared/js/background/storage/https.es6.js
@@ -9,6 +9,9 @@ class HTTPSStorage {
         this.dbc.version(1).stores({
             httpsStorage: 'name,type,data,checksum'
         })
+
+        // Update the lists every 12 hours.
+        this.updatePeriodInMinutes = 12 * 60
     }
 
     // Load https data defined in constants.httpsLists.
@@ -22,6 +25,20 @@ class HTTPSStorage {
             const etag = settings.getSetting(`${listCopy.name}-etag`) || ''
 
             return this.getDataXHR(listCopy.url, etag).then(response => {
+                // Set the lastUpdate time.
+                // Notes:
+                //  - Take the earliest time between server and local, that way if
+                //    the local time is set far in the future updates will still
+                //    happen.
+                //  - Date.parse() returns NaN for invalid (or missing) Date
+                //    headers, and Math.min() always considers NaN to be the
+                //    smallest value. So before calling Math.min(), replace
+                //    serverTime with localTime if serverTime is falsey (NaN).
+                const localTime = Date.now()
+                const serverTime = Date.parse(response.date)
+                const updateTime = Math.min(localTime, serverTime || localTime)
+                settings.updateSetting(`${listCopy.name}-lastUpdate`, updateTime)
+
                 // for 200 response we update etags
                 if (response && response.status === 200) {
                     const newEtag = response.etag || ''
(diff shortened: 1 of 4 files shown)


shared/js/background/storage/https.es6.js
```diff
@@ -22,6 +25,9 @@ class HTTPSStorage {
             const etag = settings.getSetting(`${listCopy.name}-etag`) || ''
 
             return this.getDataXHR(listCopy.url, etag).then(response => {
+                // Set the lastUpdate time.
+                settings.updateSetting(`${listCopy.name}-lastUpdate`, Date.now())
```
sammacbeth, line 29: Can we trust the system clock here? If the user has their clock set in the future, then resets it to the correct time, updates will be blocked until we reach the previous timestamp.

An alternative would be to use the server-sent timestamp in the `Date` header of the response.

sammacbeth (commented):
I don't see where we read this setting - will that be in a separate PR?

shared/js/background/storage/https.es6.js
kzar: I suppose that's true. If the user set their system clock far in the future, installed the extension, set their system clock back again, and then took care to never run the extension for more than the update period (30 mins for TDS, 12 hours for HTTPS) the extension configurations would never be updated.

That doesn't seem like an especially likely scenario though. I suppose it doesn't harm to attempt to parse the Date header, if available and then choose the lower out of Date.now() and the response header date. Will update that now.

kzar:
> I don't see where we read this setting - will that be in a separate PR?

Yea, that will be a separate PR, the `preferLocal` change.

shared/js/background/storage/https.es6.js
```diff
@@ -22,6 +25,15 @@ class HTTPSStorage {
             const etag = settings.getSetting(`${listCopy.name}-etag`) || ''
 
             return this.getDataXHR(listCopy.url, etag).then(response => {
+                // Set the lastUpdate time.
+                // Note: Take the earliest time between server and local, that way
+                //       if the local time is set far in the future updates will
+                //       still happen.
+                const localTime = Date.now()
+                const serverTime = Date.parse(response.date)
+                const updateTime = Math.min(localTime, serverTime || localTime)
```
sammacbeth, line 34: Can you mention here that the `|| localTime` is to remove `NaN` should the date now parse (as otherwise `Math.min` will pick `NaN`). This is a gotcha which I think should be pointed out to prevent a regression if future maintainers think they can refactor this.

sammacbeth (approved):
Looks good - can you just add the comment I mention, then I think this is good.

shared/js/background/storage/https.es6.js
kzar: Good idea, Done.

PR #1252, opened 2022-06-20
Mark integration tests as pending if page fails to load/reload

kzar:
We mark integration tests as pending (skipped) if the test page loads
too slowly and therefore times out. We should also mark tests as
pending when the page fails to load entirely.

Recently the fingerprinting protection tests were failing since
Wikipedia had network issues. It would have been nicer if they were
skipped instead.

**Reviewer:** @ladamski 

## Steps to test this PR:
1. Add a /etc/hosts entry of `0.0.0.0 privacy-test-pages.glitch.me`.
2. Run the integration tests.
3. Ensure that some tests are skipped with a suitable message. (Note that some tests will still fail, that is to be expected.)

## Automated tests:
- [ ] Unit tests
- [ ] Integration tests

###### Reviewer Checklist:

###### PR Author Checklist:

### DIFF
--- a/integration-test/helpers/pageWait.js
+++ b/integration-test/helpers/pageWait.js
@@ -36,7 +36,7 @@ async function forGoto (page, url) {
         if (e instanceof puppeteer.errors.TimeoutError) {
             pending('Timed out loading URL: ' + url)
         } else {
-            throw e
+            pending(`Failed to load URL: ${url} (${e.message}).`)
         }
     }
 }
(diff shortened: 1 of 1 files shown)


ladamski (commented):
This seems like it would swallow any page load failures, including permanent DNS or server changes.  Then we'd end up with tests silently failing indefinitely.  Can we be more specific about the types of failures we see here and just handle those?

PR #1269, opened 2022-07-05
Avoid using onBeforeNavigate timeStamp for onboarding banner

kzar:
When the extension is installed some onboarding messaging is displayed
the first time the user performs a DDG search. The onCommitted event
listener required for that is sometimes triggered twice in quick
succession by the browser. To avoid the onboarding messaging being
displayed twice when that happens, there were two workarounds:

1. Adding an onBeforeNavigate listener that kept track of the event's
   timestamp in a local variable. onCommitted events fired before the
   most recent recorded timestamp are ignored.
2. The showWelcomeBanner and showCounterMessaging settings are checked
   and cleared immediately when the event is fired, so that the second
   time it is fired the onboarding messaging isn't displayed.

The first approach isn't compatible with manifest V3, since state can
no longer be persisted between event listeners with local
variables[1]. Luckily however, the approach isn't necessary since
checking and clearing the settings is enough.

So let's remove the onBeforeNavigate listener and timer
variable. While at it, let's also stop listening for the onCommitted
event once the onboarding messaging has been displayed.

1 - https://developer.chrome.com/docs/extensions/mv3/migrating_to_service_workers/#state

**Reviewer:** @sballesteros 

## Steps to test this PR:
1. Install the extension (in Chrome) and perform a search from the navigation bar.
2. Ensure the onboarding banner is displayed.
3. Perform another search from the navigation bar.
4. Ensure the banner is no longer displayed.

- Ensure the integration tests still pass.

## Automated tests:
- [ ] Unit tests
- [ ] Integration tests

###### Reviewer Checklist:

###### PR Author Checklist:

### DIFF
--- a/shared/js/background/events.es6.js
+++ b/shared/js/background/events.es6.js
@@ -77,72 +77,80 @@ browser.runtime.onInstalled.addListener(onInstalled)
  * ONBOARDING
  * Logic to allow the SERP to display onboarding UI
  */
-
-let onBeforeNavigateTimeStamp = null
-browser.webNavigation.onBeforeNavigate.addListener(details => {
-    onBeforeNavigateTimeStamp = details.timeStamp
-})
-
-browser.webNavigation.onCommitted.addListener(async details => {
+async function onboardingMessaging ({ transitionQualifiers, tabId }) {
     await settings.ready()
     const showWelcomeBanner = settings.getSetting('showWelcomeBanner')
     const showCounterMessaging = settings.getSetting('showCounterMessaging')
 
-    // We show the welcome banner and counter messaging only once
-    if (showWelcomeBanner || showCounterMessaging) {
-        const isAddressBarQuery = details.transitionQualifiers.includes('from_address_bar')
-
-        if (showWelcomeBanner) {
-            settings.removeSetting('showWelcomeBanner')
-        }
-        if (isAddressBarQuery && showCounterMessaging) {
-            settings.removeSetting('showCounterMessaging')
-        }
-
-        if (onBeforeNavigateTimeStamp < details.timeStamp) {
-            if (browserName === 'chrome') {
-                browserWrapper.executeScript({
-                    target: { tabId: details.tabId },
-                    func: onboarding.onDocumentStart,
-                    args: [{
-                        duckDuckGoSerpHostname: constants.duckDuckGoSerpHostname
-                    }],
-                    injectImmediately: true
-                })
-            }
-
-            browserWrapper.executeScript({
-                target: { tabId: details.tabId },
-                func: onboarding.onDocumentEnd,
-                args: [{
-                    isAddressBarQuery,
-                    showWelcomeBanner,
-                    showCounterMessaging,
-                    browserName,
-                    duckDuckGoSerpHostname: constants.duckDuckGoSerpHostname,
-                    extensionId: browserWrapper.getExtensionId()
-                }],
-                injectImmediately: false
-            })
-        }
+    // If the onboarding messaging has already been displayed, there's no need
+    // to trigger this event listener any longer.
+    if (!showWelcomeBanner && !showCounterMessaging) {
+        browser.webNavigation.onCommitted.removeListener(onboardingMessaging)
+        return
     }
-}, {
-    // we only target the SERP (it has a `q` querystring param but not necessarily as the first querstring param)
-    url: [
-        {
-            schemes: ['https'],
-            hostEquals: constants.duckDuckGoSerpHostname,
-            pathEquals: '/',
-            queryContains: '?q='
-        },
-        {
-            schemes: ['https'],
-            hostEquals: constants.duckDuckGoSerpHostname,
-            pathEquals: '/',
-            queryContains: '&q='
-        }
-    ]
-})
+
+    // The counter messaging should only be active for the very first search
+    // navigation observed.
+    const isAddressBarQuery = transitionQualifiers.includes('from_address_bar')
+    if (isAddressBarQuery && showCounterMessaging) {
+        settings.removeSetting('showCounterMessaging')
+    }
+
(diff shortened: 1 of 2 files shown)


shared/js/background/events.es6.js
```diff
@@ -77,54 +77,61 @@ browser.runtime.onInstalled.addListener(onInstalled)
...
+        if (!tabActive ||
+            tabUrl.hostname !== constants.duckDuckGoSerpHostname ||
+            tabUrl.pathname !== '/' ||
+            !tabUrl.searchParams.get('q')) {
+            return
+        }
 
+        if (showWelcomeBanner) {
+            settings.removeSetting('showWelcomeBanner')
+        }
```
sballesteros, line 109: For some reasons, it seems that the `onCommitted` listener can be called twice.
As a result, in case of the counter-messaging, not removing the setting before the early return can cause the banner to be displayed twice.
Testing the counter-messaging is a bit annoying but can be done with the following steps:

- On the SERP (in `src/js/pages.SERP.js`), modify  `window.onFirstSearchPostExtensionInstall` so that the counter-messaging shows on your testing device. For instance, if you are on a Mac you can do:
    ```diff
       (
    -     (env.device.isOSX && env.device.chromeVersion >= 87 && env.device.chromeVersion <= 89) ||
    +     (env.device.isOSX && env.device.chromeVersion >= 87 /*&& env.device.chromeVersion <= 89*/) ||
          (env.device.isWindows && env.device.chromeVersion >= 87)
        )
    ```
- Rebuild the SERP (`npm i` `grunt` `r index`)
- Edit the extension code so that a search from the address point will hit your SERP dev instance e.g.
    - In constants.js set `duckDuckGoSerpHostname` to your dev instance
    - In manifest.json set `search_url` and `suggest_url` to your dev instance
    - Rebuild the extension
- Open Chrome
- Visit `https://<you>.duckduckgo.com`
- Clear the data (cache, cookies, local storage etc.)
- Uninstall the extension
- Quit Chrome (this is required so that the chrome message shows)
- Open Chrome
- Visit `https://<you>.duckduckgo.com` (this is required to handle auth)
- In another tab, visit [chrome://extensions](chrome://extensions)
- Install the patched version of the extension
- Due a search from the address bar (be sure that it's a search and not access to your history)
- [ ] Check that you see chrome message
- [ ] Check that you see the counter-messaging
- Click on "Keep it"
- [ ] 🐛  Check that you see 2 welcome messages displayed (this can be fixed by removing the `showWelcomeBanner` setting to before the early return, or, understanding why the callback is called twice).

sballesteros (changes requested):

shared/js/background/events.es6.js
sballesteros: Looking more, it seems that the callback is called twice because the URL filter contains 2 entries (so it seems that Chrome register as many event listeners as there are entries in the URL filter option). I didn't investigate in details but this [old comment on a bug report](https://bugs.chromium.org/p/chromium/issues/detail?id=450068#c6) seems to point at smtg similar.

kzar: > in case of the counter-messaging, not removing the setting before the early return can cause the banner to be displayed twice.

With this change we're removing the welcome banner setting after the early return, but still removing the counter-messaging setting every time. I don't think this change should affect the counter-messaging banner.

sballesteros: @kzar apologies if i wasn't clear, If you follow the testing instructions above you will see 2 welcome banners rendered _on the same page_  after clicking on "keep it".
This is a regression introduced by this change.

sballesteros: @kzar here is a screenshot of the issue (notice the 2 welcome banners)

Screen Shot 2022-07-06 at 9 27 30 AM

I can confirm that the fact that the callback is called twice is the root cause (and related to the URL filter).
I think that the issue is not the early return per se but the fact that you have an async operation before (`await browser.tabs.get(details.tabId)`).
The way I understand it, both callbacks are called on the same tick, so when they are called they both see the value for `showWelcomeBanner` to be `true` hence the bug.

kzar: Right OK, got it... I think!

I also can see our `onComitted` event listener firing twice like you said. I don't think it's due to the extra URL filter however, since I did a test with only one filter and it still fired twice. I suspect it's due to the request being redirected (I see the `TransitionQualifier` contains "server_redirect"), but that's just a guess.

The way I figure it, this `onBeforeNavigateTimeStamp` workaround is therefore serving to avoid us proceeding for both calls of our `onComitted` event listener. The first time it's called, the timestamp will have also been updated and so the function will return. The second time its called, the function will proceed since the timestamp won't have changed again.

The thing is, I think the logic to check and clear the `showWelcomeBanner` and `showCounterMessaging` should also have the same "debouncing" effect. The first time the function is called, the settings will be true and are immediately cleared. The second time the function is called, the settings will have been cleared and the function will return. To check this, I've done a test where the `onComitted` listener logs the value of the settings, and I see they are `undefined` on the second call.

I also think it's suboptimal to continue listening for the `onComitted` event after the onboarding messaging is complete. So I have updated my changes here to:
1. Stop listening for `onComitted` after the onboarding is complete. I took the conservative (and simpler to reason about) approach of waiting until the next time the event listener fires.
2. Remove the `onBeforeNavigateTimeStamp` logic.
3. Restore the `settings.removeSetting(showWelcomeBanner)` logic to happen consistently.

This all seems to be working for me, testings are passing and I see the welcome banner. That said, any chance you could perform another test of the challenge flow like you did above, to confirm the banner is not displayed twice? That would save me setting up the whole SERP dev environment which I'm not familiar with.

sballesteros (approved):
Thanks @kzar. Really nice refactoring 👍 
I re-tested the counter-messaging flow and I think that we are good to go.

PR #1276, opened 2022-07-06
Reland "Properly handle requests initiated by ServiceWorkers"

kzar:
Requests initiated by ServiceWorkers are given a tabId of -1, but we
still want to block tracking requests that they might make. So let's:
 - Stop ignoring requests with a tabId of -1.
 - Create an opaque Tab Object for those requests, but skip storing it
   for later lookup (since by definition we aren't sure which tab the
   request originated from).
 - Consider the site URL for such requests to be the request
   initiator, when available. Otherwise, the request will always be
   considered to have originated from its own URL and will therefore
   always be considered a first party request.
 - Disable 3p cookie protections for such requests for now, since that
   led to breakage (e.g. prevented Twitter users from logging out of
   their accounts). That breakage led to this change being reverted
   originally.
 - Add a unit test to ensure that 3p cookie protections are disabled
   for such requests, but not otherwise.
 - Add an integration test that verifies all tracking requests on our
   request blocking page[1] are blocked as expected.

1 - https://privacy-test-pages.glitch.me/privacy-protections/request-blocking/

**Reviewer:** @sammacbeth 
**CC:** @SlayterDev, @jonathanKingston 

## Steps to test this PR:
1. Check that ServiceWorker initiated requests can be blocked, e.g. by using the test page https://privacy-test-pages.glitch.me/privacy-protections/request-blocking/
2. Check that you can still log out of Twitter, go to https://twitter.com/ , log in, click to log out, confirm you are logged out.

## Automated tests:
- [ ] Unit tests
- [ ] Integration tests

###### Reviewer Checklist:

###### PR Author Checklist:

### DIFF
--- a/integration-test/background/request-blocking.js
+++ b/integration-test/background/request-blocking.js
@@ -1,0 +1,61 @@
+const harness = require('../helpers/harness')
+const { logPageRequests } = require('../helpers/requests')
+const backgroundWait = require('../helpers/backgroundWait')
+const pageWait = require('../helpers/pageWait')
+
+const testSite = 'https://privacy-test-pages.glitch.me/privacy-protections/request-blocking/'
+
+let browser
+let bgPage
+let teardown
+
+describe('Test request blocking', () => {
+    beforeAll(async () => {
+        ({ browser, bgPage, teardown } = await harness.setup())
+        await backgroundWait.forAllConfiguration(bgPage)
+    })
+
+    afterAll(async () => {
+        await teardown()
+    })
+
+    it('Should block all the test tracking requests', async () => {
+        // Load the test page.
+        const page = await browser.newPage()
+        await pageWait.forGoto(page, testSite)
+
+        // Start logging network requests.
+        const pageRequests = []
+        logPageRequests(page, pageRequests)
+
+        // Initiate the test requests and wait until they have all completed.
+        // Note: Waiting for network idle here does not work for some reason.
+        await page.click('#start')
+        await page.waitForFunction(
+            // eslint-disable-next-line no-undef
+            () => results.results.length >= tests.length
+        )
+
+        // Verify that no requests were allowed.
+        for (const { url, method, type, status } of pageRequests) {
+            const { hostname } = new URL(url)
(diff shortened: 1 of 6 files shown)


integration-test/background/request-blocking.js
```diff
@@ -0,0 +1,59 @@
...
+
+        // Start logging network requests.
+        const pageRequests = []
+        logPageRequests(page, pageRequests)
+
+        // Initiate the test requests.
+        // Note: Waiting for network idle here does not work for some reason, so
+        //       wait for ten seconds instead.
+        await page.click('#start')
+        await backgroundWait.forTimeout(bgPage, 10000)
```
sammacbeth, line 35: With the above change, you should be able to remove this.

If you need to wait for the test to finish, you should be able to check for the `window.results` object.

kzar: The `results` Object is not a property of `window` unfortunately (so we need the ESLint comment to turn off the undefined variable warning).

Also, I think the test page pushes results onto the `results.results` Array as they come in. Unless I'm missing something, we'd either need to hard-code the number of expected results in the integration tests, or we still need the timeout here. What do you think?

integration-test/background/request-blocking.js
```diff
@@ -0,0 +1,59 @@
+const harness = require('../helpers/harness')
+const { logPageRequests } = require('../helpers/requests')
+const backgroundWait = require('../helpers/backgroundWait')
+const pageWait = require('../helpers/pageWait')
+
+const testSite = 'https://privacy-test-pages.glitch.me/privacy-protections/request-blocking/'
```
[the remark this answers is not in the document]
kzar: I kind of prefer it as is, to keep loading the page and starting the test requests separate. But if you feel strongly I don't mind changing it, what do you think?

integration-test/background/request-blocking.js
kzar: Ah, scratch that. I just realised we can wait for `results.results.length` to match `tests.length` instead. I've got rid of the annoying ten second wait now.

sammacbeth (approved):

PR #1295, opened 2022-07-25
Update privacy grade and add typings

jonathanKingston:
**Reviewer:**

## Description:

Updates to the latest privacy-grade pr: https://github.com/duckduckgo/privacy-grade/pull/43

## Automated tests:
- [ ] Unit tests
- [ ] Integration tests

###### Reviewer Checklist:

###### PR Author Checklist:


shared/js/background/classes/tracker.es6.js
```diff
@@ -1,42 +0,0 @@
...
-            this.urls[t.fullTrackerDomain] = {
-                isBlocked: this.isBlocked(t.action),
-                reason: t.reason,
-                categories: t.tracker.categories
-            }
-        }
-    }
-
-    isBlocked (action) {
-        return !!action.match(/block|redirect/)
```
shakyShane, line 38: worth updating this to the same kind of match seen elsewhere, like

```javascript
['block', 'redirect'].includes(t.action)
```

package.json
```diff
@@ -75,7 +75,7 @@
         "@duckduckgo/autofill": "github:duckduckgo/duckduckgo-autofill#4.7.0",
         "@duckduckgo/content-scope-scripts": "github:duckduckgo/content-scope-scripts#2.1.1",
         "@duckduckgo/jsbloom": "^1.0.2",
-        "@duckduckgo/privacy-grade": "1.1.0",
+        "@duckduckgo/privacy-grade": "github:duckduckgo/privacy-grade#jkt/support-unknown-trackers",
```
shakyShane, line 78: is this the correct branch still?

shakyShane (approved):
Nothing blocking, just added a couple of questions :)

package.json
jonathanKingston: No, we'd need to update it once the other PR merges :)

shared/js/background/classes/tracker.es6.js
jonathanKingston: Yup this whole file is deleted :)

I replaced tracker.es6.js with tracker.js which has the updated check.

shakyShane: oh yeah! oops!

englehardt:
Doing some manual testing of this change with the new privacy grade changes. I'll post comments with any issues I find.

Tracker shown as blocked, when it isn't. It's a default:block tracker with a per-site exemption for cnn.com (among others).

Screen Shot 2022-07-25 at 10 30 50 PM

Screen Shot 2022-07-25 at 10 32 21 PM

englehardt:
Pulled the latest changes and that seemed to fix the amazon issue. Found a few more issues of different types on 
https://www.vice.com/en.

1/ default:ignore rules showing up in the panel

Screen Shot 2022-07-25 at 11 16 07 PM
Screen Shot 2022-07-25 at 11 15 04 PM

Screen Shot 2022-07-25 at 11 11 25 PM
Screen Shot 2022-07-25 at 11 11 36 PM

2/ default:block rule that has an action:ignore rule associated with the loaded path is blocked

Screen Shot 2022-07-25 at 11 16 51 PM
Screen Shot 2022-07-25 at 11 13 44 PM

Given the above it seems like we should set up a test page that loads resources that meet all test conditions:
* default:block
* default: ignore
* default: block which includes an action:ignore
* default: ignore which includes an action:block
* first-party CNAME --> to a tracker in each of the conditions above
* same-entity tracker loaded

Is the above enough, or do we need to further drill down into the scenarios? I.e., action:ignore/action:block but with exceptions or with type filters. I'm not sure if these would produce a different classification that the UI / blocking logic could get wrong. If so, we should add conditions for that too.

Testing on real sites is time consuming, since I'm diffing panels on the old and new extension by eye. We could use real paths on the test page to save time for now.

jonathanKingston:
> Given the above it seems like we should set up a test page that loads resources that meet all test conditions:

I think we 100% need this, I will fix the issues first and then start on this. I'm not sure right now how time consuming it will be. 

I am however convinced this logic issue is actually constrained just to the panel (and specifically the filters I added to just ignore action ignore+none for now) and not edge cases in the blocking logic you're mentioning which is obviously more concerning. We do have reference tests for those and unit tests so I'm fairly confident that hasn't changed.

In the debugger you can check out the tracker classification like this:

image

I'll investigate the nptech.com issue, it's different in that it has a NS_ERROR_INTERCEPTION_FAILED status code and the action should be ignore according to the debugger.

englehardt (approved):
I verified that the panel is showing what's expected on https://senglehardt.com/test/ddg/all_tracker_types.html and on a few test sites. I also reviewed the newest from Shane's earlier review.

PR #1309, opened 2022-08-04
Improve devtools visual elements

GuiltyDolphin:

**CC:** @kzar 

## Description:

https://app.asana.com/0/312629933896096/1202714468322458/f

This adds some visual improvements to the devtools panel, to make working with it easier:

- on/off visibility
  - disabled protection buttons now highlight in red, enabled in green
- current selection visibility
  - when hovering over a table cell, the cell and row will be highlighted for extra visibility
  - the contents of the hovered cell will expand if necessary (i.e., if the contents are cut short) to make the full contents visible
  - hovering over protection buttons will also change their colour slightly to indicate the hover
- fixes colouring on `set-cookie-tracker` rows (were not highlighting before)
- updates the red colour for "blocked" to be less harsh, easier on the eyes
- adds follow scrolling for the protections panel and table header
  - previously, you would have to scroll all the way to the top of the document to access the protection buttons, but now they will be accessible wherever you scroll

## Steps to test this PR:

1. Navigate to a web page
2. Open dev panel and set the target to the webpage. Reload the page
3. Check scrolling
    1. Scroll - the button panel should follow the scrolling
    2. Resize the window to be smaller, scoll
    3. The table header should remain at the bottom of the protections panel, and should scroll with you
5. Check that set-cookie-tracker rows have a blue/purple colouring
6. Check that hovering over cells highlights them and shows a border around the row
7. Check that blocked rows have a pale red colouring
8. Check feature buttons
    1. Check that enabled features are in green, disabled in red
    10. Toggle a feature, click "Reload page" at the top of the panel
    11. The feature colour should match the new protection status
1. Check cell content display
    1. Resize window to be small until cells show "..." at the end
    2. Hover over these cells
    3. Full cell contents should be visible
    4. Move cursor away, cell should should the "..." again

## Automated tests:
- [ ] Unit tests
- [ ] Integration tests

###### Reviewer Checklist:

###### PR Author Checklist:

### DIFF
--- a/shared/html/devtools-panel.html
+++ b/shared/html/devtools-panel.html
@@ -107,8 +107,7 @@
     td:hover {
       white-space: break-spaces;
       background-color: beige;
-      /* If the element is long enough to trigger text-overflow, but not long enough to wrap, this ensures we still show the content. */
-      text-overflow: unset;
+      word-wrap: break-word;
     }
 
     tr > td:nth-child(4) {
--- a/shared/html/devtools-panel.html
+++ b/shared/html/devtools-panel.html
@@ -92,6 +92,12 @@
       overflow: hidden;
     }
 
+    /* Make it easier to view contents of currently-hovered cell. */
+    td:hover {
+      white-space: break-spaces;
+      background-color: beige;
+    }
+
     tr > td:nth-child(4) {
       word-wrap: anywhere;
       overflow: auto;
--- a/shared/html/devtools-panel.html
+++ b/shared/html/devtools-panel.html
@@ -96,6 +96,8 @@
     td:hover {
       white-space: break-spaces;
       background-color: beige;
+      /* If the element is long enough to trigger text-overflow, but not long enough to wrap, this ensures we still show the content. */
+      text-overflow: unset;
     }
 
     tr > td:nth-child(4) {
--- a/shared/html/devtools-panel.html
+++ b/shared/html/devtools-panel.html
@@ -28,6 +28,12 @@
       padding: 10px;
     }
 
+    #settings-panel {
+      position: sticky;
+      top: 0;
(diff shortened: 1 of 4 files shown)


shared/html/devtools-panel.html
```diff
@@ -36,8 +42,24 @@
       margin: 2px;
     }
 
+    .protection-button-on {
```
jonathanKingston, line 45: image

I think we should make the borders consistent here, so specifying a radius and border.

shared/js/devtools/panel.es6.js
```diff
@@ -260,3 +274,12 @@ displayFilters.forEach((input) => {
         document.querySelectorAll('tr').forEach(setRowVisible)
     })
 })
+
+/**
+ * Observes the dev settings for resizing to ensure the table head sticks correctly to the bottom of the settings.
+ */
+const settingsResizeObserver = new ResizeObserver(function (entries) {
```
jonathanKingston, line 281: As mentioned on MM I think we can avoid this with using a scrollable tbody or similar solutions that are pure CSS.

shared/html/devtools-panel.html
```diff
@@ -92,6 +118,17 @@
...
 
+    /* Make it easier to view contents of currently-hovered cell. */
+    td:hover {
+      white-space: break-spaces;
+      background-color: beige;
+      word-wrap: break-word;
+    }
+
+    tr:hover {
+      border: 1px solid black;
```
jonathanKingston, line 129: This makes each line jump by one px even if it's not expanding in size. I think we could instead just change the shade of the tr background on highlight.

The alternative I think requires using transparent borders but I played around and forgot how to make this work :')

sammacbeth (approved):

PR #1310, opened 2022-08-05
Collapse duplicate request rows in devtools panel

GuiltyDolphin:

## Description:

When using the devpanel, there would often be duplicate requests which fill up the panel. This PR changes the behaviour to instead increment a counter next to the original request whenever duplicates are generated adjacent to one another.

## Steps to test this PR:

1. Open privacy-test-pages.glitch.me
2. Open the devtools panel, set the target to be the previously opened page
3. Open http://privacy-test-pages.glitch.me/tracker-reporting/1major-via-script.html
4. Request should be observed in dev panel
5. Action column should show `🚫 block (default block)`
6. Click "Home" (http://privacy-test-pages.glitch.me/index.html) then navigate back to http://privacy-test-pages.glitch.me/tracker-reporting/1major-via-script.html
7. Extra request should **not** be observed in devpanel
8. Action column should show `🚫 block (default block) [2]`
9. Navigate to http://privacy-test-pages.glitch.me/tracker-reporting/1major-via-img.html
10. Separate request should be observed as a new row

## Automated tests:
- [ ] Unit tests
- [ ] Integration tests

###### Reviewer Checklist:

###### PR Author Checklist:


shared/js/devtools/panel.es6.js
```diff
@@ -102,31 +139,31 @@ const actionHandlers = {
             cleanUrl.search = ''
             cleanUrl.hash = ''
             cells[1].textContent = cleanUrl.href
-            cells[2].textContent = `🍪 ${action}`
+            cells[2].querySelector('.request-action').textContent = `🍪 ${action}`
             cells[3].textContent = kind
             cells[4].textContent = type
             row.classList.add(kind)
-            table.appendChild(row)
+            return row
```
jonathanKingston, line 146: I think it would be simpler to follow calling addRequestRow here are there any downsides to that? The type signatures would be a little less complex.

```suggestion
            return addRequestRow(row)
```

shared/js/devtools/panel.es6.js
```diff
@@ -6,10 +6,46 @@ const tabPicker = document.getElementById('tab-picker')
...
+ */
+function addRequestRow (f) {
+    return (m) => {
+        const row = f(m)
+        if (row) {
+            // if duplicate request lines would be printed, we instead show a counter increment
+            const prevRow = document.querySelector('tbody > tr:last-child')
+            if (prevRow) {
+                const prevRowCopy = assertTableRowElement(prevRow.cloneNode(true))
+                prevRowCopy.querySelector('.action-count').textContent = ''
```
jonathanKingston, line 34: This handling is a little fragile if the table structure changes this will stop working.

So either:

- Add tests
- Run the table rendering from a data storage that can be interrogated for the previous row.

I think as we're only rolling up by the previous row; simply store row data in a _lastRow_ object and compare their fields.

jonathanKingston (changes requested):

shared/js/devtools/panel.es6.js
GuiltyDolphin: @jonathanKingston I've moved the calls to `addRequestRow` to the end of each segment

shared/js/devtools/panel.es6.js
GuiltyDolphin: @jonathanKingston I've added some helpers, w/ the row linked as an object

PR #1343, opened 2022-08-17
Remove parameterName and parameterValue from ad-attribution feature to simplify code

jonathanKingston:
**Reviewer:** @englehardt 

## Description:

Marking draft as CI should fail if working correcly; testCases still need amending.

This is to simplify the parsing logic as we already verify an ad click based on the adDomainParameterName parameter name in the link click.

https://app.asana.com/0/0/1202808108652811/f

Corresponding config change should probably land first so we can just run bundle-config instead of hand hacking it: https://github.com/duckduckgo/privacy-configuration/pull/345

## Automated tests:
- [ ] Unit tests
- [ ] Integration tests

###### Reviewer Checklist:

###### PR Author Checklist:

### DIFF
--- a/shared/data/bundled/extension-config.json
+++ b/shared/data/bundled/extension-config.json
@@ -1,6 +1,6 @@
 {
     "readme": "https://github.com/duckduckgo/privacy-configuration",
-    "version": 1661422548137,
+    "version": 1663586301491,
     "features": {
         "adClickAttribution": {
             "readme": "https://help.duckduckgo.com/duckduckgo-help-pages/privacy/web-tracking-protections/#3rd-party-tracker-loading-protection",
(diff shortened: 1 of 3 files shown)


englehardt:
@jonathanKingston let's rebase this on the newest develop and on https://github.com/duckduckgo/duckduckgo-privacy-extension/pull/1410, which fixes one regression I noticed when testing this.

I also saw another regression present in `develop` related to bat.js, though it isn't caused by these changes. It looks like some changes to the UI have caused the exempt domain warning to no longer show up. Here's what I see when an exemption is active:

Screen Shot 2022-09-19 at 3 33 11 PM

Here's what it looks like when inactive (which is correct):

Screen Shot 2022-09-19 at 3 33 47 PM

So the UI partially detecting the change as the description and icon are correct, but the exemption isn't listed.

englehardt (approved):
The UI issues I pointed out were related to an outdated local instance of privacy grade. I can no longer repro after updating.

Just retested and it LGTM.

PR #1358, opened 2022-08-20
Update puppeteer

jonathanKingston:
**Reviewer:** @shakyShane, @kzar, @GuiltyDolphin or @sammacbeth 

## Description:

To update https://github.com/duckduckgo/duckduckgo-privacy-extension/pull/1338 we needed to migrate from using the now removed _client API.

In doing this I was struggling to figure out the control flow of some of these files, so I pointed TS checking at them which revealed a bunch of wrong types and control flows that were unspecified.

Re-enabled some new tab checks for attribution that we decided to skip as were failing in ci. https://app.asana.com/0/312629933896096/1202736717601997/f

## Automated tests:
- [ ] Unit tests
- [ ] Integration tests

###### Reviewer Checklist:

###### PR Author Checklist:

### DIFF
--- a/integration-test/helpers/harness.js
+++ b/integration-test/helpers/harness.js
@@ -48,14 +48,18 @@ const setup = async (ops) => {
     if (loadExtension) {
         // Grab a handle on the background page for the extension.
         try {
-            const backgroundPageTarget =
-                  await browser.waitForTarget(
-                      target => (
-                          target.type() === 'background_page' ||
-                          target.type() === 'service_worker'
-                      ),
-                      { timeout: 2000 }
-                  )
+            const targetFilter = target => (
+                target.type() === 'background_page' ||
+                target.type() === 'service_worker'
+            )
+            let backgroundPageTarget = await browser.targets().find(targetFilter)
+            if (!backgroundPageTarget) {
+                backgroundPageTarget =
+                    await browser.waitForTarget(
+                        targetFilter,
+                        { timeout: 2000 }
+                    )
+            }
             bgPage = backgroundPageTarget.type() === 'background_page'
                 ? await backgroundPageTarget.page()
                 : await backgroundPageTarget.worker()
--- a/integration-test/helpers/harness.js
+++ b/integration-test/helpers/harness.js
@@ -48,18 +48,13 @@ const setup = async (ops) => {
     if (loadExtension) {
         // Grab a handle on the background page for the extension.
         try {
-            const targetFilter = target => (
-                target.type() === 'background_page' ||
-                target.type() === 'service_worker'
+            const backgroundPageTarget = await browser.waitForTarget(
+                target => (
+                    target.type() === 'background_page' ||
+                    target.type() === 'service_worker'
+                ),
+                { timeout: 2000 }
             )
-            let backgroundPageTarget = await browser.targets().find(targetFilter)
-            if (!backgroundPageTarget) {
-                backgroundPageTarget =
-                    await browser.waitForTarget(
-                        targetFilter,
-                        { timeout: 2000 }
-                    )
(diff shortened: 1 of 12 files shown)


jonathanKingston:
Nice our MV3 test caught https://github.com/puppeteer/puppeteer/pull/8807/files bumping to update to 16.2.0 :)

integration-test/helpers/harness.js
```diff
@@ -17,40 +17,29 @@ const setup = async (ops) => {
...
+        args.push('--disable-extensions-except=' + extensionPath)
+        args.push('--load-extension=' + extensionPath)
     }
 
     // github actions
     if (process.env.CI) {
-        puppeteerOps.args.push('--no-sandbox')
-    }
-
-    if (process.env.TRAVIS) {
```
jonathanKingston, line 41: Deleted this travis code because we don't use it and was invalid to the latest puppeteer anyway.

integration-test/helpers/harness.js
```diff
@@ -59,21 +48,29 @@ const setup = async (ops) => {
...
-                          target.type() === 'background_page' ||
-                          target.type() === 'service_worker'
-                      ),
-                      { timeout: 2000 }
-                  )
+            const targetFilter = target => (
+                target.type() === 'background_page' ||
+                target.type() === 'service_worker'
+            )
+            let backgroundPageTarget = await browser.targets().find(targetFilter)
```
kzar, line 55: I don't think it's necessary to do `browser.targets().find(targetFilter)` first, we can just do `await browser.waitForTarget` initially, right?

jonathanKingston: It's technically a race condition with their API I'm happy to remove but I don't think it harms being there either.

kzar: I don't think it is necessary or a race condition. If the target already exists, I think `waitForTarget` should return it right away. See [the example in the Puppeteer docs](https://github.com/puppeteer/puppeteer/blob/main/docs/index.md#working-with-chrome-extensions) and [the docs for waitForTarget()](https://github.com/puppeteer/puppeteer/blob/main/docs/api/puppeteer.browser.waitfortarget.md).

jonathanKingston: We're calling `waitForTarget` after the page has loaded, thus the service_worker or background_page might have been triggered in that time prior to us setting up the listener.
By definition it will race; unless they do something magic under the hood like pooling targets for a period of time for a listener to emerge... which seems doubtful.

I'm aware the docs don't have this but also they broke it themselves recently and didn't have tests for that.

kzar: My understanding is that `waitForTarget` already checks for existing targets that match the predicate, before waiting for new ones to be created. See https://github.com/puppeteer/puppeteer/blob/bdcb748b98b621510de72f30b24f8c47fee0ad11/src/common/Browser.ts#L659

jonathanKingston: Ah that seems non obvious :') thanks!

kzar (approved):
LGTM with one optional comment

integration-test/helpers/harness.js
```diff
@@ -59,21 +48,24 @@ const setup = async (ops) => {
...
         try {
-            const backgroundPageTarget =
-                  await browser.waitForTarget(
-                      target => (
-                          target.type() === 'background_page' ||
-                          target.type() === 'service_worker'
-                      ),
-                      { timeout: 2000 }
-                  )
+            const backgroundPageTarget = await browser.waitForTarget(
```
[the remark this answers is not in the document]
jonathanKingston: It matches our code style more, I'm going to drop this.

PR #1385, opened 2022-09-02
Correct some type checking issues with referrer trimming

kzar:
The argumentsObject and referrer trimming code were failing TypeScript
linting, so linting was disabled for those files. Let's fix the
problems and enable TypeScript linting again for them.

## Automated tests:
- [ ] Unit tests
- [ ] Integration tests

###### Reviewer Checklist:

###### PR Author Checklist:

### DIFF
--- a/shared/js/background/classes/tab.es6.js
+++ b/shared/js/background/classes/tab.es6.js
@@ -57,6 +57,9 @@ class Tab {
 
         /** @type {null | import('./ad-click-attribution-policy').AdClick} */
         this.adClick = null
+
+        /** @type {null | import('../events/referrer-trimming').Referrer} */
+        this.referrer = null
     }
 
     /**
--- a/shared/js/background/events/referrer-trimming.js
+++ b/shared/js/background/events/referrer-trimming.js
@@ -4,17 +4,23 @@ const utils = require('../utils.es6')
 const browserName = utils.getBrowserName()
 
 /**
+ * @typedef Referrer
+ * @property {string} site
+ *   The referrer URL.
+ * @property {string} referrerHost
+ *   The referrer host.
+ * @property {string} referrer
+ *   The truncated referrer.
+ */
+
+/**
  * @param {{tabId: number, url: string, requestHeaders: Array<{name: string, value:string}>}} e
  *
  * @returns {{requestHeaders: Array<{name: string, value:string}>} | { redirectUrl: URL } | undefined}
  */
 module.exports = function limitReferrerData (e) {
-    let referrer = e.requestHeaders.find(header => header.name.toLowerCase() === 'referer')
-    if (referrer) {
-        referrer = referrer.value
-    } else {
-        return
-    }
+    const referrer = e.requestHeaders.find(header => header.name.toLowerCase() === 'referer')?.value
+    if (!referrer) return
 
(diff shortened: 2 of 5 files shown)


shared/js/background/events/referrer-trimming.js
```diff
@@ -3,18 +3,24 @@ const trackerutils = require('../tracker-utils')
...
  * @returns {{requestHeaders: Array<{name: string, value:string}>} | { redirectUrl: URL } | undefined}
  */
 module.exports = function limitReferrerData (e) {
-    let referrer = e.requestHeaders.find(header => header.name.toLowerCase() === 'referer')
-    if (referrer) {
-        referrer = referrer.value
-    } else {
-        return
-    }
+    const referrer = e.requestHeaders.find(header => header.name.toLowerCase() === 'referer')?.value
```
jonathanKingston, line 22: Was this needed, what it was previously is much more readable to me. (I'd rather not get into opinionated reviews just interested why it was changed)

shared/js/background/events/referrer-trimming.js
```diff
@@ -3,18 +3,24 @@ const trackerutils = require('../tracker-utils')
...
 
+/**
+ * @typedef Referrer
+ * @property {string} site
+ *   The referrer URL.
+ * @property {string} referrerHost
+ *   The referrer host.
+ * @property {string} referrer
+ *   The truncated referrer.
+ */
```
jonathanKingston, line 14: Not a bug; just a suggestion we could discuss, if you differ.

I think we should prefer classes where possible like:
```suggestion
 class Referrer {
     /**
      * @property {string} site The referrer URL.
      * @property {string} referrerHost The referrer host.
      * @property {string} referrer The truncated referrer.
      */
     constructor(site, referrerHost, referrer) {
         this.site = site
         this.referrerHost = referrerHost
         this.referrer = referrer
     }
 }
```

This means the usage down below would become typed too with:
```
tab.referrer = new Referrer(tab.site.url, new URL(referrer).hostname, modifiedReferrer)
```

- The consumer site in tab would then import a class, in some cases this also has the benefit that it can be a regular import in JS rather than a TS comment.
- I also like this model as it can serve as an extension point for anything that this object represents, the concerns can be methods instead.

shared/js/background/classes/tab.es6.js
```diff
@@ -57,6 +57,9 @@ class Tab {
 
         /** @type {null | import('./ad-click-attribution-policy').AdClick} */
         this.adClick = null
+
+        /** @type {null | import('../events/referrer-trimming').Referrer} */
+        this.referrer = null
```
jonathanKingston, line 62: Nice I think this was missing from my other patch too 👍🏻

jonathanKingston (commented):
No real issues just two discussion points 👍🏻

shared/js/background/events/referrer-trimming.js
jonathanKingston: Actually ignore this the other side is equally suboptimal because of shadowing and I'm guessing TS wasn't happy there.

In general I prefer more temporary vars over chaining like this.

Will close though.

jonathanKingston (approved):

PR #1389, opened 2022-09-05
Get script injection working for Chrome MV3 builds of the extension

kzar:
Chrome Manifest V3 introduces the new scripting API[1] that can inject
scripts directly into the "main world" of a website. Along with that,
Manifest V3 prevents the old pattern of injecting a script element
into the DOM. Adjust to those changes here, to get our injected
protections working again.

1 - https://developer.chrome.com/docs/extensions/reference/scripting

## Steps to test this PR:
1. Navigate to the [storage blocking test page](https://privacy-test-pages.glitch.me/privacy-protections/storage-blocking/).
2. Click "Store data", then "Retrieve data", expand the results, and note which are blocked.
3. Build MV3 extension: npm install && npm run dev-chrome-mv3
4. Install that in Chrome.
5. Load the test page again and retry the test in step 2.
6. Ensure that more third party cookies are blocked.

## Automated tests:
- [ ] Unit tests
- [x] Integration tests

###### Reviewer Checklist:

###### PR Author Checklist:


shared/js/background/script-injection.js
```diff
@@ -0,0 +1,25 @@
+const browserWrapper = require('./wrapper.es6')
+
+// With Manifest V2 the content scripts are run directly from the manifest and
+// they inject into the main world of websites by adding a <script> element to
+// the DOM. With Manifest V3 however the chrome.scripting API must be used to
+// inject scripts into the main world.
+if (browserWrapper.getManifestVersion() === 3) {
+    // @ts-ignore - The chrome type does not yet know about
```
jonathanKingston, line 8: I'll merge https://github.com/duckduckgo/duckduckgo-privacy-extension/pull/1391 now I think this might be fixed.

kzar: Dang, just checked but it's still not in there :(.

jonathanKingston: :( looks like they're working on that area though from the quick scan.

shared/js/content-scripts/content-scope-messaging.js
```diff
@@ -0,0 +1,52 @@
+function getSecretElement () {
+    // Main-world script ran first, grab the element.
+    if (document.getElementById('ddg-secret')) {
```
jonathanKingston, line 3: I think we should consider trying the following approaches instead:

1. window.sessionStorage and addEventListener('storage')
2. Firing a custom event and storing the data on window.navigator.duckduckgo

Both I think will have less side effects to the page, if not we should probably use an inert tag that's never visible; like <script> but with a type of application/json or something.

kzar: _(For reference, Jonathan and I discussed this in MatterMost. Since we figured out how to ensure the isolated world script was executed first, it was enough to simply emit the secret event from the main world script.)_

shared/js/background/script-injection.js
```diff
@@ -0,0 +1,27 @@
+const browserWrapper = require('./wrapper.es6')
+
+// With Manifest V2 the content scripts are run directly from the manifest and
+// they inject into the main world of websites by adding a <script> element to
+// the DOM. With Manifest V3 however the chrome.scripting API must be used to
+// inject scripts into the main world.
+// Note: It's important that the isolated world script runs first. Script
+//       execution order is determined by the script ID.
```
jonathanKingston, line 8: I'm not sure it is, https://source.chromium.org/chromium/chromium/src/+/main:components/on_load_script_injector/browser/on_load_script_injector_host.cc;l=46;drc=d00c1ed3052794f3bae53ad76b56209ecb290902 this code pushes it to the back of the queue if the ID doesn't match.

kzar: Hmm, well I'm going from my testing. The order of the scripts in the `registerContentScripts` call made no difference, but the ID of the content script consistently determined which ran first. I have not looked at the Chromium code in detail though. I guess our integration tests will catch this if it changes.

jonathanKingston: Cool happy to keep both strategies in case it got changed or will be etc :)

kzar: I think it's since registerContentScripts puts the new script IDs into a std::set and then later iterates through those IDs to add the scripts. See https://source.chromium.org/chromium/chromium/src/+/main:chrome/browser/extensions/api/scripting/scripting_api.cc;l=929;drc=13dced01cce80364299ad7663479ad65069db73d;bpv=0;bpt=1. I will update the comment to say that the ordering is based on ID within the call.

kzar: (Updated that comment.)

PR #1398, opened 2022-09-09
Add Chrome MV3 support for "surrogate script" redirections

kzar:
The extension redirects requests to common third-party scripts to subbed out "surrogate scripts". We do that instead of simply blocking the third-party scripts since this way, we avoid breaking websites that assume the APIs provided by the third-party scripts will always be present.

## Steps to test this PR:
1. Build the chrome-mv3 extension: `npm install && npm run dev-chrome-mv3`
2. Install that
3. Browse to https://www.theguardian.com/
4. Open the developer tools, switch to the "Network" panel and ensure that the following surrogate scripts are redirected: `gpt.js` and `analytics.js`.

Note: It is expected that some redirections will fail with an "net::ERR_UNSAFE_REDIRECT" error. See https://crbug.com/1361350

## Automated tests:
- [x] Unit tests
- [ ] Integration tests

###### Reviewer Checklist:

###### PR Author Checklist:

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10,7 +10,7 @@
             "dependencies": {
                 "@duckduckgo/autofill": "github:duckduckgo/duckduckgo-autofill#5.1.0",
                 "@duckduckgo/content-scope-scripts": "github:duckduckgo/content-scope-scripts#2.6.0",
-                "@duckduckgo/ddg2dnr": "github:duckduckgo/ddg2dnr#0.1.4",
+                "@duckduckgo/ddg2dnr": "github:duckduckgo/ddg2dnr#0.2.0",
                 "@duckduckgo/jsbloom": "^1.0.2",
                 "@duckduckgo/privacy-grade": "github:duckduckgo/privacy-grade#2.1.1",
                 "@duckduckgo/privacy-reference-tests": "github:duckduckgo/privacy-reference-tests",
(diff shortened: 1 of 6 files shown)


Makefile
```diff
@@ -40,7 +40,8 @@ web-resources:
 	cp shared/data/web_accessible_resources/* build/$(browser)/$(type)/web_accessible_resources/
 	cp shared/tracker-surrogates/surrogates/*.js build/$(browser)/$(type)/web_accessible_resources/
 	mkdir -p build/$(browser)/$(type)/data
-	node scripts/generateListOfSurrogates.js -i build/$(browser)/$(type)/web_accessible_resources/ >> build/$(browser)/$(type)/data/surrogates.txt
```
jonathanKingston, line 43: I don't think we use this .txt file at all (perhaps in tests).

Does https://github.com/duckduckgo/tracker-surrogates/blob/main/mapping.json serve the same purpose? I think we could import that as other clients do also.

jonathanKingston: Ignore that we're not using it processSurrogateList is in privacy-grade.

However I wonder if consuming the mapping.json would be cleaner?

jonathanKingston (approved):

jonathanKingston:
LGTM thanks for making the changes :)

PR #1410, opened 2022-09-19
Bugfix: getBaseDomain should return null when invalid

englehardt:
This fixes a regression in the ad domain detection where we expect getBaseDomain to return null when a domain provided by SERP is unparsable, but getBaseDomain will always return the hostname (even if it's nonsensical).

This is Ad 11 and Add 12 in https://www.search-company.site/

I also fixed a no-op test that we forgot to update when we first merged the bat.js work. I only detected it because I had an outdated privacy grade reference that was still returning `firstParty` when running the tests locally.

## Automated tests:
- [ ] Unit tests
- [ ] Integration tests

###### Reviewer Checklist:

###### PR Author Checklist:


jonathanKingston (approved):
Thanks for this!

PR #1450, opened 2022-10-11
Update #1313 for MV3

jdorweiler:
**Reviewer:**

## Description:

Updating #1313 for Chrome MV3

## Automated tests:
- [ ] Unit tests
- [ ] Integration tests

###### Reviewer Checklist:

###### PR Author Checklist:


shared/js/background/classes/ad-click-attribution-policy.js
```diff
@@ -143,6 +161,8 @@ export class AdClick {
      */
     shouldPropagateAdClickForNewTab (tab) {
         if (tab.site.baseDomain === this.adBaseDomain) {
+            this.adClickDNR.rule.condition.tabIds.push(tab._tabState.tabId)
+            this.adClickDNR.updateAdClickDNR()
```
jonathanKingston, line 165: Do we need this here?

I suspect if the tab has been navigated the existing rule should work or should be removed?

shared/js/background/classes/ad-click-attribution-policy.js
```diff
@@ -173,4 +198,50 @@ export class AdClick {
...
+export class AdClickDNR {
+    constructor (tabId, allowlist) {
+    this.allowlist = allowlist
+    this.tabId = tabId
+    this.initiatorDomain = null
+    this.rule  = generateDNRRule({
+            "id" : tabId,
+            "priority": CEILING_PRIORITY,
+            "actionType": "allow",
+            "requestDomains": allowlist.reduce((lst, entry) => { lst.push(entry.host); return lst}, []),
```
jonathanKingston, line 213: Doesn't this do the same thing:

```suggestion
            "requestDomains": allowlist.map((entry) => entry.host),
```

shared/js/background/classes/ad-click-attribution-policy.js
```diff
@@ -173,4 +198,50 @@ export class AdClick {
...
+     **/
+    createInitialAdClickDNR () {
+        chrome.declarativeNetRequest.updateSessionRules({addRules: [this.rule]})
+    }
+
+    /*
+     * Update the tab DNR with initiator domain
+     */
+    addAdClickDNRInitiatorDomain (domain) {
+        this.initiatorDomain = domain
```
jonathanKingston, line 231: Is this method called?

shared/js/background/classes/ad-click-attribution-policy.js
```diff
@@ -173,4 +198,50 @@ export class AdClick {
...
+        this.rule.condition.initiatorDomain = [domain]
+
+    }
+
+    removeAdClickDNR () {
```
jonathanKingston, line 236: This needs to be triggered when the tab is navigated via the top URL.

I think we have similar clean up code in tab manager (or calls into it) I think we want to ensure we clear this out whenever a tab is removed etc.

shared/js/background/events.es6.js
```diff
@@ -183,8 +183,8 @@ browser.webRequest.onBeforeRequest.addListener(
     additionalOptions
 )
 
-if (manifestVersion === 2) {
```
jonathanKingston, line 186: Clean up to ensure MV2+3 both work :)

shared/js/background/classes/ad-click-attribution-policy.js
```diff
@@ -109,31 +120,38 @@ export class AdClick {
...
     }
 
     clone () {
-        const adClick = new AdClick(this.navigationExpiration, this.totalExpiration)
+        const adClick = new AdClick(this.navigationExpiration, this.totalExpiration, this.tabId)
         adClick.adBaseDomain = this.adBaseDomain
         adClick.adClickRedirect = this.adClickRedirect
         adClick.expires = this.expires
         adClick.clickExpires = Date.now() + (this.navigationExpiration * 1000)
+        adClick.adClickDNR = this.adClickDNR
```
jonathanKingston, line 144: I think when we clone, we probably want to remove the rule and re-create perhaps?

shared/js/background/classes/ad-click-attribution-policy.js
jdorweiler: Here or do we need this at all? We need to add the new tab id to the existing rule for it to work.

jonathanKingston: But then we have two tabId's within a single AdClick that's within the Tab object.

- If you're trying to minimise rules to DNR then you'll need to use a different global to store them under.
- Essentially your Tab object would be: `{tabId: 12, adClick: {adClickDNR: {tabIds: [12, 14]}}, ...}` which doesn't make much sense.

I think this should be handled in the new Tab being constructed and creating a new AdClick object.

shared/js/background/classes/ad-click-attribution-policy.js
```diff
@@ -174,3 +207,52 @@ export class AdClick {
         return this.hasNotExpired()
     }
 }
+
+/**
+ * @param {number} tabId
+ * @param {object} allowlist
+ * **/
```
kzar, line 214: Nit: Typo with the comment close here.
```suggestion
 **/
```

shared/js/background/classes/ad-click-attribution-policy.js
```diff
@@ -174,3 +207,52 @@ export class AdClick {
...
+            priority: AD_ATTRIBUTION_POLICY_PRIORITY,
+            actionType: 'allow',
+            requestDomains: allowlist.map((entry) => entry.host)
+        })
+        this.rule.condition.tabIds = [tabId]
+        this.createInitialAdClickDNR()
+    }
+
+    /**
+     * Create initial tab scoped DNR not limited to an initiatorDomain
```
kzar, line 231: Nit:
```suggestion
     * Create initial tab scoped DNR rule not limited to any initiatorDomains.
```

shared/js/background/classes/ad-click-attribution-policy.js
```diff
@@ -174,3 +207,52 @@ export class AdClick {
...
+     **/
+    createInitialAdClickDNR () {
+        chrome.declarativeNetRequest.updateSessionRules({ addRules: [this.rule] })
+    }
+
+    /*
+     * Update the tab DNR with initiator domain
```
kzar, line 238: Nit:
```suggestion
     * Update the tab scope DNR rule with an initiatorDomains condition.
```

shared/js/background/classes/ad-click-attribution-policy.js
```diff
@@ -174,3 +207,52 @@ export class AdClick {
...
+    }
+
+    /**
+     * Create initial tab scoped DNR not limited to an initiatorDomain
+     **/
+    createInitialAdClickDNR () {
+        chrome.declarativeNetRequest.updateSessionRules({ addRules: [this.rule] })
+    }
+
+    /*
```
kzar, line 237: Nit:
```suggestion
    /**
```

shared/js/background/dynamic-rule-id.js
```diff
@@ -0,0 +1,40 @@
+const browserWrapper = require('./wrapper.es6')
+const DYNAMIC_RULE_ID_START = 100000
```
kzar, line 2: I'm wondering if this dynamic rule ID logic should be in the `declarative-net-request.js` module. But if not, I think that:
1. we should export the `DYNAMIC_RULE_ID_START` constant from the `delcarative-net-request.js` module and import it here. We should add a comment in  `declarative-net-request.js` for the constant to mention that rule IDs > 100k are reserved for dynamic rule IDs, and to reference dynamic-rule-is.js.
2. We should add DNR or similar to the name of this module, just so it's a bit clearer what the module does from the name. E.g. how about `dynamic-dnr-rule-id.js`? (While working on this, I often found the term "rule" confusing since it could be a tracker blocking rule, or a declarativeNetRequest rule... so good to specify where possible.)
3. We should add a comment near the top of this module to explain a bit what the point of all this dynamic rule ID stuff is.

shared/js/background/dynamic-rule-id.js
```diff
@@ -0,0 +1,40 @@
...
+    rules = rules.concat(set1.filter(r => r.id >= DYNAMIC_RULE_ID_START))
+    const set2 = await chrome.declarativeNetRequest.getSessionRules()
+
+    // get all rules that were already dynamically created
+    rules = rules.concat(set2.filter(r => r.id >= DYNAMIC_RULE_ID_START))
+    return rules
+}
+
+// set offset and update based on rules that have been created
+(async () => {
```
kzar, line 27: Might as well call this Promise `ready` and export it, so that other modules can wait for this to be ready if necessary.

shared/js/background/dynamic-rule-id.js
```diff
@@ -0,0 +1,40 @@
...
+    // get existing offset in storage
+    dynamicRuleOffset = await browserWrapper.getFromSessionStorage(DYNAMIC_RULE_STORAGE_KEY) || 0
```
kzar, line 29: I figure that if reading the existing offset from session storage is successful, we should just use that. If reading wasn't successful _then_ we want to figure out what the offset is the hard way. I figure that's kind of the whole point of the session variable, to avoid us having to constantly re-read all the DNR rules every time the background ServiceWorker restarts.

shared/js/background/events.es6.js
```diff
@@ -183,75 +183,79 @@ browser.webRequest.onBeforeRequest.addListener(
...
-    const extraInfoSpec = ['blocking', 'responseHeaders']
-    if (browser.webRequest.OnHeadersReceivedOptions.EXTRA_HEADERS) {
-        extraInfoSpec.push(browser.webRequest.OnHeadersReceivedOptions.EXTRA_HEADERS)
-    }
+const extraInfoSpec = manifestVersion === 3 ? ['responseHeaders'] : ['blocking', 'responseHeaders']
```
kzar, line 186: Nit: Doesn't really matter, but my preference would be to do something like:
```suggestion
const extraInfoSpec = ['responseHeaders']
if (manifestVersion === 2) {
    extraInfoSpec.push('blocking')
}
```

Might be nice to add a comment too, something to explain that for MV2 webRequest is doing the blocking but we still need the event listener on MV3 to help do the reporting stuff.

shared/js/background/events.es6.js
```diff
@@ -183,75 +183,79 @@ browser.webRequest.onBeforeRequest.addListener(
...
 
-            if (ATB.shouldUpdateSetAtb(request)) {
-                // returns a promise
-                return ATB.updateSetAtb()
-            }
+// We determine if browsingTopics is enabled by testing for availability of its
+// JS API.
+// Note: This approach will not work with MV3 since the background
+//       ServiceWorker does not have access to a `document` Object.
+const isTopicsEnabled = manifestVersion === 3 ? false : ('browsingTopics' in document) && utils.isFeatureEnabled('googleRejected')
```
kzar, line 196: I recommend against combining the inline `manifestVersion === 3 ?` condition with this, since the precedence can get kind of confusing (is the `&&` applying inside the else condition, or to the overall condition?)

How about this?
```suggestion
const isTopicsEnabled = manifestVersion === 2 && 'browsingTopics' in document && utils.isFeatureEnabled('googleRejected')
```

shared/js/background/events.es6.js
```diff
@@ -183,75 +183,79 @@ browser.webRequest.onBeforeRequest.addListener(
...
-                // According to Google's docs a site can opt out of browsing topics the same way as opting out of FLoC
-                // https://privacysandbox.com/proposals/topics (See FAQ)
-                responseHeaders.push({ name: 'permissions-policy', value: 'interest-cohort=()' })
+            const tab = tabManager.get({ tabId: request.tabId })
+            // SERP ad click detection
+            if (
+                utils.isRedirect(request.statusCode)
+            ) {
+                tab.setAdClickIfValidRedirect(request.url)
+            } else if (tab && tab.adClick && tab.adClick.adClickRedirect && !utils.isRedirect(request.statusCode)) {
```
kzar, line 209: Nit: Could use optional properties?
```suggestion
            } else if (tab && tab.?adClick.?adClickRedirect && !utils.isRedirect(request.statusCode)) {
```

kzar (changes requested):

shared/js/background/dynamic-rule-id.js
jdorweiler: 1. I've had timing issues using `delcarative-net-request.js`. Seems like it's used elsewhere on startup and requiring it in other areas can cause errors.
2. 👍 
3. 👍

shared/js/background/classes/ad-click-attribution-policy.js
```diff
@@ -65,7 +77,11 @@ export class AdClickAttributionPolicy {
         const linkFormat = this.getMatchingLinkFormat(resourceURL)
         if (!linkFormat) return
 
-        const adClick = new AdClick(this.navigationExpiration, this.totalExpiration)
+        const adClick = new AdClick(this.navigationExpiration, this.totalExpiration, this.allowlist)
+
+        if (manifestVersion === 3) {
+            adClick.adClickDNR = new AdClickDNR(tab._tabState.tabId, this.allowlist)
```
jonathanKingston, line 83: Don't access _tabState directly
```suggestion
            adClick.adClickDNR = new AdClickDNR(tab.tabId, this.allowlist)
```

shared/js/background/classes/ad-click-attribution-policy.js
```diff
@@ -174,3 +207,54 @@ export class AdClick {
...
 }
+
+/**
+ * @param {number} tabId
+ * @param {object} allowlist
+ **/
+export class AdClickDNR {
+    constructor (tabId, allowlist) {
+        this.allowlist = allowlist
+        this.tabId = tabId
```
jonathanKingston, line 218: Neither of these are currently needed
```suggestion
```

shared/js/background/tab-manager.es6.js
```diff
@@ -44,6 +44,7 @@ class TabManager {
     }
 
     delete (id) {
+        this.tabContainer[id]?.adClick?.adClickDNR?.removeAdClickDNR()
```
jonathanKingston, line 47: Same here I think a remove method on adClick would make the most sense rather than many consumers know the internals of the class structures.

shared/js/background/classes/ad-click-attribution-policy.js
```diff
@@ -174,3 +207,54 @@ export class AdClick {
...
+        const ruleId = await getNextSessionRuleId()
+        this.rule.id = ruleId
+        chrome.declarativeNetRequest.updateSessionRules({ addRules: [this.rule] })
+    }
+
+    /*
+     * Update the tab DNR with initiator domain
+     */
+    addAdClickDNRInitiatorDomain (domain) {
+        if (!this.initiatorDomain) {
```
jonathanKingston, line 243: Can you explain why not set this up front or when it's set in adClick?

Are we perhaps creating a lot of additional work by separating out the DNR class?

shared/js/background/events.es6.js
jonathanKingston: Could we make this and the indenting change a separate PR; largely I don't think much of this file has changed but reviewing it is a little harder than it should be ideally.

shared/js/background/dnr-session-rule-id.js
```diff
@@ -0,0 +1,47 @@
...
+    
+    // if the rule offset is undef or zero let's check existing rules to see if it needs to be updated
+    if (!currentSessionRuleOffset) {
+
+        // get any session rules that have been created
+        const currentSessionRules = await chrome.declarativeNetRequest.getSessionRules()
+
+        if (currentSessionRules) {
+            // get offsets for session rules already created
+            const offsetsAssigned = currentSessionRules.map(r => (r.id - SESSION_RULE_ID_START)).sort((a, b) => b - a)
```
jonathanKingston, line 34: I think this map should be a filter no?
```suggestion
            const offsetsAssigned = currentSessionRules.filter(r => (r.id >= SESSION_RULE_ID_START)).sort((a, b) => b - a)
```

shared/js/background/classes/ad-click-attribution-policy.js
jdorweiler: `tab.id`

shared/js/background/classes/ad-click-attribution-policy.js
jdorweiler: I might be able to set it in adClick. Let me test that out.

shared/js/background/classes/ad-click-attribution-policy.js
```diff
@@ -134,9 +157,26 @@ export class AdClick {
...
+     **/
+    setAdBaseDomain (domain, adClickRedirect) {
+        this.adBaseDomain = domain
+        this.adClickRedirect = adClickRedirect
+        this.adClickDNR?.updateInitiator(domain)
+    }
+
+    removeAdClickDNR () {
+        if (this.adClickDNR) {
+            chrome.declarativeNetRequest.updateSessionRules({ removeRuleIds: [this.adClickDNR.rule.id] })
```
jonathanKingston, line 176: This either belongs in AdClickDNR or AdClick should take care of all AdClickDNR's code too.

So either:
```suggestion
           this.adClickDNR?.remove()
```

Or bring all the class in here :)

jonathanKingston: If the former, it totally could be called from within hasNotExpired

shared/js/background/classes/ad-click-attribution-policy.js
```diff
@@ -134,9 +157,26 @@ export class AdClick {
         restoredAdClick.adClickRedirect = adClick.adClickRedirect
         restoredAdClick.expires = adClick.expires
         restoredAdClick.clickExpires = adClick.clickExpires
+        restoredAdClick.adClickDNR = adClick.adClickDNR
```
jonathanKingston, line 160: Doesn't this need to be constructed again?
If you kill the worker, is this restoring properly?

shared/js/background/events.es6.js
jonathanKingston: Scrap my indenting comment, it's easy to filter on the cli and in github options (the patch looks really great without whitspace now :) ).

shared/js/background/classes/ad-click-attribution-policy.js
jdorweiler: At this point all we need is the rule ID. Reconstructing also generates a new rule id and a new DNR, which we don't want. 
We could recreate it here, but then we also need to explicitly call some getID and createRule methods.

shared/js/background/classes/ad-click-attribution-policy.js
jdorweiler: I'll just move the class in here.  Calling remove on adClickDNR would require recreating it when cloning which gets complicated.

shared/js/background/classes/ad-click-attribution-policy.js
```diff
@@ -109,34 +125,62 @@ export class AdClick {
...
+        adClick.adClickDNR = this.adClickDNR
+        return adClick
+    }
+
+    propagate (tabId) {
+        const adClick = this.clone()
+
+        if (this.adClickDNR) {
+            adClick.adClickDNR = adClick.getAdClickDNR(tabId, this.allowlist)
+            adClick.createDNR()
```
jonathanKingston, line 155: I think these calls can be combined.
```suggestion
            adClick.createDNR(tabId, this.allowlist)
```

jonathanKingston: We don't actually need to pass this.allowlist around now we're in the same class.

shared/js/background/dnr-session-rule-id.js
```diff
@@ -0,0 +1,20 @@
...
+let sessionRuleOffset = 0
+browserWrapper.getFromSessionStorage(SESSION_RULE_STORAGE_KEY).then(offset => {
+    if (offset) {
+        sessionRuleOffset = offset
+    }
+})
+
+export function getNextSessionRuleId () {
+    const nextRuleId = SESSION_RULE_ID_START + sessionRuleOffset
+    sessionRuleOffset += 1
```
jonathanKingston, line 17: This should await the getting from storage.

shared/js/background/dnr-session-rule-id.js
```diff
@@ -0,0 +1,20 @@
...
+    browserWrapper.setToSessionStorage(SESSION_RULE_STORAGE_KEY, sessionRuleOffset)
```
jonathanKingston, line 18: We should probably await here also

shared/js/background/classes/ad-click-attribution-policy.js
```diff
@@ -65,15 +77,19 @@ export class AdClickAttributionPolicy {
         const linkFormat = this.getMatchingLinkFormat(resourceURL)
         if (!linkFormat) return
 
-        const adClick = new AdClick(this.navigationExpiration, this.totalExpiration)
+        const adClick = new AdClick(this.navigationExpiration, this.totalExpiration, this.allowlist)
```
jonathanKingston, line 80: Nit: We could just pass a reference to `this` in here and *not* store any of these three variables; we can regenerate the `AdClickAttributionPolicy` whenever or have a global to store one in like we do for tab.

shared/js/background/dnr-session-rule-id.js
jdorweiler: Can we stop the tab from loading while we wait?

shared/js/background/classes/ad-click-attribution-policy.js
```diff
@@ -109,34 +125,62 @@ export class AdClick {
...
         adClick.adBaseDomain = this.adBaseDomain
         adClick.adClickRedirect = this.adClickRedirect
         adClick.expires = this.expires
         adClick.clickExpires = Date.now() + (this.navigationExpiration * 1000)
+        adClick.adClickDNR = this.adClickDNR
+        return adClick
+    }
+
+    propagate (tabId) {
```
jonathanKingston, line 150: Let's document what this is for, something like:

```suggestion
    /**
     * Propagate an adclick to a new tab, used when a user navigates to a new tab.
     * @param {number}
     */
    propagate (tabId) {
```

shared/js/background/classes/ad-click-attribution-policy.js
```diff
@@ -109,34 +125,62 @@ export class AdClick {
...
         return restoredAdClick
     }
 
+    /**
+     * @param {string} domain
+     * @param {boolean} adClickRedirect
+     **/
+    setAdBaseDomain (domain, adClickRedirect) {
+        this.adBaseDomain = domain
+        this.adClickRedirect = adClickRedirect
```
jonathanKingston, line 177: I think this should always be false.

shared/js/background/dnr-session-rule-id.js
jonathanKingston: No so you probably want to precalculate this and continue to have this method synchronous.

kzar: We're reading the ID offset from session storage when the background ServiceWorker starts, so that `getNextSessionRuleId` can return synchronously. We're doing that to avoid the race condition where the allowing DNR rule is not added quickly enough to ensure the corresponding request isn't blocked.

shared/js/background/dnr-session-rule-id.js
```diff
@@ -0,0 +1,20 @@
...
+const SESSION_RULE_ID_START = 100000
+const SESSION_RULE_STORAGE_KEY = 'sessionRuleOffset'
+let sessionRuleOffset = 0
+browserWrapper.getFromSessionStorage(SESSION_RULE_STORAGE_KEY).then(offset => {
+    if (offset) {
+        sessionRuleOffset = offset
+    }
+})
+
+export function getNextSessionRuleId () {
```
kzar, line 15: Please can you add a JSDoc comment and unit tests for this? We're likely going to rely on it for Click to Load as well.

shared/js/background/dnr-session-rule-id.js
```diff
@@ -0,0 +1,20 @@
+/**
+ * For managing dynamically created MV3 session rules
+ * getNextSessionRuleId will return the next unique session rule id to use when creating new session rules
+ **/
+const browserWrapper = require('./wrapper.es6')
+const SESSION_RULE_ID_START = 100000
```
kzar, line 6: Please can you exportt `SESSION_RULE_ID_START` from the declarative-net-request module and add a comment there to explain?

shared/js/background/dnr-session-rule-id.js
```diff
@@ -0,0 +1,20 @@
+/**
+ * For managing dynamically created MV3 session rules
+ * getNextSessionRuleId will return the next unique session rule id to use when creating new session rules
+ **/
+const browserWrapper = require('./wrapper.es6')
```
kzar, line 5: Nit: Please use ES Module syntax for this import.
```suggestion
import * as browserWrapper from './wrapper.es6'
```

kzar (dismissed):

shared/js/background/dnr-session-rule-id.js
jonathanKingston: I think we're saying the same thing here.

We don't want an await on what we call sync in a tab I agree; however we should await this somewhere otherwise we risk overriding the id. We need an async queue to block getting the next value (somewhere) otherwise two writes could write different values to the store.

The two issues as I see them:

1. `getNextSessionRuleId` could be triggered before the `sessionRuleOffset` is called.
2. `getNextSessionRuleId` could be called twice whilst `setToSessionStorage` happens resulting in an off by 1 error.

Perhaps the risk of 2. happening and a service worker shut down is too low to be bothered with the defensive code? 1. however seems pretty likely.

shared/js/background/dnr-session-rule-id.js
```diff
@@ -0,0 +1,23 @@
...
+const SESSION_RULE_ID_START = 100000
+const SESSION_RULE_STORAGE_KEY = 'sessionRuleOffset'
+let sessionRuleOffset = 0
+
+export async function setSessionRuleOffsetFromStorage () {
+    await browserWrapper.getFromSessionStorage(SESSION_RULE_STORAGE_KEY).then(offset => {
+        if (offset) {
+            sessionRuleOffset = offset
+        }
+    })
```
jonathanKingston, line 15:
```suggestion
    const offset = await browserWrapper.getFromSessionStorage(SESSION_RULE_STORAGE_KEY)
    if (offset) {
        sessionRuleOffset = offset
    }
```

shared/js/background/classes/ad-click-attribution-policy.js
jdorweiler: Do you have a preference? Seems like extra work to me.

jonathanKingston: I think the latter; let's leave it for a follow up actually though :)

shared/js/background/dnr-session-rule-id.js
jonathanKingston: Let's tackle this in a small follow up commit and land this for now.

PR #1464, opened 2022-10-21
Add Chrome MV3 support for user allowlisting

kzar:
Users can allowlist (disable protections) for websites from the popup
UI and also from the options page. To make that work with Chrome MV3,
we need to add a declarativeNetRequest rule that ensures requests
initiated by domains that the user has allowlisted aren't blocked or
redirected.

**Reviewer:** @jdorweiler 
**CC:** @jonathanKingston 

## Steps to test this PR:
1. Produce a Chrome MV3 build of the extension (`npm install && npm run dev-chrome-mv3`).
2. Install that in Chrome.
3. Try allowlisting websites, both from the options page and the popup UI and ensure it works.
4. Click to reload the extension from the extensions page.
5. Make sure the websites you allowlisted are still allowlisted (try opening them in a new tab too).
6. Remove any allowlisted websites from the options page and ensure they are no longer allowlisted.

## Automated tests:
- [x] Unit tests
- [ ] Integration tests

###### Reviewer Checklist:
- [ ] **Ensure the PR solves the problem**
- [ ] **Review every line of code**
- [ ] **Ensure the PR does no harm by testing the changes thoroughly**
- [ ] **Get help if you're uncomfortable with any of the above!**
- [ ] Determine if there are any quick wins that improve the implementation

###### PR Author Checklist:
- [ ] Get advice or leverage existing code
- [ ] Agree on technical approach with reviewer (if the changes are nuanced)
- [ ] Ensure that there is a testing strategy (and documented non-automated tests)
- [ ] Ensure there is a documented monitoring strategy (if necessary)
- [ ] Consider systems implications

PR #1471, opened 2022-10-26
Fix circular dependencies resulting from importing startup.ready()

kzar:
The startup module exported a ready() function that could be used to check if the extension's startup sequence had completed. That was useful to help prevent race conditions at startup. The problem with that though, was that the startup module itself imports quite a few other modules. That resulted in a trap, importing startup.ready() from the wrong module (one that startup itself needs to import) would result in a circular dependency and cause peculiar timeouts and failures.

Split out startup.ready() into the ready module (ready.startupReady() and ready.startupReadyResolve()). That module can be safely imported as it has no dependencies of its own. In the future, the ready module could be also used for other similar ready Promises.

**Reviewer:** @sammacbeth 

## Steps to test this PR:
1. Ensure the tests still pass.

## Automated tests:
- [x] Unit tests
- [x] Integration tests

###### Reviewer Checklist:

###### PR Author Checklist:

### DIFF
--- a/shared/js/background/events.es6.js
+++ b/shared/js/background/events.es6.js
@@ -242,7 +242,7 @@ browser.webRequest.onHeadersReceived.addListener(
 //       to be compatible with MV3. Registering the listener asynchronously
 //       is only possible here as this is a MV2-only event listener!
 // See https://developer.chrome.com/docs/extensions/mv3/migrating_to_service_workers/#event_listeners
-startupReady().then(() => {
+startupReady.ready().then(() => {
     browser.webRequest.onHeadersReceived.addListener(
         dropTracking3pCookiesFromResponse,
         { urls: ['<all_urls>'] },
@@ -415,7 +415,7 @@ if (manifestVersion === 2) {
     //       to be compatible with MV3. Registering the listener asynchronously
     //       is only possible here as this is a MV2-only event listener!
     // See https://developer.chrome.com/docs/extensions/mv3/migrating_to_service_workers/#event_listeners
-    startupReady().then(() => {
+    startupReady.ready().then(() => {
         browser.webRequest.onBeforeSendHeaders.addListener(
             dropTracking3pCookiesFromRequest,
             { urls: ['<all_urls>'] },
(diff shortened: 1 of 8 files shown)


shared/js/background/events.es6.js
```diff
@@ -19,7 +19,7 @@ const tdsStorage = require('./storage/tds.es6')
 const browserWrapper = require('./wrapper.es6')
 const limitReferrerData = require('./events/referrer-trimming')
 const { dropTracking3pCookiesFromResponse, dropTracking3pCookiesFromRequest } = require('./events/3p-tracking-cookie-blocking')
-const startup = require('./startup.es6')
+const { startupReady } = require('./ready.es6')
```
jonathanKingston, line 22: Can this module just export `startup: {ready()}`? It would simplify the PR a lot.

shared/js/background/ready.es6.js
```diff
@@ -0,0 +1,36 @@
+/**
+ * Module that exports an API for announcing that, or checking if, a part of the
+ * extension's code is ready for use. This helps to avoid circular dependencies
+ * that can otherwise result from exporting ready functions directly from each
+ * module.
+ */
+
+/**
+ * @typedef readyResolverResult
```
jonathanKingston, line 9: Can we make this a class:
```suggestion
class ReadyResolverResult {
    constructor() {
        this._readyResolver = null
        this._readyPromise = new Promise(resolve => { _readyResolver = resolve })
    }
    ready () { return this._readyPromise },
    resolve () {
        if (this._readyResolver) {
            this._readyResolver()
            this._readyResolver = null
        }
    }
}

export const startup = new ReadyResolverResult()
```

shared/js/background/startup.es6.js
```diff
@@ -90,6 +81,5 @@ function registerUnloadHandler () {
 }
 
 module.exports = {
-    onStartup,
```
jonathanKingston, line 93: IIUC this is only needed as we're mixing cjs and esm right?

You're limiting startup to a single inclusion point via require; could we just convert this to ESM as only background.js needs it after this change.

jonathanKingston:
I think an alternative here is: https://github.com/duckduckgo/duckduckgo-privacy-extension/pull/1473

shared/js/background/startup.es6.js
kzar: No, unfortunately not. This change is a prerequisite for https://github.com/duckduckgo/duckduckgo-privacy-extension/pull/1464 since we now require `tab-manager` from startup with the recent changes that landed, we require `startup` from `declarative-net-request` already, and with my upcoming changes we need to require `declarative-net-request` from `tab-manager` too. I initially fixed it for just my case, but then I realised this is a trap that folks will continue falling in going forwards.

shared/js/background/ready.es6.js
kzar: Sure if you prefer, done.

PR #1476, opened 2022-10-26
Update checkout, cache and setup-node versions to v3. Remove tests on…

jonathanKingston:
… push and instead run on PR only

**Reviewer:**

## Automated tests:
- [ ] Unit tests
- [ ] Integration tests

###### Reviewer Checklist:

###### PR Author Checklist:

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -7,3 +7,10 @@ updates:
     target-branch: "develop"
     labels:
       - "dependencies"
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+    target-branch: "develop"
+    labels:
+      - "dependencies"
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -8,34 +8,24 @@ on:
 
 jobs:
   chrome:
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
       - uses: actions/setup-node@v3
         with:
-          node-version: 16.x
-      - uses: actions/cache@v3
-        with:
-          path: ~/.npm
-          key: ${{ runner.os }}-node-${{ hashFiles('**/package-lock.json') }}
-          restore-keys: |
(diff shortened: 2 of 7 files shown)


.github/workflows/build.yml
```diff
@@ -7,37 +7,42 @@ on:
...
+      - uses: actions/checkout@v3
+      - uses: actions/setup-node@v3
         with:
-          path: ~/.npm
-          key: ${{ runner.os }}-node-${{ hashFiles('**/package-lock.json') }}
-          restore-keys: |
-            ${{ runner.os }}-node-
+          node-version: 16
+          cache: 'npm'
+      - run: npm run install-ci
```
sammacbeth, line 45: I believe this is already run in the `release-firefox` script (via the Makefile). Same for `release-chrome` above.

.github/dependabot.yml
```diff
@@ -7,3 +7,10 @@ updates:
     target-branch: "develop"
     labels:
       - "dependencies"
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "weekly"
```
sammacbeth, line 13: Do we want to update actions so often? We use some third-party actions pinned to commits. When we update them, we have to audit them and verify they still work with our rarely-run release workflows. Ideally, I'd just leave them unless they cause an issue.

sammacbeth (approved):
Looks good. Just a potentially redundant npm install and a question about the actions update frequency.

.github/dependabot.yml
jonathanKingston: Yeah not bothered about the cadence just it seems easy to ignore the deprecation warnings like we have.

At some point many of the deprecations would have become fails; I personally like the little and often approach of mending things.

I suspect this will also be silent almost always; we'd have probably only got 4 PRs over the last 2 years maybe?

.github/workflows/build.yml
jonathanKingston: Ah nice; thanks! I spent an inordinate amount of time thinking there was something special about this workflow when splitting it out into dependent workflows.

PR #1485, opened 2022-10-27
Fix detection of popup UI close for Chrome MV3

kzar:
The extension's popup UI allows the user to do several things,
including disabling/enabling protections for the current website.

When the user disables protections for a website, the website is added
to the allowlist. After a delay (or when the user clicks away) the
popup UI is closed and the website is reloaded automatically. Until
now, that worked by listening for the window.unload event, and then
using the chrome.extension.getBackgroundPage() API to fetch the
background page before triggering the unload function in the
background.

With Chrome Manifest v3, the getBackgroundPage() API is no longer
available and so a different approach must be used. We're now opening
a messaging connection to the background, sending through user actions
as they happen, then finally having the background detect when the
connection is closed (popup is closed) to trigger any corresponding
actions like reloading the active website.

1 - https://developer.chrome.com/docs/extensions/mv3/mv3-migration-checklist/#api-background-context

**Reviewer:** @sammacbeth 
**CC:** @jdorweiler 

## Steps to test this PR:

Chrome MV2
1. Check that the developer tools panel still works OK
2. Check that when you allowlist a website in the popup, the website is refreshed automatically.
3. Check that when you don't toggle allowlist for a website in the popup, the website is _not_ refreshed automatically.

Chrome MV3
1. Check that when you allowlist a website in the popup, the website is refreshed automatically[1].
2. Check that when you don't toggle allowlist for a website in the popup, the website is _not_ refreshed automatically.
3. Open popup, allowlist a website, wait 5+ minutes for background SW to close, close popup UI and check website is reloaded automatically.

1 - Note that the actual allowlisting of the website won't work until https://github.com/duckduckgo/duckduckgo-privacy-extension/pull/1464 lands.

## Automated tests:
- [ ] Unit tests
- [ ] Integration tests

###### Reviewer Checklist:

###### PR Author Checklist:

### DIFF
--- a/shared/js/ui/views/mixins/unload-listener.es6.js
+++ b/shared/js/ui/views/mixins/unload-listener.es6.js
@@ -1,27 +1,41 @@
-const browserUIWrapper = require('./../../base/ui-wrapper.es6.js')
+import browser from 'webextension-polyfill'
 
 module.exports = {
     /**
-     * Collect any 'user-action' messages and pass them along
-     * when we call `browserUIWrapper.popupUnloaded` - this allows the background script
-     * to make a decision about whether to reload the current page.
+     * Passes any 'user-action' messages through a messaging connection to the
+     * extension's background, so that they can be used at the point this window
+     * is closed. This allows the background script to make a decision about
+     * whether to reload the current page.
      *
      * For example: If a user has manually disabled protections in the popup,
-     * we don't immediately reload the page behind - but instead we want to wait until the
-     * popup is closed.
+     * we don't immediately reload the page behind - but instead we want to wait
+     * until the popup is closed.
      *
      * @param store
      */
     registerUnloadListener: function (store) {
         /** @type {string[]} */
         const userActions = []
+        let connection = null
+
+        // Create a messaging connection to the background. If the connection is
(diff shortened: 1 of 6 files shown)


shared/js/ui/views/mixins/unload-listener.es6.js
```diff
@@ -1,27 +1,41 @@
...
         const userActions = []
+        let connection = null
+
+        // Create a messaging connection to the background. If the connection is
+        // broken by the background, the background ServiceWorker was restarted
+        // and the connection must be recreated + any earlier user actions
+        // resent.
+        const reconnect = () => {
+            connection = browser.runtime.connect({ name: 'popup' })
+            connection.onDisconnect.addListener(reconnect)
```
sammacbeth, line 27: Probably not a large issue, but do we need to clean up these listeners attached to closed connections?

sammacbeth (approved):
The looks good to me. I just had one comment about cleaning up dead listeners, but it's not a blocker.

shared/js/ui/views/mixins/unload-listener.es6.js
kzar: No I don't think so, since the connection is gone along with any of its listeners.

PR #1487, opened 2022-10-28
Avoid failing YouTube CTL tests when network is slow

kzar:
The YouTube Click to Load integration tests check a bunch of things, including that interacting with the YouTube Iframe API works correctly. Unfortunately, these tests fail sometimes when run inside a GitHub action since the test runner is too slow to load (or even request) the Iframe API script. When the script has failed to be requested in time, let's mark the test as pending (aka skipped) instead of declaring a failure.

1 - https://developers.google.com/youtube/iframe_api_reference

## Automated tests:
- [ ] Unit tests
- [x] Integration tests

###### Reviewer Checklist:

###### PR Author Checklist:


jonathanKingston (approved):
Seems fine to me, it's better than disabling entirely.

Thanks for looking into this, I was going to spend some time on it next so you saved me :)

PR #1527, opened 2022-11-14
MV3 click to load support

ladamski:
**Reviewer:**

## Description:

Add support for click to load in MV3
https://app.asana.com/0/72649045549333/1202839073279061/f
supporting PRs:
https://github.com/duckduckgo/content-scope-scripts/pull/189
https://github.com/duckduckgo/ddg2dnr/pull/37

## Automated tests:
- [ ] Unit tests
- [ ] Integration tests

###### Reviewer Checklist:

###### PR Author Checklist:


shared/js/background/before-request.es6.js
```diff
@@ -97,6 +97,17 @@ function handleRequest (requestData) {
     // don't have a tab instance for this tabId or this is a new requestId.
     if (requestData.type === 'main_frame') {
         if (!thisTab || thisTab.requestId !== requestData.requestId) {
+            if(thisTab) {
```
kzar, line 100: I'm suprised ESLint didn't catch this, perhaps the GitHub action isn't working?

shared/js/background/before-request.es6.js
```diff
@@ -97,6 +97,17 @@ function handleRequest (requestData) {
...
+                // upon navigation, remove any custom action session rules
+                const customRulesIds = []
+                for (const ruleSet in thisTab.customActionRules) {
+                    customRulesIds.push(...thisTab.customActionRules[ruleSet])
+                }
+                if (customRulesIds) {
```
kzar, line 106: I think an empty array is considered truthy in JavaScript. How about this?
```suggestion
                if (customRulesIds.length === 0) {
```

shared/js/background/before-request.es6.js
```diff
@@ -97,6 +97,17 @@ function handleRequest (requestData) {
     // don't have a tab instance for this tabId or this is a new requestId.
     if (requestData.type === 'main_frame') {
         if (!thisTab || thisTab.requestId !== requestData.requestId) {
+            if(thisTab) {
+                // upon navigation, remove any custom action session rules
```
kzar, line 101: Would be nice to explain the idea in this comment here a bit more, even if you only give CTL rules as an example.

shared/js/background/before-request.es6.js
```diff
@@ -97,6 +97,17 @@ function handleRequest (requestData) {
...
+                const customRulesIds = []
+                for (const ruleSet in thisTab.customActionRules) {
+                    customRulesIds.push(...thisTab.customActionRules[ruleSet])
+                }
+                if (customRulesIds) {
+                    chrome.declarativeNetRequest.updateSessionRules({ removeRuleIds: customRulesIds})
```
kzar, line 107:
```suggestion
                    chrome.declarativeNetRequest.updateSessionRules({ removeRuleIds: customRulesIds })
```

shared/js/background/before-request.es6.js
```diff
@@ -97,6 +97,17 @@ function handleRequest (requestData) {
     // don't have a tab instance for this tabId or this is a new requestId.
     if (requestData.type === 'main_frame') {
         if (!thisTab || thisTab.requestId !== requestData.requestId) {
+            if(thisTab) {
+                // upon navigation, remove any custom action session rules
+                const customRulesIds = []
```
kzar, line 102: IMO let's move this logic into your `custom-rules-manager.js` module as a function and import+call it here.

shared/js/background/classes/custom-rules-manager.js
```diff
@@ -0,0 +1,40 @@
...
+    generateDNRRule
+} from '@duckduckgo/ddg2dnr/lib/utils'
+const { getNextSessionRuleId } = require('../dnr-session-rule-id')
+
+function getInverseRules (action) {
+    const inverseCustomRules = JSON.parse(JSON.stringify(settings.getSetting('inverseCustomRules')))
+    return inverseCustomRules?.filter(rule => rule.customAction === action)
+}
+
+export function enableInverseRules (customAction, tabId) {
```
kzar, line 14: Please can you add a JSDoc comment for this? Then as a bonus tslint will check the input types are correct.

shared/js/background/classes/tab-state.js
```diff
@@ -32,6 +32,8 @@ export class TabState {
         this.trackers = {}
         /** @type {null | import('../events/referrer-trimming').Referrer} */
         this.referrer = null
+        /** @type {Object} */
```
kzar, line 35: Could you specify this type in more detail? So we avoid typos with the property names?

shared/js/background/declarative-net-request.js
```diff
@@ -327,6 +331,16 @@ export async function refreshUserAllowlistRules (allowlistedDomains) {
     await updateUserAllowlistRule(normalizedAllowlistedDomains)
 }
 
+export async function flushSessionRules () {
+    chrome.declarativeNetRequest.getSessionRules().then(rules => {
+        if(rules.length) {
+            console.log('removing session rules', ruleIds)
+            const ruleIds = rules.filter(r => isValidSessionId(r.id)).map(r => r.id)
```
kzar, line 338:
```suggestion
            const ruleIds = rules.map(({ id }) => id).filter(isValidSessionId)
```

shared/js/background/declarative-net-request.js
```diff
@@ -327,6 +331,16 @@ export async function refreshUserAllowlistRules (allowlistedDomains) {
     await updateUserAllowlistRule(normalizedAllowlistedDomains)
 }
 
+export async function flushSessionRules () {
+    chrome.declarativeNetRequest.getSessionRules().then(rules => {
+        if(rules.length) {
+            console.log('removing session rules', ruleIds)
```
kzar, line 337: Maybe remove this comment? Or if not, we should at least move this below where we define `ruleIds`.

shared/js/background/declarative-net-request.js
```diff
@@ -327,6 +331,16 @@ export async function refreshUserAllowlistRules (allowlistedDomains) {
     await updateUserAllowlistRule(normalizedAllowlistedDomains)
 }
 
+export async function flushSessionRules () {
```
kzar, line 334: IMO let's move this into `dnr-session-rule-id.js`. Perhaps we could rename stuff in that module a bit to make it clear the distinction between session rules with incremented ID and other session rules too?

shared/js/background/declarative-net-request.js
```diff
@@ -327,6 +331,16 @@ export async function refreshUserAllowlistRules (allowlistedDomains) {
...
+    chrome.declarativeNetRequest.getSessionRules().then(rules => {
```
kzar, line 335: Nit: Might as well use `await` here, since the function is `async` anyway.

shared/js/background/declarative-net-request.js
```diff
@@ -327,6 +331,16 @@ export async function refreshUserAllowlistRules (allowlistedDomains) {
...
+        if(rules.length) {
+            console.log('removing session rules', ruleIds)
+            const ruleIds = rules.filter(r => isValidSessionId(r.id)).map(r => r.id)
+            return chrome.declarativeNetRequest.updateSessionRules({ removeRuleIds: ruleIds})
```
kzar, line 339: Nit: Might as well `await` here too, so that the returned Promise only resolves when this is done. So far, it will resolve before then which could be a foot-gun in the future.

shared/js/background/dnr-session-rule-id.js
```diff
@@ -17,7 +22,7 @@ export async function setSessionRuleOffsetFromStorage () {
 }
 
 /**
- * Get the next unique session rule id to use when craeting session DNR rules
+ * Get the next unique session rule id to use when creating session DNR rules
```
kzar, line 25: Nice, thanks for fixing some typos :1st_place_medal:

shared/js/background/events.es6.js
```diff
@@ -69,6 +69,9 @@ async function onInstalled (details) {
             }
         }
 
+        // remove any orphaned session rules (can happen on extension update/restart)
```
kzar, line 72: Perhaps expand this comment, or better we could add a JSDoc comment for the function. Either way, we should explain the real idea of this somewhere.  (e.g. "We increment the rule IDs for some session rules, starting at STARTING_RULE_ID and keep a note of the next rule ID in session storage. Sometimes (e.g. on extension update) session storage can be cleared, while session rules are not. To avoid an inconsistent state, let's clear session rules on extension update too.")

shared/js/content-scripts/content-scope-messaging.js
```diff
@@ -10,6 +10,23 @@ function getSecret () {
 async function init () {
     const secret = await getSecret()
 
+    window.addEventListener('sendMessage', event => {
+        // MV3 message proxy for click to load
+        event.stopImmediatePropagation()
+        const detail = event && event.detail
+        if(!detail) {
```
kzar, line 17:
```suggestion
        if (!detail) {
```

shared/js/content-scripts/content-scope-messaging.js
```diff
@@ -10,6 +10,23 @@ function getSecret () {
 async function init () {
     const secret = await getSecret()
 
+    window.addEventListener('sendMessage', event => {
```
kzar, line 13: Is this message coming from our main_world script? If so, couldn't the website spoof such a message too? Perhaps we could use the secret in an event name to mitigate?

shared/js/content-scripts/content-scope-messaging.js
```diff
@@ -10,6 +10,23 @@ function getSecret () {
...
+        // MV3 message proxy for click to load
+        event.stopImmediatePropagation()
+        const detail = event && event.detail
+        if(!detail) {
+            console.warn('no details in sendMessage proxy', event)
+            return
+        }
+        chrome.runtime.sendMessage(detail, response => {
+            const msg = { type: 'update', detail: { func: detail.messageType, response } }
+            const stringifiedArgs = JSON.stringify(msg)
```
kzar, line 23: Doesn't look like `stringifiedArgs` is being used anywhere?

shared/js/background/before-request.es6.js
ladamski: 👍  Went through and cleaned up some lint errors in general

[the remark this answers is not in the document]
ladamski: This seems to work?  Good idea on the comment either way, esp if we run into issues later.

shared/js/background/before-request.es6.js
ladamski: 👍

shared/js/background/dnr-session-rule-id.js
ladamski: 👍

shared/js/background/before-request.es6.js
ladamski: Good catch! Fixed a few of these but missed this one

shared/js/background/before-request.es6.js
ladamski: 👍

shared/js/background/declarative-net-request.js
ladamski: Not sure I follow, `dnr-session-rule-id.js` seems to be focused on managing rule IDs, while `declarative-net-request.js` is about managing the rules themselves (eg. `toggleUserAllowlistDomain`) which is what `flushSessionRules` does?

shared/js/background/declarative-net-request.js
ladamski: That is handled by the caller (which is the right pattern IMHO, let the caller decide whether to await or resolve as promise)

shared/js/content-scripts/content-scope-messaging.js
ladamski: No the communication from the main world content scripts hasn't changed, it just required some additional proxying internally for MV3

shared/js/background/classes/tab-state.js
ladamski: I expanded it a bit but the keys are dynamically generated

shared/js/background/declarative-net-request.js
kzar: The caller will be returned a Promise either way, since this is an async function. The caller can wait (or not) for that Promise to resolve either way. My suggestion would just mean that the returned Promise would resolve to the result (if any) of updateSessionRules, instead of resolving to a Promise that in turn resolves to a Promise which in turn resolves to the result on updateSessionRules.

ladamski: Oops you're right, insofar I forgot to remove the `async`.  :)

I noticed we seem to have fallen a bit into the pattern of using `await` just for return values, which in turn requires the containing function itself to be async, which is a NOOP in practice and creates cognitive overhead.   `refreshUserAllowlistRules()` right above is an example of this.

Maybe this better illustrates my point?

```
async function go() {
    function a1() {
        return Promise.resolve(true)
    }

    function a2() {
        return a1()
    }

    function a3() {
        return a2()
    }

    console.log('go', a3())
    console.log('await go ', await a3())
}

async function gonuts() {
    async function a1() {
        return await Promise.resolve(true)
    }

    async function a2() {
        return await a1()
    }

    async function a3() {
        return await a2()
    }

    console.log('gonuts', a3())
    console.log('await gonuts', await a3())
}

go()
gonuts()
```
results in
```
go Promise { true }
gonuts Promise { <pending> }
await go  true
await gonuts true
```

So I think if we see patterns where we're just doing `async .... return await x()`, that should give us pause?

ladamski:
Ok should be good to go here (modulo merging https://github.com/duckduckgo/content-scope-scripts/pull/189 and updating dependencies accordingly).  

Local int & int-mv3 tests passing too.

ladamski:
Figured out the issue with the integration test (hopefully), the content scope script dependency was just pointing at the branch name which was cached, so I updated to the latest commit.  Once that branch is merged, we'll need to update this one to that release before merging.

ladamski:
Ok I spoke too soon.  A simple timeout obscured the error, but its still there.   Really not sure what's going on, but please try it locally and see if it reproduces?  Still passing consistently locally for me.

shared/js/content-scripts/content-scope-messaging.js
kzar: OK, so long as you're confident that websites can't just throw a "sendMessage" event on window to trick this logic then it sounds good to me.

shared/js/background/events.es6.js
```diff
@@ -41,19 +43,31 @@ async function onInstalled (details) {
...
+    }
+
     // Inject the email content script on all tabs upon installation (not needed on Firefox)
-    if (browserName !== 'moz') {
-        const tabs = await browser.tabs.query({})
-        for (const tab of tabs) {
-            // Ignore URLs that we aren't permitted to access
-            if (tab.url.startsWith('chrome://')) {
-                continue
+    // FIXME the below code throws an unhandled exception in MV3
```
kzar, line 54: Have you filed an issue for this? Cc @jdorweiler

shared/js/background/declarative-net-request.js
```diff
@@ -465,6 +470,24 @@ export async function ensureServiceWorkerInitiatedRequestException () {
...
+* We increment the rule IDs for some session rules, starting at STARTING_RULE_ID and
+* keep a note of the next rule ID in session storage. During extesion update/restarts
+* session storage is cleared, while session rules are not, which causes errors due to
+* session rule ID conflicts
+ * @return {Promise}
+ */
+
+export function flushSessionRules () {
+    return chrome.declarativeNetRequest.getSessionRules().then(rules => {
+        if (rules.length) {
```
kzar, line 484: Think we should check `ruleIds.length > 0` after filtering instead.

kzar (approved):
LGTM, but would be nice to get the integration tests passing.

shared/js/content-scripts/content-scope-messaging.js
kzar: In fact, could you test that? I am a little concerned this could be a security vulnerability, especially considering even the message type is taken from the event - if it's spoofed that would be bad.

shared/js/background/events.es6.js
jdorweiler: @ladamski looks like this is used for email autofill. Can you check that it's still working?

ladamski: We discussed this prior somewhere in the CTL project, trying to find a reference to it (Asana permitting).  It only runs on ext update/install.

ladamski: Regarding sendMessage, I think we're probably due for a bit of a small security pass through the CTL code.  We have multiple messaging channels in different browsers (incl Mac), and they have all changed a bit over time, so I'll file a separate task to test this more thoroughly.

jdorweiler: 👍 for autofill. Can you make a followup to make sure we test it after this.

jdorweiler:
Not sure about the failing test. I can't get it to fail locally.

kzar:
> Not sure about the failing test. I can't get it to fail locally.

Same. I suspect it might be the block list loading slowly, perhaps when we update the bundled block list this will be resolved. Also, Sam's WIP is adding a way to wait for loading a test block list which could help here.

Since we have a deadline here, do you want to disable the Facebook CTL integration tests as well here? I will look into re-enabling them afterwards.

shared/js/content-scripts/content-scope-messaging.js
kzar: OK, I have tested it and it seems to be a security vulnerability. Try going to https://testpages.kzar.co.uk/insecure-extension-messaging/ and the website tell the extension to open its options page + feedback page.

kzar: It turns out that already exists in the released extension, resolving this therefore.

shared/js/content-scripts/content-scope-messaging.js
```diff
@@ -10,6 +10,22 @@ function getSecret () {
 async function init () {
     const secret = await getSecret()
 
+    window.addEventListener('sendMessage', event => {
```
kzar, line 13: I was wondering why we need to add the listener again here, since it's included in content-scope-scripts. I looked into it, I think that's since we don't have a listener in [content-scope-scripts/inject/chrome-mv3.js](https://github.com/duckduckgo/content-scope-scripts/blob/main/inject/chrome-mv3.js) like we do in [content-scope-scripts/inject/chrome.js](https://github.com/duckduckgo/content-scope-scripts/blob/main/inject/chrome.js). I think it would be better to add this listener to content-scope-scripts instead.

kzar (approved):

PR #1543, opened 2022-11-17
Add test for non-blocking of iframe cookies when site allowlisted.

sammacbeth:
**Reviewer:**

## Description:
A bug in JS cookie protection caused iframe cookies to be blocked when protections were disabled for a site. This was fixed in #1534. This PR adds a test for this case.

## Automated tests:
- [ ] Unit tests
- [ ] Integration tests

###### Reviewer Checklist:

###### PR Author Checklist:

### DIFF
--- a/integration-test/background/storage.js
+++ b/integration-test/background/storage.js
@@ -143,5 +143,37 @@ describe('Storage blocking Tests', () => {
             await page.close()
             await teardown()
         })
+
+        it('does not block safe third party iframe JS cookies when protections are disabled', async () => {
+            // https://app.asana.com/0/1201614831475344/1203336793368587
+            const { page, bgPage, teardown } = await setup()
+            // add testPageDomain to the allowlist
+            await bgPage.evaluate(async (domain) => {
+                return new Promise((resolve, reject) => {
+                    chrome.storage.local.get('settings', ({ settings }) => {
+                        try {
+                            settings.allowlisted = {
+                                [domain]: true
+                            }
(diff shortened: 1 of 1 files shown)


integration-test/background/storage.js
```diff
@@ -143,5 +143,37 @@ describe('Storage blocking Tests', () => {
...
             await teardown()
         })
+
+        it('does not block safe third party iframe JS cookies when protections are disabled', async () => {
+            // https://app.asana.com/0/1201614831475344/1203336793368587
+            const { page, bgPage, teardown } = await setup()
+            // add testPageDomain to the allowlist
+            await bgPage.evaluate(async (domain) => {
+                return new Promise((resolve, reject) => {
+                    chrome.storage.local.get('settings', ({ settings }) => {
```
jonathanKingston, line 153: This is good, but the test will fail open if we're changing the storage layout which isn't ideal.

Could you file a follow up nit / fix to call the allowlist methods instead rather than manipulating the storage directly?

jonathanKingston (approved):
LGTM thanks!
One nit, either fix or file a follow up issue imo.

sammacbeth:
Thanks for raising that @jonathanKingston - I forgot we expose settings in the dev builds so it's easy to just call that API directly. See https://github.com/duckduckgo/duckduckgo-privacy-extension/pull/1543/commits/95b924f2e7c036914347b648f82ef161be4d4ae3

PR #1544, opened 2022-11-17
Avoid blocking ServiceWorker initiated requests for Chrome MV3

kzar:
We already avoid blocking ServiceWorker initiated requests in the MV2
extension, since each time we enabled the blocking of those we saw
major website breakage. In the future, we'd like to start blocking
those requests again, but we'll have to do so carefully.

Anyway, it's important that request blocking is consistent between MV2
and MV3 builds of the extension. So for now, let's also avoid blocking
ServiceWorker initiated requests for Chrome MV3 builds too.

Note: The way we do this, is to ignore requests with a tabId of
      -1 (aka requests not associated with a tab). This could also
      lead to other requests being ignored, but since that's
      consistent with the MV2 code-path[1] it seems the right approach.

1 - https://github.com/duckduckgo/duckduckgo-privacy-extension/blob/92222c85b04894a7f411cd844e2612b7a7d7c923/shared/js/background/before-request.es6.js#L85

**Reviewer:** @jdorweiler 

## Steps to test this PR:
1. Navigate to https://privacy-test-pages.glitch.me/privacy-protections/request-blocking and ensure ServiceWorker initiated requests aren't blocked, but other requests still are.
2. Smoke test that tracker blocking is still working generally.
3. Ensure the integration and unit tests are passing.

## Automated tests:
- [x] Unit tests
- [x] Integration tests

###### Reviewer Checklist:

###### PR Author Checklist:

### DIFF
--- a/shared/js/background/declarative-net-request.js
+++ b/shared/js/background/declarative-net-request.js
@@ -14,6 +14,7 @@ import {
     generateDNRRule
 } from '@duckduckgo/ddg2dnr/lib/utils'
 import {
+    SERVICE_WORKER_INITIATED_ALLOWING_PRIORITY,
     USER_ALLOWLISTED_PRIORITY
 } from '@duckduckgo/ddg2dnr/lib/rulePriorities'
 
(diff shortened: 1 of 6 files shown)


shared/js/background/declarative-net-request.js
```diff
@@ -429,7 +437,36 @@ export async function updateUserDenylist () {
...
+ * Ensure that the allowing rule for ServiceWorker initiated requests is
+ * enabled. Since the rule needs to be restricted to matching requests not
+ * associated with a tab (tabId of -1) and so must be a session rule. Session
+ * rules don't persist past a browsing session, so must be re-added.
+ * Note: Only exported for use by unit tests, do not call manually.
+ * @return {Promise}
+ */
+export async function ensureServiceWorkerInitiatedRequestException () {
+    const removeRuleIds = [SERVICE_WORKER_INITIATED_ALLOWING_RULE_ID]
+    const addRules = [generateDNRRule({
```
jdorweiler, line 450: Did you want to restrict this to `xmlhttprequest` or do we need to allow all types?

kzar: I'd prefer to do all types, since that matches the MV2 logic. (I agree that might include other requests, but I figure the main thing for now is to be consistent between MV2 and MV3.)

jdorweiler (approved):

PR #1551, opened 2022-11-21
Fix up service worker tab navigation issues

jonathanKingston:
**Reviewer:** @sammacbeth 

## Description:

On tab navigate we're losing navigations of pages that are fully served by service workers. We need to model the tab data from webNavigations instead.

See: https://app.asana.com/0/312629933896096/1203403741315745/f

## Steps to test this PR:

**STR**

1. Type into the address bar [browsersync.io](https://browsersync.io/) and enter
2. Type into the address bar [cnn.com](https://cnn.com/) and enter (in the same tab)
3. Type into the address bar [browsersync.io](https://browsersync.io/) and enter (in the same tab)

**Expected results**

- No trackers are loaded
- Some third parties

**Actual results**
- CNN trackers are shown

## Automated tests:
- [ ] Unit tests
- [ ] Integration tests

###### Reviewer Checklist:

###### PR Author Checklist:

### DIFF
--- a/shared/js/background/events.es6.js
+++ b/shared/js/background/events.es6.js
@@ -273,7 +273,14 @@ browser.webNavigation.onCommitted.addListener(details => {
     // ignore navigation on iframes
     if (details.frameId !== 0) return
 
-    const tab = tabManager.get({ tabId: details.tabId })
+    let tab = tabManager.get({ tabId: details.tabId })
+
+    const newTab = tabManager.create({ tabId: details.tabId, url: details.url })
+    // persist the last URL the tab was trying to upgrade to HTTPS
+    if (tab && tab.httpsRedirects) {
+        newTab.httpsRedirects.persistMainFrameRedirect(tab.httpsRedirects.getMainFrameRedirect())
+    }
+    tab = newTab
 
     if (!tab) return
 
--- a/shared/js/background/events.es6.js
+++ b/shared/js/background/events.es6.js
@@ -250,14 +250,9 @@ startup.ready().then(() => {
     )
 })
 
+const createdTargets = new Map()
 browser.webNavigation.onCreatedNavigationTarget.addListener(details => {
-    const currentTab = tabManager.get({ tabId: details.sourceTabId })
-    if (currentTab && currentTab.adClick) {
-        const newTab = tabManager.createOrUpdateTab(details.tabId, { url: details.url })
-        if (currentTab.adClick.shouldPropagateAdClickForNewTab(newTab)) {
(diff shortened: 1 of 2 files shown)


shared/js/background/events.es6.js
```diff
@@ -266,35 +266,46 @@ startup.ready().then(() => {
...
+const createdTargets = new Map()
 browser.webNavigation.onCreatedNavigationTarget.addListener(details => {
-    const currentTab = tabManager.get({ tabId: details.sourceTabId })
-    if (currentTab && currentTab.adClick) {
-        const newTab = tabManager.createOrUpdateTab(details.tabId, { url: details.url })
-        if (currentTab.adClick.shouldPropagateAdClickForNewTab(newTab)) {
-            newTab.adClick = currentTab.adClick.propagate(newTab.id)
-        }
-    }
+    createdTargets.set(details.tabId, details.sourceTabId)
```
sammacbeth, line 272: I assume this will be short-lived enough that in MV3 the service-worker lifecycle won't break this?

shared/js/background/events.es6.js
```diff
@@ -266,35 +266,46 @@ startup.ready().then(() => {
...
-browser.webNavigation.onCommitted.addListener(details => {
+browser.webNavigation.onBeforeNavigate.addListener(details => {
     // ignore navigation on iframes
     if (details.frameId !== 0) return
 
-    const tab = tabManager.get({ tabId: details.tabId })
-
-    if (!tab) return
+    const currentTab = tabManager.get({ tabId: details.tabId })
+    const newTab = tabManager.create({ tabId: details.tabId, url: details.url })
```
sammacbeth, line 289: After `browser.webNavigation.onBeforeNavigate` requests can still happen from the previous document (e.g. things triggered on `onbeforeunload` and things using the `Beacon` API). Those will now probably get attributed to the next page.

sammacbeth (commented):
Do you think it would be possible to cover this with a test-case?

shared/js/background/events.es6.js
jonathanKingston: Yeah it's literally between calls the risk is minimal enough I think we can accept it. Happy to switch to the session storage helper if you're worried?

jonathanKingston:
> Do you think it would be possible to cover this with a test-case?

I think so, I agree a test is worthwhile but as this isn't my TP I think we should make a quick wins task as I suspect I could lose a few hours into writing a test here. If that's ok with you?

sammacbeth (approved):

PR #1554, opened 2022-11-22
YouTube CTL part 2 - migration

franfaccin:
https://app.asana.com/0/0/1203325458386638/f

**Reviewer:**

## Automated tests:
- [ ] Unit tests
- [ ] Integration tests

###### Reviewer Checklist:

###### PR Author Checklist:

### DIFF
--- a/integration-test/background/click-to-load-youtube.js
+++ b/integration-test/background/click-to-load-youtube.js
@@ -8,6 +8,7 @@ const { setupAPISchemaTest } = require('../helpers/apiSchema')
 const testSite = 'https://privacy-test-pages.glitch.me/privacy-protections/youtube-click-to-load/'
 const youTubeStandardDomains = new Set(['youtu.be', 'youtube.com', 'www.youtube.com'])
 const youTubeNocookieDomains = new Set(['youtube-nocookie.com', 'www.youtube-nocookie.com'])
+const youTubeImageDomains = new Set(['i.ytimg.com'])
 
 let browser
 let bgPage
@@ -20,6 +21,7 @@ function summariseYouTubeRequests (requests) {
     const youTubeIframeApi = { checked: false, alwaysRedirected: true }
     const youTubeStandard = { blocked: 0, allowed: 0, total: 0 }
     const youTubeNocookie = { blocked: 0, allowed: 0, total: 0 }
+    const youTubeImage = { blocked: 0, allowed: 0, total: 0 }
 
     for (const request of requests) {
         if (request.url.href === 'https://www.youtube.com/iframe_api') {
@@ -40,6 +42,8 @@ function summariseYouTubeRequests (requests) {
             stats = youTubeStandard
         } else if (youTubeNocookieDomains.has(request.url.hostname)) {
             stats = youTubeNocookie
+        } else if (youTubeImageDomains.has(request.url.hostname)) {
+            stats = youTubeImage
         } else {
             continue
         }
@@ -55,7 +59,16 @@ function summariseYouTubeRequests (requests) {
         }
     }
 
-    return { youTubeIframeApi, youTubeStandard, youTubeNocookie }
+    return { youTubeIframeApi, youTubeStandard, youTubeNocookie, youTubeImage }
+}
+
+/**
+ * @param {import('../helpers/requests').LoggedRequestDetails[]} request
+ */
+function hasAutoplay (requests) {
+    return requests.some(request =>
+        youTubeNocookieDomains.has(request.url.hostname) && request.url.searchParams.autoplay === 1
+    )
 }
 
 describe('Test YouTube Click To Load', () => {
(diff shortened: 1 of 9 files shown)


integration-test/background/click-to-load-youtube.js
```diff
@@ -244,6 +258,152 @@ describe('Test YouTube Click To Load', () => {
...
+            )
+            await enablePreviewsButton.click()
+            await pageWait.forNetworkIdle(page)
+
+            const totalEmbeddedVideosPreviews = (await page.$$('div[id^=yt-ctl-preview-')).length
+
+            expect(totalEmbeddedVideosBlocked).toEqual(totalEmbeddedVideosPreviews)
+        }
+
+        {
```
kzar, line 303: Nit: Could combine this block with the above block?

integration-test/background/click-to-load-youtube.js
```diff
@@ -244,6 +258,152 @@ describe('Test YouTube Click To Load', () => {
...
+                youTubeIframeApi, youTubeStandard, youTubeNocookie, youTubeImage
+            } = summariseYouTubeRequests(pageRequests)
+
+            expect(youTubeIframeApi.checked).toBeFalse()
+            expect(youTubeIframeApi.alwaysRedirected).toBeTrue()
+            expect(youTubeStandard.blocked).toEqual(0)
+            expect(youTubeNocookie.blocked).toEqual(0)
+            expect(youTubeNocookie.allowed).toEqual(0)
+            expect(youTubeImage.blocked).toEqual(0)
+            expect(youTubeImage.allowed).toEqual(youTubeImage.total)
```
kzar, line 314: Maybe check that `youtubeImage.total` is > 0? Or do we somehow already know that at least one YouTube image has loaded?

shared/js/background/classes/tab-state.js
```diff
@@ -32,6 +32,8 @@ export class TabState {
         this.trackers = {}
         /** @type {null | import('../events/referrer-trimming').Referrer} */
         this.referrer = null
+        /** @type {boolean} */
+        this.ctlYouTubeAdded = false
```
kzar, line 36: Not quite clear what this variable represents, perhaps we could rename it to make it more obvious? (Is it to say that a CTL YouTube placeholder was shown for the page?)

shared/js/background/message-handlers.js
```diff
@@ -87,12 +87,21 @@ export function getTopBlockedByPages (options) {
...
 // Click to load interactions
-export async function initClickToLoad (config, sender) {
+export async function initClickToLoad (incomingConfig, sender) {
     await settings.ready()
     const tab = tabManager.get({ tabId: sender.tab.id })
 
     // Remove first-party entries.
     await startup.ready()
+
+    // Only return config for ClickToLoadConfig configured in storage
```
kzar, line 97: Could you explain the idea here? How could the configuration that this function receives be different from the configuration that's in storage?

Also, I find this implementation kind of obtuse. Am I right in saying we are creating a new Object, that has only the keys from the original Object that are also found in tdsStorage? If so, could we not just use a for loop instead?

```javascript
const config = {};
for (let entity in incomingConfig) {
  if (tdsStorage.ClickToLoadConfig[entity]) {
    config[entity] = incomingConfig[entity];
  }
}
```

Maybe I'm misunderstanding though?

shared/js/ui/models/submit-breakage-form.es6.js
```diff
@@ -25,13 +25,15 @@ module.exports = function (category) {
     const blockedTrackers = collectAllUrls(aggregationStats.blockAction.list)
     const surrogates = collectAllUrls(aggregationStats.redirectAction.list)
     const urlParametersRemoved = this.tab.urlParametersRemoved ? 'true' : 'false'
+    const ctlYouTubeAdded = this.tab.ctlYouTubeAdded ? 'true' : 'false'
```
kzar, line 28: Nit: Cool use Boolean again here?
```suggestion
    const ctlYouTubeAdded = Boolean(this.tab.ctlYouTubeAdded)
```

kzar (changes requested):
Looks good @franfaccin, in particular I'm happy to see the tests reenabled + some extra ones added for previews and auto-play. I've not tested this or looked at the `content-scope-scripts` change yet, so sorry if I've misunderstood anything.

shared/js/background/message-handlers.js
franfaccin: This configuration is now different because part of the CTP configuration is now hard-coded in [click-to-play.js](https://github.com/duckduckgo/content-scope-scripts/blob/main/src/features/click-to-play.js#L335) in C-S-S, meanwhile the surrogates and blocking rules still live in `social_ctp_configuration.json`. `tdsStorage.ClickToLoadConfig` data comes from `social_ctp_configuration.json` but the selectors and logic to find embedded content in the page and replace them lives in [click-to-play.js](https://github.com/duckduckgo/content-scope-scripts/blob/main/src/features/click-to-play.js#L335), so they can be mismatched. It happened with me when I started the migration, I was replacing the videos with the placeholders, but the videos were not being blocked until I updated `social_ctp_configuration.json`. Yes, it is very confusing and this makes for an obtuse solution from my side :/

But I'm not quite sure how to handle then. I could just comment out the [yt config](https://github.com/duckduckgo/content-scope-scripts/pull/196/files#diff-658f5706013e4033baa9dd3e8885fcbd34cf73ea331a3bfd34764a39b0d0186aR950) in `click-to-play.js` and it would me this not necessary anymore, but we'll have to sync both config releases.

Regarding the for loop suggestion, I agree, it's easier to read it.

integration-test/background/click-to-load-youtube.js
```diff
@@ -40,10 +43,14 @@ function summariseYouTubeRequests (requests) {
             stats = youTubeStandard
         } else if (youTubeNocookieDomains.has(request.url.hostname)) {
             stats = youTubeNocookie
+        } else if (youTubeImageDomains.has(request.url.hostname)) {
+            stats = youTubeImage
         } else {
             continue
         }
 
+        hasYouTubeAutoPlay = hasYouTubeAutoPlay || request.url.searchParams.get('autoplay') === '1'
```
kzar, line 52: Nit: Perhaps put this inside the `if (youTubeStandardDomains.has(request.url.hostname)) {` condition (or whichever is the correct one)?

Nit: I would prefer `if (request.url.searchParams.get('autoplay') === '1') hasYouTubeAutoPlay = true`, but up to you.

integration-test/background/click-to-load-youtube.js
```diff
@@ -103,7 +116,7 @@ describe('Test YouTube Click To Load', () => {
         }
```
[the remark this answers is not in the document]
kzar: OK, fair enough :+1:

integration-test/background/click-to-load-youtube.js
```diff
@@ -244,6 +256,145 @@ describe('Test YouTube Click To Load', () => {
...
+        const page = await browser.newPage()
+        const pageRequests = []
+        const clearRequests = await logPageRequests(page, pageRequests)
+
+        // Navigate to test page
+        await pageWait.forGoto(page, testSite)
+
+        // Once the user clicks to preview a video, should change from blocked state
+        // to preview state and request only YouTube Preview images
+        clearRequests()
```
kzar, line 271: I don't think clearing the requests here is correct. We want to know if the test page manages to sneak any requests through during load.

shared/js/background/classes/tab-state.js
kzar: Hmm, maybe just call it `ctlYouTube`? That seems more consistent with the other similar names above.

Maybe add a comment here to explain though? E.g. "True when at least one YouTube Click to Load placeholder was displayed for the tab."

shared/js/background/message-handlers.js
kzar: Ah right, I see. Yes, this is a mess currently, I'm sorry you're being stuck in the middle of it :(.

I think your implementation here is pretty solid now, let's just improve the names and comments to make the purpose clearer:
- How about we rename `incomingConfig` to `contentScopeConfig`?
- Maybe change the comment to say something like "The Click to Load configuration is in the process of being migrated, but is currently split between tdsStorage.ClickToLoadConfig (aka social_ctp_configuration.json), the block list (aka tds.json), and contentent-scope-scripts/features/click-to-play.js. Take care to remove any entries from the content-scope-scripts configuration that aren't also found in tdsStorage.ClickToLoadConfig - otherwise the feature can end up in an inconsistent state."

shared/js/background/wrapper.es6.js
```diff
@@ -9,6 +9,10 @@ export function getExtensionVersion () {
     return manifest.version
 }
 
+export const openExtensionPage = (path) => {
```
kzar, line 12: Nit: As this is a named top-level function, I would prefer to use `function` here instead of an arrow function.

shared/js/ui/models/submit-breakage-form.es6.js
```diff
@@ -25,13 +25,15 @@ module.exports = function (category) {
     const blockedTrackers = collectAllUrls(aggregationStats.blockAction.list)
     const surrogates = collectAllUrls(aggregationStats.redirectAction.list)
     const urlParametersRemoved = this.tab.urlParametersRemoved ? 'true' : 'false'
+    const ctlYouTubeAdded = Boolean(this.tab.ctlYouTubeAdded)
```
kzar, line 28: Nit: Is `Boolean` necessary here? I figure we know tab.ctlYouTubeAdded must be a bool right?

kzar (changes requested):

Rappid6974 (approved):

shared/js/ui/models/submit-breakage-form.es6.js
kzar: Ah, good catch!

kzar (approved):
LGTM

shared/js/background/message-handlers.js
```diff
@@ -145,12 +145,27 @@ export function getTopBlockedByPages (options) {
...
     await startup.ready()
+
+    /**
+     * The Click to Load configuration is in the process of being migrated, but it is currently split
+     * between tdsStorage.ClickToLoadConfig (aka social_ctp_configuration.json), the block list (aka tds.json),
+     * and content-scope-scripts/features/click-to-play.js.
+     * Here we take care to remove any entries from the content-scope-scripts configuration (aka contentScopeConfig)
+     * that aren't also found in tdsStorage.ClickToLoadConfig - otherwise the feature can end up in an inconsistent state.
+     */
+    const config = {}
```
ladamski, line 162: Not sure I follow the logic here.  The legacy config should be deprecated and no longer being updated, so should not override the embedded one.  There's no reason to really update the legacy one at this point?

franfaccin: From what I got when testing this, we still seem to use `tdsStorage.ClickToLoadConfig` to define which domains we block and surrogate rules. And this configuration is still being pulled from `social_ctp_configuration.json`. Without this, what was happening with YT CTL is that we were using the embedded configuration and adding placeholders to the page without actually blocking any of the YT requests. So I added this filtering here to make both configurations match. Am I missing something here along the way and is there a better way to link YT CTL in C-S-S and the list of domains we block?

shared/js/background/message-handlers.js
```diff
@@ -145,12 +145,27 @@ export function getTopBlockedByPages (options) {
...
+    for (const entity in contentScopeConfig) {
+        if (tdsStorage.ClickToLoadConfig[entity]) {
+            config[entity] = contentScopeConfig[entity]
+        }
+    }
+
```
franfaccin, line 168: Remove this in favor of the tds config to be rolled out soon.

ladamski (approved):

shared/js/background/message-handlers.js
kzar: I don't fully understand this, would be nice to add context to the above comment @ladamski so it's explained properly. But I know you two synced on it, so will click Approve again. The changes otherwise LGTM.

kzar (approved):

shared/js/background/message-handlers.js
franfaccin: Update: @ladamski and I synced and he explained that `social_ctp_configuration.json` will be deprecated soon in favor of using rules coming from `tds`. So by the time YT CTL releases, live configuration won't be used anymore for this and we can sync landing the changes to enable YT CTL in both `tds` and `content-scope-scripts` in the same release. So this filtering won't be necessary anymore.

shared/js/background/message-handlers.js
franfaccin: Good point @kzar! I just left some notes during the call so I didn't forget what needed to be updated, but this one would benefit from more context indeed. I've added an update in the original thread https://github.com/duckduckgo/duckduckgo-privacy-extension/pull/1554#discussion_r1038052473

shared/js/background/message-handlers.js
Rappid6974: []()

Rappid6974: > # - [ ] - **_@~~``~~ _**

PR #1565, opened 2022-11-28
MV3 third-party cookie blocking

sammacbeth:
**Reviewer:**

## Automated tests:
- [ ] Unit tests
- [ ] Integration tests

###### Reviewer Checklist:

###### PR Author Checklist:


shared/js/background/declarative-net-request.js
```diff
@@ -25,7 +26,8 @@ export const SETTING_PREFIX = 'declarative_net_request-'
 // of removing rules associated with different configurations.
 const ruleIdRangeByConfigName = {
     tds: [1, 10000],
-    config: [10001, 20000]
+    config: [10001, 20000],
```
kzar, line 29: I know these changes are subject to change, but currently the new rule range will clash with the constant IDs (e.g. `USER_ALLOWLIST_RULE_ID`) below. This seems like a "foot-gun waiting to happen" (my bad), so how about adding a "_RESERVED" range of 1 to 1000, and tweaking those constant IDs to be within that?
```javascript
const ruleIdRangeByConfigName = {
    _RESERVED: [1, 1000],
    tds: [1001, 11000],
    config: [11001, 21000],
    combined: [21001, 30000]
}
```

integration-test/background/storage.js
```diff
@@ -7,12 +10,22 @@ const testPageDomain = 'privacy-test-pages.glitch.me'
 const thirdPartyDomain = 'good.third-party.site'
 const thirdPartyTracker = 'broken.third-party.site'
 
+async function loadTestTds (bgPage, tdsFile) {
```
kzar, line 13: Nice idea, maybe put this in with `integration-test/helpers/testConfig.js`?

shared/js/background/message-handlers.js
```diff
@@ -361,13 +361,26 @@ export function getListContents (list) {
     }
 }
 
-export function setListContents ({ name, value }) {
+/**
+ * Manually override the value of a list
+ * @param {{ name: string, value: object}} list value
+ */
+export async function setListContents ({ name, value }) {
```
kzar, line 368: Maybe we should give this a comment or name that makes it clear it's only for testing purposes? E.g. `setListContentsForTest` or something?

sammacbeth: This is also used by the `list-editor` debug page.

shared/js/background/declarative-net-request.js
```diff
@@ -217,7 +220,7 @@ async function updateExtensionConfigRules (etag = null, configValue = null) {
 
     if (!configValue) {
         await tdsStorage.ready('config')
-        configValue = await tdsStorage.getListFromLocalDB('config')
+        configValue = await tdsStorage.config
```
sammacbeth, line 223: @kzar This line was causing issues in the tests, where we're loading a config and then updating the allowlist in quick succession. In this case we'd get invalid data from the DB, because the data didn't get flushed yet, causing the config rule generation to throw. I hope it's OK to just load the live version of the config here, or was there another reason to go to the DB here?

shared/js/background/declarative-net-request.js
sammacbeth: > and tweaking those constant IDs to be within that?

You mean the allowlist IDs should be in 1-1000?

shared/js/background/declarative-net-request.js
```diff
@@ -262,36 +279,42 @@ async function updateExtensionConfigRules (etag = null, configValue = null) {
...
-            supportedSurrogates,
-            '/web_accessible_resources/',
-            chrome.declarativeNetRequest.isRegexSupported,
-            ruleIdStart + 1
-        )
-
-        await updateConfigRules(configName, latestState, ruleset, matchDetailsByRuleId)
+    const extensionVersion = browserWrapper.getExtensionVersion()
+    // Run an async lock on all blocklist updates so the latest update is always processed last
+    ruleUpdateLock = ruleUpdateLock.then(async () => {
```
sammacbeth, line 284: This implements a serial queue for updates, so all changes to tds and config must be completed before the next update is processed. This prevents out of order calls to `updateDynamicRules` leaving the rules in an arbitrary state.

shared/js/background/events/3p-tracking-cookie-blocking.js
```diff
@@ -32,13 +36,16 @@ function shouldBlockHeaders (request, tab, requestIsTracker) {
  * @returns {{responseHeaders: Array<{name: string, value:string}>} | undefined}
  */
 function dropTracking3pCookiesFromResponse (request) {
+    if (request.type === 'main_frame' || request.responseHeaders.findIndex(header => header.name.toLowerCase() === 'set-cookie') === -1) {
```
kzar, line 39: Sounds good to me to return here if it's not a main_frame request.

I'm not sure either way about checking for the set-cookie header here though. Maybe that would be worse than first checking the tab is known and the request is considered a tracker. Equally maybe it's better, up to you anyway.

kzar (changes requested):

unit-test/background/declarative-net-request.js
```diff
@@ -464,7 +464,7 @@ describe('declarativeNetRequest', () => {
         expectState([])
     })
 
-    it('User denylisting updates', async () => {
+    fit('User denylisting updates', async () => {
```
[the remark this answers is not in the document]
sammacbeth: Fit only runs that test - I removed it and found another test failure...

browsers/chrome-mv3/manifest.json
```diff
@@ -69,7 +69,8 @@
         "storage",
         "tabs",
         "webRequest",
-        "webNavigation"
+        "webNavigation",
+        "cookies"
```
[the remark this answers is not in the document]
sammacbeth: No.

shared/js/background/declarative-net-request.js
sammacbeth: The issue I had is that if the tds list update happens twice in quick succession, it is not guaranteed that your dynamic rules are those generated from the last list update. This could be because when the TDS rule generation hits a regex rule, it has to do an async call. That tick then allows the next update that sitting on the event queue to run and generate rules before the first update finishes.

shared/js/background/events.es6.js
```diff
@@ -425,20 +411,26 @@ if (manifestVersion === 2) {
...
-            extraInfoSpecSendHeaders
-        )
-    })
+browser.webRequest.onBeforeSendHeaders.addListener(
+    dropTracking3pCookiesFromRequest,
+    { urls: ['<all_urls>'] },
+    extraInfoSpecSendHeaders
+)
+
+browser.webRequest.onHeadersReceived.addListener(
```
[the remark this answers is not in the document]
sammacbeth: It's possible to get those exceptions, but unlikely. It looks like we're already throwing exceptions from `blockHandleResponse` due to this anywhere. Is the fix to silently suppress the error when tds is not loaded? Showing an error (or warning) at least allows us to see the startup issue.

shared/js/background/events/3p-tracking-cookie-blocking.js
sammacbeth: Looking for the set-cookie header is an array scan of max. 5-10 values. It also will likely excludes more than the other two checks, so doing it first means we have fewer tab and tracker lookups to do in total. My intuition is that this is the most performant check order.

shared/js/background/declarative-net-request.js
kzar: Ah right, got it.

shared/js/background/events.es6.js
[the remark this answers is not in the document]
kzar: Alright, fair enough. Not the biggest issue for now, let's punt it.

unit-test/background/declarative-net-request.js
```diff
@@ -321,7 +321,7 @@ describe('declarativeNetRequest', () => {
             config: {
                 etag: TEST_ETAGS[2], extensionVersion: TEST_EXTENION_VERSIONS[0]
             }
-        }, 2)
+        }, 3)
```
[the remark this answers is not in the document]
sammacbeth: I updated the tests so combined updates also run correctly.

kzar (changes requested):
I think we should change the rule IDs (see comment below), but everything else looks great now :+1:

kzar (approved):
LGTM

PR #1585, opened 2022-12-05
ATB parameter in MV3

jdorweiler:
**Reviewer:**

## Description:

Adds ATB to search URL. 

## Steps to test this PR:
1. From a new Chrome profile, install the extension
2. Do a search from URL bar, atb should be added to the URL
3. Check other DDG pages, they should not have atb added to them

###### Reviewer Checklist:

###### PR Author Checklist:

### DIFF
--- a/integration-test/background/atb.js
+++ b/integration-test/background/atb.js
@@ -104,7 +104,7 @@ describe('install workflow', () => {
         })
         it('should get its ATB param from the success page when one is present', async () => {
             if (manifestVersion === 3) {
-                return 
+                return
             }
 
             // open a success page and wait for it to have finished loading
--- a/integration-test/background/atb.js
+++ b/integration-test/background/atb.js
@@ -5,6 +5,8 @@ const pageWait = require('../helpers/pageWait')
 // eslint-disable-next-line no-shadow
 const fetch = (...args) => import('node-fetch').then(({ default: fetch }) => fetch(...args))
 
+const manifestVersion = harness.getManifestVersion()
+
 let browser
 let bgPage
 let requests
@@ -68,6 +70,10 @@ describe('install workflow', () => {
         })
 
         it('should get its ATB param from atb.js when there\'s no install success page', async () => {
+            if (manifestVersion === 3) {
+                return
+            }
+
             // try get ATB params
             await bgPage.evaluate(() => globalThis.dbg.atb.updateATBValues())
             await backgroundWait.forSetting(bgPage, 'extiSent')
@@ -97,6 +103,10 @@ describe('install workflow', () => {
             expect(numExtiCalled).toEqual(1)
         })
         it('should get its ATB param from the success page when one is present', async () => {
+            if (manifestVersion === 3) {
+                return 
+            }
+
             // open a success page and wait for it to have finished loading
             const successPage = await browser.newPage()
             await pageWait.forGoto(successPage, 'https://duckduckgo.com/?natb=v123-4ab&cp=atbhc')
@@ -141,6 +151,10 @@ describe('install workflow', () => {
         })
 
         it('should retreive stored ATB value on reload', async () => {
+            if (manifestVersion === 3) {
+                return
+            }
+
             // set an ATB value from the past
             const pastATBValue = 'v123-1'
             await bgPage.evaluate((pagePastATBValue) => globalThis.dbg.settings.updateSetting('atb', pagePastATBValue), pastATBValue)
--- a/integration-test/background/atb.js
+++ b/integration-test/background/atb.js
@@ -71,6 +71,8 @@ describe('install workflow', () => {
 
         it('should get its ATB param from atb.js when there\'s no install success page', async () => {
             if (manifestVersion === 3) {
+                // TODO: Can't see requests from service workers. Re-enable once this is fixed
+                // https://github.com/puppeteer/puppeteer/issues/2781
                 return
             }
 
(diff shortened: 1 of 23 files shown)


shared/js/background/declarative-net-request.js
```diff
@@ -465,6 +473,43 @@ export async function ensureServiceWorkerInitiatedRequestException () {
...
+                    addOrReplaceParams: [{ key: 'atb', value: atb }]
+                }
+            }
+        },
+        resourceTypes: ['main_frame'],
+        requestDomains: ['duckduckgo.com'],
+        regexFilter
+    })
+
+    if (atbRule?.id === ATB_PARAM_RULE_ID) {
```
kzar, line 505: Why do we have this check? Won't `atbRule.id` always equal `ATB_PARAM_RULE_ID` at this point, or am I missing something?

shared/js/background/declarative-net-request.js
```diff
@@ -465,6 +473,43 @@ export async function ensureServiceWorkerInitiatedRequestException () {
...
+ * @param {string} atb
+ * @param {object} regexFilter
+ */
+export async function setOrUpdateATBdnrRule (atb, regexFilter) {
+    if (!(atb && regexFilter)) {
+        return
+    }
+
+    // format regex for json and dnr
+    regexFilter = regexFilter.toString().replace(/\\/g, '\\').replace(/^\//, '').replace(/\/$/, '')
```
kzar, line 487:
```suggestion
    regexFilter = regexFilter.source
```

shared/js/background/declarative-net-request.js
```diff
@@ -465,6 +473,43 @@ export async function ensureServiceWorkerInitiatedRequestException () {
     })
 }
 
+/**
+ * Creates a DNR rule for ATB parameters
+ * @param {string} atb
+ * @param {object} regexFilter
```
kzar, line 479:
```suggestion
 * @param {RegExp} regexpAboutPage
 *     Regular expression that matches URLs that we need to add the atb query
 *     parameter to, as returned by `ATB.getRegExpAboutPage()`.
```

kzar (changes requested):
Can we enable the ATB integration tests for MV3 too?

shared/js/background/declarative-net-request.js
jdorweiler: It's coming from ddg2dnr. It's our library, but still and external library.

shared/js/background/atb.es6.js
```diff
@@ -23,6 +24,10 @@ const ATB = (() => {
     const ddgAtbURL = 'https://duckduckgo.com/atb.js?'
 
     return {
+        getRegExpAboutPage () {
```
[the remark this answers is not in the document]
jdorweiler: done

shared/js/background/atb.es6.js
```diff
@@ -199,6 +210,39 @@ const ATB = (() => {
                 !domain.match(regExpSoftwarePage)
         },
 
+        /**
+        * Creates a DNR rule for ATB parameters
+        * @param {string} atb
+        */
+        setOrUpdateATBdnrRule: (atb) => {
+            if (!(manifestVersion === 3 && atb)) {
```
kzar, line 218: Nit:
```suggestion
            if (!atb || manifestVersion !== 3)) {
```

integration-test/background/atb.js
```diff
@@ -141,6 +151,10 @@ describe('install workflow', () => {
         })
 
         it('should retreive stored ATB value on reload', async () => {
+            if (manifestVersion === 3) {
```
kzar, line 154: There's a FIXME for MV3 comment below, I wonder if you addressed that we could leave this test case enabled?

integration-test/background/atb.js
```diff
@@ -68,6 +70,10 @@ describe('install workflow', () => {
         })
 
         it('should get its ATB param from atb.js when there\'s no install success page', async () => {
+            if (manifestVersion === 3) {
```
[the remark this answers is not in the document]
kzar: I don't understand why these test cases don't work for MV3, are you sure it's a problem with the tests and not the feature itself? If so, could you add a comment to explain?

shared/js/background/declarative-net-request.js
kzar: I'm probably missing something, but it seems like we create the `atbRule` Object directly above and we're sure the "id" property will equal `ATB_PARAM_RULE_ID`, so I don't understand the need for the check . Do you mean it's in case `generateDNRRule` isn't behaving like we expect?

kzar (changes requested):
Looks good % a couple of nits. The main thing that concerns me is why we needed to disable some of the integration tests, are you sure the tests are broken and not the feature itself?

integration-test/background/atb.js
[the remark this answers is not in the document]
jdorweiler: We're looking that a request went out below in line 87 for the test to pass. There are never any requests to the test always fails, but you can run `dbg.setting.removeSetting('extiSent')` and `dbg.atb.updateATBValues()` manually then check the network log to see that a request went out. 

https://github.com/puppeteer/puppeteer/issues/2781 sounds like we can't see requests coming from a service worker.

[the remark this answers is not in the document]
jdorweiler: There are a few testing that are looking at requests. I turned all of those cases off for MV3.

integration-test/background/atb.js
jdorweiler: I tried, the test just errors. I'm not sure that `runtime.reload()` works.

shared/js/background/declarative-net-request.js
jdorweiler: It works fine now. I suppose this case is covered in ddg2dnr tests anyway?

jdorweiler: I removed it.

kzar (approved):
LGTM

PR #1607, opened 2022-12-15
Fix test that depends on the number of enabled features in the config

sammacbeth:
We just turned on a new feature in the config, which caused this test to start failing.


GioSensation (approved):
LGTM. 🚀

PR #1634, opened 2023-01-09
Fix building with local content-scope-scripts/tracker-surrogates

kzar:
Our build scripts include the dubious step of copying the
content-scope-scripts and tracker-surrogates node modules into another
location before use. It turns out that the source paths need a
trailing "/" for that to work correctly when testing with local
versions of those modules.

**Reviewer:** @sammacbeth 

## Steps to test this PR:
1. Ensure builds still work generally.
2. Try changing `content-scope-scripts` or `tracker-surrogate dependencies` in packages.json to a local copy, then try the build again (after clearing `shared/content-scope-scripts` and `shared/tracker-surrogates`.

## Automated tests:
- [ ] Unit tests
- [ ] Integration tests

###### Reviewer Checklist:

###### PR Author Checklist:

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,7 +9,7 @@
     "scripts": {
         "install-ci": "npm ci --ignore-scripts && npm rebuild puppeteer && npm run postinstall",
         "postinstall": "npm run copy-modules && export PATH=\"$PATH:$PWD/node_modules/chromedriver/bin\" && npm run setup-test-int",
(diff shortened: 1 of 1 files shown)


jonathanKingston:
These should be directories in `shared/` what's the actual problem? because I don't think this fixes it.

kzar:
> These should be directories in `shared/` what's the actual problem? because I don't think this fixes it.

I tried to describe the problem in the commit and issue description, and this change did fix it. When testing with local copies of the `content-scope-scripts` and `tracker-surrogates` dependencies, the `copy-modules` step failed for me since Node.js created symlinks for those directories.

sammacbeth (changes requested):
I spoke too soon - this fixes the command, but the data is not copied to the correct location! You are now copying the contents of the `content-scope-scripts` folder to `shared`, rather than the folder itself.

PR #1635, opened 2023-01-11
feat: Tracker stats as a data provider to a New Tab page

shakyShane:
https://app.asana.com/0/0/1203587853222875/f

## Description:
This PR adds support for the tracker stats to be communicated to the New Tab Page.

### New classes/features

- `NewTabTrackerStats` - `shared/js/background/newtab-tracker-stats.js` - this is interface between this feature and the extension. All handlers/side-effects live here
   - This gets instantiated in startup, tries to restore previous data, and then registers all handlers 
- `TrackerStats` - `shared/js/background/classes/tracker-stats.js`
   - a container for storing the data in memory, with helpers for evicting expired entries etc
- A handler for redirecting `https://duckduckgo.com/tracker-stats.html` to our new HTML file (see next)
- A new HTML file that gets served in the invisible iframe in the newtab page
- A new JS entry/build file `newtab.js`
    - this is what the iframe will request
    - it contains all logic for proxying data between the iframe and the new tab page

### New storage

- added storage in the following format
- we prune storage every 5 min to ensure expired timestamps are evicted
- there are rules around which company names we store, it was all discussed/decided in the [original privacy review](https://app.asana.com/0/69071770703008/1203324001603601/f)
```json
{
  "trackerStats": {
    "stats": {
      "totalCount": 12,
      "entries": {
        "Google": [1673438914778],
        "Facebook": [1673438914778, 1673438914778, 1673438914792]
      }
    }
  }
}
```

### FAQ

- **Question:** Why not use the existing `Companies` data that we track/store?

**Answer:** I did try this, but I learned quickly that it has completely difference concerns to what this feature does. `Companies` tracks every entities, tracker or not, and builds a list of top sites etc over time. 

`New Tab Tracker Stats` (this feature) is a very narrow and different scope. We keep a simple running total of every blocking event that occurred along with timestamps for the Top 100 companies - but we only store for an hour. There's no overlap in features actually, even though it might seem like it initially.

The goal is to introduce this feature without tying it into any existing ones. Separate storage, separate handlers/alarms etc, all packaged into the `NewTabTrackerStats` class

- **Question**: Why have the complexity of the iframe proxying data to the New Tab Page?

**Answer**: Please see the [Asana Task](https://app.asana.com/0/0/1203634805204524/f)

### Outstanding questions

- **Question:** should we allow this feature to be remotely disabled?

Screenshot 2023-01-19 at 19 21 15

## Automated tests:
- [x] Unit tests
- [ ] Integration tests (TODO)

###### Reviewer Checklist:

###### PR Author Checklist:


browsers/chrome/manifest.json
```diff
@@ -23,6 +23,9 @@
             "public/js/background.js"
         ]
     },
+    "chrome_url_overrides" : {
+        "newtab": "html/redirect.html"
```
shakyShane, line 27: This is only here to help testing locally, it will be removed

shared/html/redirect.html
```diff
@@ -0,0 +1,15 @@
...
+<head>
+    <meta charset="UTF-8">
+    <meta name="viewport"
+          content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0">
+    <meta http-equiv="X-UA-Compatible" content="ie=edge">
+    <title>Redirecting...</title>
+</head>
+<body>
+<script>
+    console.log('should not get here...');
```
shakyShane, line 12: This file won't exist in the final build/PR, it's just here to aid testing

shared/js/background/newtab-tracker-stats.js
```diff
@@ -0,0 +1,348 @@
...
+     */
+    register () {
+        /**
+         * This listener will redirect the request for tracker-stats.html
+         * on the new tab page to our own HTML file under `web_accessible_resources`
+         */
+        browser.webRequest.onBeforeRequest.addListener(redirectIframeForTrackerStatsMV2,
+            {
+                // todo(Shane): limit this to only the urls we care about
+                urls: ['<all_urls>']
```
sammacbeth, line 68: For the final version you should be able to cut down the triggering here quite a bit:
```suggestion
                urls: [`https://${constants.trackerStats.allowedOrigin}/*`],
                types: ['sub_frame']
```

shared/js/background/newtab-tracker-stats.js
```diff
@@ -0,0 +1,348 @@
...
+            ['blocking']
+        )
+
+        /**
+         * Handle runtime messages sent from the new tab page
+         */
+        browser.runtime.onMessage.addListener((event, sender) => {
+            if (sender.id !== browserWrapper.getExtensionId()) return
+            this.handleIncomingEvent(event)
+        })
```
sammacbeth, line 79: Is there a potential conflict here with the existing central message listener? https://github.com/duckduckgo/duckduckgo-privacy-extension/blob/develop/shared/js/background/events.es6.js#L375

Particularly in the case where a return value is expected there, could this method returning undefined break the flow?

shared/js/background/classes/tracker-stats.js
```diff
@@ -0,0 +1,192 @@
+import z from 'zod'
+
+/**
+ * A single place to collect information about the number of requests
+ * we have blocked, grouped by the owner's display name.
+ */
+export class TrackerStats {
+    maxAgeMs = 1000 * 60 * 60 // 1hr
```
sammacbeth, line 8: Do we only keep tracker counts for 1 hour? Won't this then be mostly empty most of the time for light users, and at the beginning of a new session?

shared/js/background/newtab-tracker-stats.js
```diff
@@ -0,0 +1,348 @@
...
+         */
+        emitter.on(TrackerBlockedEvent.eventName, (event) => {
+            if (!(event instanceof TrackerBlockedEvent)) return
+            this.record(event.companyDisplayName)
+        })
+
+        /**
+         * Assign the entities from the `tds` data
+         */
+        this.assignTopCompanies(tdsStorage.tds.entities)
```
sammacbeth, line 105: Would it be worth listening for tds updates here too? If a new version of the blocklist comes with new entity names we may need to update the top 100, and maybe also migrate old data.

shared/js/background/newtab-tracker-stats.js
```diff
@@ -0,0 +1,348 @@
...
+         * 1) the `displayName` of the company is in our top100Companies list
+         * 2) the `displayName` of the company is NOT in our `excludedCompanies` list
+         */
+        if (this.top100Companies?.has(displayName) && !constants.trackerStats.excludedCompanies.includes(displayName)) {
+            this.stats.increment(displayName, timestamp)
+        } else {
+            /**
+             * Otherwise just increase the 'unknown' count
+             */
+            this.stats.increment(NewTabTrackerStats.unknownCompanyKey, timestamp)
```
sammacbeth, line 146: Shouldn't we call this 'others' instead of 'unknown'? Should we differentiate between actual 'unknown' trackers (where `findParentDisplayName` returns 'unknown'), and known by not top100 trackers?

shared/js/background/newtab-tracker-stats.js
```diff
@@ -0,0 +1,348 @@
...
+    debounceTimers = {}
+
+    /**
+     * @param {string} name
+     * @param {number} timeout
+     * @param {() => unknown} fn
+     */
+    debounced (name, timeout, fn) {
+        clearTimeout(this.debounceTimers[name])
+        this.debounceTimers[name] = setTimeout(fn, timeout)
```
sammacbeth, line 309: This debounce implementation seems to delay indefinitely when new events come in regularly. e.g. if there is a tracker that is making a request on a 500ms interval, we will keep debouncing that update, blocking the update of the new tab page until that tab is closed.

shared/js/background/newtab-tracker-stats.js
```diff
@@ -0,0 +1,348 @@
...
+    _publish (reason = 'unknown') {
+        if (this._debug) {
+            console.info(`sending new tab data because: ${reason}`)
+        }
+        /** @type {import("zod").infer<typeof import("../newtab/schema").dataMessage>} */
+        const msg = {
+            messageType: outgoing.newTabPage_data,
+            options: this.toDisplayData()
+        }
+        chrome.runtime.sendMessage(msg)
```
sammacbeth, line 230: We should check other places that will receive this message and make sure they discard it correctly. I see, for example, [this code](https://github.com/duckduckgo/duckduckgo-privacy-extension/blob/develop/shared/js/content-scripts/content-scope-messaging.js#L41-L45) which looks like it might forward these messages blindly.

shared/js/background/utils.es6.js
```diff
@@ -437,3 +454,13 @@ export async function reloadCurrentTab () {
         browser.tabs.reload(tab.id)
     }
 }
+
+export function perf (name) {
```
sammacbeth, line 458: Is this being used anywhere?

shared/js/background/before-request.es6.js
```diff
@@ -321,6 +346,10 @@ function blockHandleResponse (thisTab, requestData) {
             // @ts-ignore
             Companies.add(tracker.tracker.owner)
 
+            // publish the parent's display name only
+            const displayName = utils.findParentDisplayName(requestData.url)
+            emitter.emit(TrackerBlockedEvent.eventName, new TrackerBlockedEvent({ companyDisplayName: displayName }))
```
sammacbeth, line 351: Should we skip sending stats for incognito pages maybe?

sammacbeth (commented):
I've done a first pass with a bunch of questions. I think it's worth at least considering the MV3 implementation here too, as it might end up changing some of the MV2 code. I'll also follow up on some issues in Asana.

shared/data/constants.js
```diff
@@ -143,5 +143,22 @@ module.exports = {
...
     },
-    supportedLocales: ['cimode', 'en'] // cimode is for testing
+    supportedLocales: ['cimode', 'en'], // cimode is for testing
+    trackerStats: /** @type {const} */({
+        allowedOrigin: 'https://use-devtesting13.duckduckgo.com',
+        // allowedOrigin: 'https://eun-sosbourne1.duckduckgo.com',
+        allowedPathname: 'tracker-stats.html',
+        redirectTarget: 'html/tracker-stats.html',
+        /** @type {ReadonlyArray<string>} */
+        excludedCompanies: ['ExoClick'],
```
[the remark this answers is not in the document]
shakyShane: By manually checking every link in our current top 100 and identifying this as a tracker-network that's 99% used on NSFW sites. PLease see https://app.asana.com/0/0/1203708901942551/f

shared/js/background/classes/tracker-stats.js
shakyShane: Yep - that's the intention though. :)

shared/js/background/before-request.es6.js
shakyShane: Sounds good, do we have any examples of doing that elsewhere?

shared/js/background/newtab-tracker-stats.js
```diff
@@ -0,0 +1,412 @@
...
+        types: ['sub_frame']
+    },
+    ['blocking'])
+}
+
+function mv3Redirect () {
+    const targetUrl = chrome.runtime.getURL(constants.trackerStats.redirectTarget)
+    const incomingUrl = new URL(constants.trackerStats.allowedPathname, constants.trackerStats.allowedOrigin)
+    chrome.declarativeNetRequest.updateDynamicRules({
+        removeRuleIds: [100000000001],
```
shakyShane, line 394: Is there a better way to do this?

shakyShane: CC: @kzar 

can you suggest a way to insert this dynamic rule at runtime, without needing this 'magic number'?

shared/js/background/before-request.es6.js
sammacbeth: You can use the `incognito` property on `requestData` to check if the request is from a private tab. (When looking for a docs reference, it seems that this attributed is not in the Chrome docs, but still on MDN, so maybe check that this still exists https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/webRequest/onBeforeRequest).

shared/js/background/newtab-tracker-stats.js
shakyShane: I'm moving to Ports to solve this :)

shared/js/background/before-request.es6.js
shakyShane: From what I can see, our handlers do not see any requests from incognito tabs anyway? Unless there's some special way to test this that you know of?

shakyShane: Yeah it seems like our extension cannot be enabled in incognito mode anyway, so this is probably not an issue?

shared/js/background/newtab-tracker-stats.js
shakyShane: thanks for catching this, @sammacbeth!

I've moved to a trailing throttle implementation instead, so that events will always occur at most once during a time window.

shared/js/background/newtab-tracker-stats.js
shakyShane: Yep, done :)

shakyShane:
@sammacbeth @kzar changes since the last review:

- add mv3 support Shane Osbourne 687d4745
  - ^ also includes the switch to Ports
- Switch to throttling instead of debouncing afb85e88
- Switch from 'unknown' to 'others' to track companies without a name 480f4202
- update tds entities when they change bfae3d7b
- remove global event guard (not needed with Ports) f7ca4cc2
- remove unused `perf` helper 442c767d
- used @duckduckgo/ddg2dnr for the MV3 tracker-stats redirect rule 40c91de7

shared/js/background/before-request.es6.js
sammacbeth: You have to manually allow the extension in incognito mode:
 1. From `chrome://extensions/`, click 'details' for the extension. Screenshot 2023-02-01 at 13 26 24
 2. Toggle 'Allow in Incognito' on.

Screenshot 2023-02-01 at 13 26 28

shared/js/background/startup.js
```diff
@@ -39,6 +42,25 @@ export async function onStartup () {
 
     Companies.buildFromStorage()
 
+    /**
+     * in Chrome only, try to initiate the `NewTabTrackerStats` feature
+     */
+    if (utils.getBrowserName() === 'chrome') {
```
sammacbeth, line 48: Can you confirm that this is false for edge and brave? What happens in chromium?

shared/js/newtab/newtab.js
```diff
@@ -0,0 +1,130 @@
...
+
+    port.onMessage.addListener((msg) => {
+        if (msg.messageType in events.outgoing) {
+            sendToNewTabPage(msg)
+        }
+    })
+
+    port.onDisconnect.addListener((msg) => {
+        console.log('🚨 PORT onDisconnect', msg)
+    })
```
sammacbeth, line 33: In MV3, if the service worker is shutdown, this port will disconnect, and the stats will stop updating for that new tab. Should we try to reconnect in that case?

browsers/chrome-mv3/manifest.json
```diff
@@ -21,6 +21,9 @@
     "background": {
         "service_worker": "public/js/background.js"
     },
+    "chrome_url_overrides" : {
+        "newtab": "html/redirect.html"
+    },
```
sammacbeth, line 26: To be removed?

browsers/chrome-mv3/manifest.json
```diff
@@ -81,7 +84,9 @@
                 "/web_accessible_resources/*",
                 "/img/logo-small.svg",
                 "/public/css/autofill.css",
-                "/public/font/*"
+                "/public/font/*",
+                "/public/js/newtab.js",
```
sammacbeth, line 88: Does this have to be web-accessible? As it's just loaded by tracker-stats.html, maybe it's enough to have the html web-accessible?

sammacbeth (commented):
Looking good - I've just added a few more comments.

shared/js/newtab/newtab.js
shakyShane: We'd have to reload the iframe either from within, or get the SERP side to detect the error and reload it.

short version is that you cannot re-connect once the port connection was lost (I tried lots of ways)

browsers/chrome-mv3/manifest.json
shakyShane: yep

browsers/chrome-mv3/manifest.json
shakyShane: ahh nice! yeah that's a good change

shared/js/background/startup.js
shakyShane: Ok, tested Brave, it's `brave` there so we're good - I can test edge later too :)

shakyShane: edge gives `edg` -> so it looks like we're all good here :)

shared/js/background/before-request.es6.js
shakyShane: @sammacbeth - thanks for this

I looked into it, you can no longer detect that a request came from an incognito window - the property shown in those MDN docs is absent on all requests.

It seems like this is by design and I don't think we can/should pursue solving it right now :)

sammacbeth (approved):

PR #1642, opened 2023-01-12
Consolidate and parallelize PR check actions

sammacbeth:
**Reviewer:**

## Description:
- Consolidate all PR checks (lint, build and tests) into a single file.
- Split MV2 and MV3 integration tests into separate jobs.
- Use node LTS (18) as the default for all actions.
- Also test building against node current (19).

## Automated tests:
- [ ] Unit tests
- [ ] Integration tests

###### Reviewer Checklist:

###### PR Author Checklist:


.github/workflows/checks.yml
```diff
@@ -13,30 +13,44 @@ jobs:
...
           cache: 'npm'
       - run: npm run install-ci
       - run: npm run lint
 
-  unit:
+  build:
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        version: [18, 19]
```
kzar, line 25: Maybe have 16 in the list too? My system is still on 16 at least.

kzar (approved):
LGTM with one suggestion

.github/workflows/checks.yml
sammacbeth: I wanted to avoid testing too many combinations. Either we go with the two most common versions in use (probably 16 and 18), or test against LTS + current (18 and 19), so we get an early warning of future build breaking issues. Thinking about it, maybe 16 and 18 makes more sense here?

PR #1643, opened 2023-01-13
MV3 Smarter encryption

sammacbeth:
**Reviewer:**

## Description:
Smarter encryption for MV3:
- Fetch SE domain list at build time, convert it into DNR rules, then bundle that file with the build.
- When we detect a redirect loop, create a session rule to disable upgrades for that domain.
- For upgrades beyond the static DNR rule list, generate session upgrade rules when we get an upgrade hit from the SE API.
- Update SE API caching for MV3, using session storage.

Depends on https://github.com/duckduckgo/ddg2dnr/pull/68

## Steps to test this PR:
Upgrades: http://privacy-test-pages.glitch.me/privacy-protections/https-upgrades/
Loop protection: https://privacy-test-pages.glitch.me/privacy-protections/https-loop-protection/

## Automated tests:
- [ ] Unit tests
- [ ] Integration tests

###### Reviewer Checklist:

###### PR Author Checklist:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -79,7 +79,7 @@ beta-firefox-zip: remove-firefox-id
 	cd build/firefox/release/ && web-ext build
 
 shared/data/smarter_encryption.txt:
-	curl https://staticcdn.duckduckgo.com/trackerblocking/smarter_encryption_test.txt.gz | gunzip -c > shared/data/smarter_encryption.txt
+	curl https://staticcdn.duckduckgo.com/https/smarter_encryption.txt.gz | gunzip -c > shared/data/smarter_encryption.txt
 
 shared/data/bundled/smarter-encryption-rules.json: shared/data/smarter_encryption.txt
 	npm run bundle-se
--- a/Makefile
+++ b/Makefile
@@ -21,7 +21,11 @@ setup-artifacts-dir:
 	mkdir -p integration-test/artifacts/screenshots
 	mkdir -p integration-test/artifacts/api_schemas
 
+ifeq ('$(browser)','chrome-mv3')
+setup-build-dir: shared/data/bundled/smarter-encryption-rules.json
+else
 setup-build-dir:
+endif
 	mkdir -p build/$(browser)
 	rm -rf build/$(browser)/$(type)
 	mkdir build/$(browser)/$(type)
@@ -73,3 +77,12 @@ remove-firefox-id:
 
 beta-firefox-zip: remove-firefox-id
 	cd build/firefox/release/ && web-ext build
+
+shared/data/smarter_encryption.txt:
+	curl https://staticcdn.duckduckgo.com/trackerblocking/smarter_encryption_test.txt.gz | gunzip -c > shared/data/smarter_encryption.txt
+
+shared/data/bundled/smarter-encryption-rules.json: shared/data/smarter_encryption.txt
+	npm run bundle-se
+
+clean:
+	rm -f shared/data/smarter_encryption.txt shared/data/bundled/smarter-encryption-rules.json
--- a/package.json
+++ b/package.json
@@ -83,7 +83,7 @@
     "dependencies": {
         "@duckduckgo/autofill": "github:duckduckgo/duckduckgo-autofill#6.1.1",
         "@duckduckgo/content-scope-scripts": "github:duckduckgo/content-scope-scripts#4.0.1",
-        "@duckduckgo/ddg2dnr": "github:duckduckgo/ddg2dnr#0.3.0",
+        "@duckduckgo/ddg2dnr": "github:duckduckgo/ddg2dnr#0.3.1",
         "@duckduckgo/jsbloom": "^1.0.2",
         "@duckduckgo/privacy-dashboard": "github:duckduckgo/privacy-dashboard#1.2.0",
         "@duckduckgo/privacy-grade": "github:duckduckgo/privacy-grade#2.1.2",
--- a/package.json
+++ b/package.json
@@ -11,6 +11,7 @@
         "postinstall": "npm run copy-modules && export PATH=\"$PATH:$PWD/node_modules/chromedriver/bin\" && npm run setup-test-int",
         "copy-modules": "rsync -av --exclude='.git/' node_modules/@duckduckgo/privacy-reference-tests/ unit-test/data/reference-tests/ && cp -r node_modules/@duckduckgo/content-scope-scripts shared/ && cp -r node_modules/@duckduckgo/tracker-surrogates shared/",
         "bundle-config": "node scripts/bundleConfig.mjs && node scripts/bundleTrackers.mjs",
+        "bundle-se": "ddg2dnr smarter-encryption shared/data/smarter_encryption.txt shared/data/bundled/smarter-encryption-rules.json",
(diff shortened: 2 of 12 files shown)


shared/js/background/declarative-net-request.js
```diff
@@ -567,6 +573,44 @@ export function flushSessionRules () {
     })
 }
 
+/**
+ * Find the session rule with the given ID, if it exists.
+ * @param {number} id Rule ID
+ * @returns {Promise<chrome.declarativeNetRequest.Rule | undefined>}
+ */
+async function findSessionRuleFromId (id) {
```
jdorweiler, line 581: I guess we still need this for now, but fyi this was just fixed https://bugs.chromium.org/p/chromium/issues/detail?id=1379699

Makefile
```diff
@@ -73,3 +77,12 @@ remove-firefox-id:
 
 beta-firefox-zip: remove-firefox-id
 	cd build/firefox/release/ && web-ext build
+
+shared/data/smarter_encryption.txt:
+	curl https://staticcdn.duckduckgo.com/trackerblocking/smarter_encryption_test.txt.gz | gunzip -c > shared/data/smarter_encryption.txt
```
sammacbeth, line 82: This URL needs to be updated to a production version before merging.

shared/js/background/declarative-net-request.js
kzar: Yea, also see `findExistingDynamicRule` above which does nearly the same. Perhaps you could expand the scope of `findExistingDynamicRule` to handle both session and dynamic rules?

shared/js/background/declarative-net-request.js
```diff
@@ -571,6 +577,44 @@ export function flushSessionRules () {
...
+    return sessionRules.find(r => r.id === id)
+}
+
+/**
+ * Update a Smarter Encryption session rule, adding the given domain to the list of domains in the condition.
+ * @param {number} ruleId Session rule ID
+ * @param {string} addDomain Domain to add to this rule's requestDomains condition.
+ * @param {'allow' | 'upgrade'} type If the rule should be an allow or upgrade rule.
+ */
+async function updateSeSessionRule (ruleId, addDomain, type) {
```
kzar, line 596: Nit: I would go with `updateSmarterEncryptionSessionRule`, just because `updateSeSessionRule` looks a bit like a typo on first glance.

shared/js/background/declarative-net-request.js
```diff
@@ -35,14 +38,17 @@ const ruleIdRangeByConfigName = {
...
-export const SERVICE_WORKER_INITIATED_ALLOWING_RULE_ID = 20002
 export const ATB_PARAM_RULE_ID = 20003
 const RESERVED_RULE_IDS = [
     USER_ALLOWLIST_RULE_ID,
-    SERVICE_WORKER_INITIATED_ALLOWING_RULE_ID,
     ATB_PARAM_RULE_ID
 ]
 
+// Rule IDs for static session rules
+export const SERVICE_WORKER_INITIATED_ALLOWING_RULE_ID = 100
```
kzar, line 48: My vote would be to keep these in the reserved rule ID range, and to get rid of `clearInvalidRules` + `RESERVED_RULE_IDS`. I haven't checked if session/dynamic rule IDs can overlap, but it seems kind of confusing to have them overlap even if it's allowed IMO.

package.json
```diff
@@ -82,7 +83,7 @@
     "dependencies": {
         "@duckduckgo/autofill": "github:duckduckgo/duckduckgo-autofill#6.1.1",
         "@duckduckgo/content-scope-scripts": "github:duckduckgo/content-scope-scripts#4.0.1",
-        "@duckduckgo/ddg2dnr": "github:duckduckgo/ddg2dnr#0.3.0",
+        "@duckduckgo/ddg2dnr": "github:duckduckgo/ddg2dnr#f6f1d9c968d645641489dec38ce11ff843e59141",
```
jdorweiler, line 86: Don't forget to update after we do a new dnr version.

jdorweiler:
1 question, but otherwise looks good. 

Are we worried about stored cache results being out of sync with the bundled data or are those cleaned up? I'm thinking there could be a case where we cache an upgrade, but it was later removed from smarter encryption data.

jdorweiler (approved):

package.json
kzar: I've [cut a release (0.3.1)](https://github.com/duckduckgo/ddg2dnr/commit/c11effd859087313a80b382623648b3a612a98c6) in case that helps.

shared/js/background/declarative-net-request.js
```diff
@@ -63,16 +70,23 @@ function generateEtagRule (id, etag) {
...
 
 /**
- * Find an existing dynamic declarativeNetRequest rule with the given rule ID
+ * Find an existing session or dynamic declarativeNetRequest rule with the given rule ID
  * and return it.
  * @param {number} desiredRuleId
  * @returns {Promise<import('@duckduckgo/ddg2dnr/lib/utils.js').DNRRule|null>}
  */
-async function findExistingDynamicRule (desiredRuleId) {
+async function findExistingSessionOrDynamicRule (desiredRuleId) {
```
kzar, line 78: I think the code should be shared for fetching session/dynamic rules here, but I don't think the function should check both. I think the caller should specify which (session or dynamic) it wants to check.

My vote would be to have this function take another argument which is a bool to specify session/dynamic. Then to have two wrapper functions to make it nicer to use. Something like:

```javascript
async function findExistingSessionOrDynamicRule (session, desiredRuleId) {
    const rules = await chrome.declarativeNetRequest[session ? 'getSessionRules' : 'getDynamicRules']()
    ...
}

const findExistingDynamicRule = findExistingSessionOrDynamicRule.bind(null, false);
const findExistingSessionRule = findExistingSessionOrDynamicRule.bind(null, true);
```

shared/js/background/declarative-net-request.js
kzar: We previously didn't have the `ruleIdRangeByConfigName['_RESERVED']` range (20001 to 21000), but we had some hard-coded rule IDs in that range (e.g. `USER_ALLOWLIST_RULE_ID` with ID of 20001). That was a "foot-gun waiting to happen", since when a new config rule ID range got added (combined) the logical place to put it was (20001, 30000) and that then immediately clashed with the hard-coded rule IDs.

Sam and I went back-and-forth on this a few times, weighing changing the IDs around vs making things clear. We settled on adding the "_RESERVED" range to make it explicit without changing the rule IDs for any existing installs.

Along the way, Sam added the `RESERVED_RULE_IDS` and `clearInvalidRules` to try to and clear rules with the wrong IDs. The idea there was to fix installs that were broken by us changing the rule IDs about, but since that didn't happen in the end IMO that logic should be removed. This PR makes me feel stronger to remove it, since it's getting kind of confusing with session VS dynamic rules and this doesn't help.

sammacbeth:
@jdorweiler It is true that the session rule we create for an upgrade may last longer than we would previously cache the API response. However, if there is serious breakage (i.e. redirect loop), we should still detect it and override the upgrade rule with an allow one. I'm not sure how often a working upgrade rule is removed, but it seems unlikely.

PR #1657, opened 2023-01-25
Re-enable serviceworker blocking

sammacbeth:
**Reviewer:**

## Automated tests:
- [ ] Unit tests
- [ ] Integration tests

###### Reviewer Checklist:

###### PR Author Checklist:

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1203879775591697

### DIFF
--- a/integration-test/background/request-blocking.js
+++ b/integration-test/background/request-blocking.js
@@ -4,20 +4,21 @@ const backgroundWait = require('../helpers/backgroundWait')
 const pageWait = require('../helpers/pageWait')
 const { loadTestConfig } = require('../helpers/testConfig')
 
-const testSite = 'https://privacy-test-pages.glitch.me/privacy-protections/request-blocking/'
+const testHost = 'privacy-test-pages.glitch.me'
+const testSite = `https://${testHost}/privacy-protections/request-blocking/`
 
 let browser
 let bgPage
 let teardown
 
 describe('Test request blocking', () => {
-    beforeAll(async () => {
+    beforeEach(async () => {
         ({ browser, bgPage, teardown } = await harness.setup())
         await backgroundWait.forAllConfiguration(bgPage)
         await loadTestConfig(bgPage, 'serviceworker-blocking.json')
     })
 
-    afterAll(async () => {
+    afterEach(async () => {
         await teardown()
     })
 
@@ -99,5 +100,57 @@ describe('Test request blocking', () => {
         })
 
         await page.close()
+    })
+
+    it('serviceworkerInitiatedRequests exceptions should disable service worker blocking', async () => {
+        const page = await browser.newPage()
+        await bgPage.evaluate(async (domain) => {
+            /* global dbg */
+            const { data: config } = dbg.getListContents('config')
+            config.features.serviceworkerInitiatedRequests.exceptions.push({
+                domain,
+                reason: 'test'
+            })
+            await dbg.setListContents({
+                name: 'config',
+                value: config
+            })
+        }, testHost)
+        await pageWait.forGoto(page, testSite)
+
+        // Start logging network requests.
+        const pageRequests = []
+        await logPageRequests(
+            page,
+            pageRequests,
+            ({ url }) => url.hostname === 'bad.third-party.site'
+        )
+        await page.click('#start')
+        const testCount = await page.evaluate(
+            // eslint-disable-next-line no-undef
+            () => tests.filter(({ id }) => !id.includes('worker')).length
+        )
+        while (pageRequests.length < testCount) {
+            await backgroundWait.forTimeout(bgPage, 100)
+        }
+        // Verify that no logged requests were allowed.
+        for (const { url, method, type, status } of pageRequests) {
+            const description = `URL: ${url}, Method: ${method}, Type: ${type}`
+            expect(status).withContext(description).toEqual('blocked')
+        }
+
+        // Also check that the test page itself agrees that no requests were
+        // allowed.
+        const pageResults = await page.evaluate(
+            () => results.results // eslint-disable-line no-undef
+        )
+        for (const { id, category, status } of pageResults) {
+            const description = `ID: ${id}, Category: ${category}`
+            if (id === 'serviceworker-fetch') {
+                expect(status).withContext(description).toEqual('loaded')
+            } else {
+                expect(status).withContext(description).not.toEqual('loaded')
+            }
+        }
     })
 })
--- a/integration-test/background/request-blocking.js
+++ b/integration-test/background/request-blocking.js
@@ -2,6 +2,7 @@ const harness = require('../helpers/harness')
 const { logPageRequests } = require('../helpers/requests')
 const backgroundWait = require('../helpers/backgroundWait')
 const pageWait = require('../helpers/pageWait')
+const { loadTestConfig } = require('../helpers/testConfig')
 
 const testSite = 'https://privacy-test-pages.glitch.me/privacy-protections/request-blocking/'
 
@@ -13,6 +14,7 @@ describe('Test request blocking', () => {
     beforeAll(async () => {
         ({ browser, bgPage, teardown } = await harness.setup())
         await backgroundWait.forAllConfiguration(bgPage)
+        await loadTestConfig(bgPage, 'serviceworker-blocking.json')
     })
 
     afterAll(async () => {
@@ -64,14 +66,6 @@ describe('Test request blocking', () => {
         )
         for (const { id, category, status } of pageResults) {
             const description = `ID: ${id}, Category: ${category}`
-
-            // ServiceWorker initiated request blocking is not yet supported.
-            // TODO: Remove this condition once they are blocked again.
-            if (id === 'serviceworker-fetch') {
-                expect(status).withContext(description).toEqual('loaded')
-                continue
-            }
-
             expect(status).withContext(description).not.toEqual('loaded')
         }
 
--- a/integration-test/background/request-blocking.js
+++ b/integration-test/background/request-blocking.js
@@ -11,6 +11,39 @@ let browser
 let bgPage
 let teardown
 
+async function runRequestBlockingTest (page) {
+    await pageWait.forGoto(page, testSite)
+
+    // Start logging network requests.
+    const pageRequests = []
+    await logPageRequests(
+        page,
+        pageRequests,
+        ({ url }) => url.hostname === 'bad.third-party.site'
+    )
+
+    // Initiate the test requests and wait until they have all completed.
+    // Note:
+    //  - Waiting for network idle here does not work for some reason.
+    //  - ServiceWorker + WebWorker initiated requests are not yet logged by
+    //    logPageRequests.
+    await page.click('#start')
+    const testCount = await page.evaluate(
+        // eslint-disable-next-line no-undef
+        () => tests.filter(({ id }) => !id.includes('worker')).length
+    )
+    while (pageRequests.length < testCount) {
+        await backgroundWait.forTimeout(bgPage, 100)
+    }
+
+    // Wait an additional second to ensure that the WebWorker +
+    // ServiceWorker initiated requests have finished. This can be removed
+    // once logPageRequests can intercept those requests.
+    await backgroundWait.forTimeout(bgPage, 1000)
+
+    return [testCount, pageRequests]
+}
+
 describe('Test request blocking', () => {
     beforeEach(async () => {
         ({ browser, bgPage, teardown } = await harness.setup())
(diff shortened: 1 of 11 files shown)


integration-test/background/request-blocking.js
```diff
@@ -106,4 +101,56 @@ describe('Test request blocking', () => {
 
         await page.close()
     })
+
+    it('serviceworkerInitiatedRequests exceptions should disable service worker blocking', async () => {
```
kzar, line 105: If I understand it right, this test case is pretty similar to the above one, except we're turning off ServiceWorker blocking. Could we not have a function that does the work of loading the test page and getting the list of results? I figure each test case would then just need to call that function, compare the results to what's expected and optionally disable ServiceWorker initaited request blocking.

shared/js/background/before-request.es6.js
```diff
@@ -274,6 +269,15 @@ function blockHandleResponse (thisTab, requestData) {
             tracker.reason = 'content blocking disabled'
         }
 
+        if (requestData.tabId === -1 && (tracker.action === 'block' || tracker.action === 'redirect')) {
```
kzar, line 272: I'm sure you've thought about this, but it makes me nervous that this check is made later than the old one. Is there a reason you can't or don't want to keep the check in the same place? What's your thinking?

shared/js/background/before-request.es6.js
```diff
@@ -274,6 +269,15 @@ function blockHandleResponse (thisTab, requestData) {
...
+            if (!thisTab.site.isFeatureEnabled('serviceworkerInitiatedRequests')) {
+                tracker.action = 'ignore-user'
+                tracker.reason = 'service worker blocking disabled'
```
kzar, line 275: Nit: 'service worker initiated request blocking disabled'

shared/js/background/before-request.es6.js
```diff
@@ -285,14 +289,15 @@ function blockHandleResponse (thisTab, requestData) {
...
+        thisTab.postDevtoolsMessage(devtools, 'tracker', {
             tracker: {
                 ...reportedTracker,
                 matchedRule: reportedTracker.matchedRule?.rule?.toString()
             },
             url: cleanUrl,
             requestData,
-            siteUrl: thisTab.site.url
+            siteUrl: thisTab.site.url,
+            serviceworkerInitiated: requestData.tabId === -1
```
kzar, line 300: Nit:
```suggestion
            serviceWorkerInitiated: requestData.tabId === -1
```

shared/js/background/before-request.es6.js
```diff
@@ -323,7 +328,7 @@ function blockHandleResponse (thisTab, requestData) {
 
             console.info('blocked ' + utils.extractHostFromURL(thisTab.url) +
                         // @ts-ignore
-                        ' [' + tracker.tracker.owner.name + '] ' + requestData.url)
+                        ' [' + tracker.tracker.owner.name + '] ' + requestData.url + (requestData.tabId === -1 ? ' (serviceworker)' : ''))
```
kzar, line 331: Nit: Assign `serviceWorkerInitiated` bool to a variable above, then you can pass it through to the postDevToolsMessage with the Object shorthand. Also, then you can reuse the variable here to avoid doing the `tabId === -1` check again.

shared/js/background/classes/sw-tab.es6.js
```diff
@@ -0,0 +1,62 @@
+const Tab = require('./tab.es6')
+
+/**
+ * A stub tab implementation for service workers.
+ *
+ * For most actions, this just mimics a tab with the service worker URL as the tab URL. However, for cases
+ * such as tracker counts, it distributes stats to all known tabs with the same origin.
+ */
+class ServiceWorkerTab extends Tab {
```
kzar, line 9: Nice idea :+1:

shared/js/background/classes/sw-tab.es6.js
```diff
@@ -0,0 +1,62 @@
...
+
+    /**
+     * Find the list of tabs which share the same origin as this service worker.
+     * @returns {Tab[]}
+     */
+    _findMatchingTabs () {
+        return Object.keys(this.tabContainer).filter(tabId => {
+            const tab = this.tabContainer[tabId]
+            try {
+                return Number(tabId) > -1 && new URL(tab.url).origin === this.origin
```
kzar, line 32: Why cast the tabId to a number? If it's not a number, what value will that result in?

shared/js/background/classes/sw-tab.es6.js
```diff
@@ -0,0 +1,62 @@
...
+            tabId: -1,
+            url: swUrl,
+            status: null
+        })
+        this.origin = new URL(swUrl).origin
+        this.tabContainer = tabContainer
+    }
+
+    /**
+     * Find the list of tabs which share the same origin as this service worker.
```
kzar, line 25: Shouldn't we be checking for same entity, rather than same origin? I'm wondering if we could compare `tab.site.parentEntity` instead?

shared/js/background/classes/sw-tab.es6.js
```diff
@@ -0,0 +1,62 @@
...
+     * @returns {Tab[]}
+     */
+    _findMatchingTabs () {
```
kzar, line 28: I'm figuring this will be called quite often, like for every request with a tabId of -1. So perhaps we should have a more efficient way of finding the relevant tabs?

How about we give ServiceWorker tabs a WeakMap of associated Tab Objects (same origin/entity). Each time a normal Tab Object is created, we update the WeakMaps for all the ServiceWorker Tabs. Then this function could just return the values of the WeakMap. WDYT?

shared/js/background/classes/sw-tab.es6.js
```diff
@@ -0,0 +1,62 @@
...
+    }
+
+    /**
+     * Post a message to the devtools panel for all matching
+     * @param {Object} devtools
+     * @param {string} action
+     * @param {Object} message
+     */
+    postDevtoolsMessage (devtools, action, message) {
+        this._findMatchingTabs().forEach(tab => tab.postDevtoolsMessage(devtools, action, message))
```
kzar, line 58: _I would prefer we posted ServiceWorker requests to all tabs for devtools, but clearly labeled them as being ServiceWorker initiated. But you know this UI better, so I defer to you if you think this makes more sense._

shared/js/background/before-request.es6.js
sammacbeth: Before we block the serviceworker initiated request, we need to know:
- The 'tab' context (initiator)
- If the URL is a tracker request we should block (taking the tracker allowlist into account)
- That there isn't an exception for the first party (user allowlist, feature allowlist etc)

This is the point in the code where we have all that information, and can make the blocking decision.

shared/js/background/classes/sw-tab.es6.js
sammacbeth: No - we only want to distribute stats to Tabs which _could have_ initiated this request. Service workers must share the [same origin](https://github.com/w3c/ServiceWorker/blob/main/explainer.md#getting-started) ("The service worker script must be on the same origin as the page") as the sites they operate for, so the origin check should be sufficient.

shared/js/background/classes/sw-tab.es6.js
sammacbeth: So you're suggesting to keep a live weak mapping of origin -> tab (updating on every webnavigation event), in order to make this lookup faster? I'd imagine there's be a load of edge cases to handle there (e.g. with extension SW restarts).

I'd argue that this is fast enough - we're just doing one loop through the tabs (probably N<100 for most cases), and only when there is a serviceWorker initiated request that is also a tracker (not common).

shared/js/background/classes/sw-tab.es6.js
sammacbeth: This is due to type checking - because object keys are strings, then `tabId` here is a string. Using `Number` casts it to a number for the comparison. If it's not a number, `Number(tabId) === NaN`, and `NaN > -1` is `false`, so I think we're good here.

shared/js/background/classes/sw-tab.es6.js
sammacbeth: This posts them to all panels on the same origin. I think that's what you want, right?

shared/js/background/classes/sw-tab.es6.js
kzar: Got it, that makes sense.

shared/js/background/classes/sw-tab.es6.js
kzar: OK cool.

shared/js/background/classes/sw-tab.es6.js
kzar: Well, yea. I'm not set on the WeakMap approach. Another approach might be to maintain an origin to Tab mapping for efficient lookup. But parsing the URL and comparing origin for every tab, every time there's one of these requests does make me nervous. Some users have a lot of tabs. Maybe there's more of these requests than we expect. I guess it's not a problem until it is...

I suppose we can optimise it in the future too if necessary. How about adding a comment here at least? I figure it could explain that for large numbers of tabs + Serviceworker initiated tracking requests this could get slow. In the future we might want to consider maintaining a more efficient way to find Tab Objects that correspond to a given origin.

shared/js/background/classes/sw-tab.es6.js
kzar: No, I'd rather they were posted to all panels regardless of origin. But perhaps that's just me.

shared/js/background/classes/sw-tab.es6.js
kzar: (So I guess this is just another way of writing `parseInt(tabId, 10)`?)

shared/js/background/before-request.es6.js
kzar: Could we not move the check back to where it was*,  but expand it to check `tabId === -1 && thisTab.site.isFeatureEnabled('serviceworkerInitiatedRequests')`? I figure that would reduce the chance of breakage caused by some other privacy protection (e.g tracker parameter protection, smarter encryption) applying to ServiceWorker initiated requests when the `serviceworkerInitiatedRequests` feature is disabled.

* Well a line lower, so just after the `thisTab` definition in `handleRequest()`

sammacbeth: I'm not sure that actually makes much difference. All that happens between these two points in the code is:
1. `web_accessible_resources` check: required to protect from SWs loading our web-accessible resources.
2. `main_frame` request handling: Not possible from service workers.
3. Skip blocking for 'special' pages: prevents us from running the checks on extension + internal pages.
4. (in `blockHandleResponse`) CTL blocking: not sure if this is likely to trigger breakage, but it can be controlled via its own config.
5. Other allowlist checks: These all disable blocking in specific cases, which would therefore not introduce breakage.

The other effect of the current position is that the allowlist is fall through, meaning that, in this case, Smarter Encryption will still apply, but that will be a no-op as [mixed-content fetch from service-workers is forbidden](https://github.com/w3c/ServiceWorker/issues/493)

Note, the _advantage_ of having the block where it is now, is that it is:
 - alongside other blocking exceptions,
 - only has to be checked when the URL is a tracker URL, and
 - can therefore set the blocking metadata correctly to allow the SW exception to be shown in the dashboard and debugger.

shared/js/background/classes/sw-tab.es6.js
sammacbeth: I think this is premature optimization - the current implementation is `number of tracker requests via SW x number of tabs` `URL` parses for a page load, which is max in the order of 1000s. Note, we already parse the site domain at least once for every request, which is a much hotter path than this.

Additionally, sketching out the alternative, it sounds potentially tricky to implement, particularly with our tab (de)serialization going on (in MV3):
-  Create a mapping of `Map<string, WeakSet<Tab>>`
- Whenever we create a tab object, insert it into the map using the URL origin as the key.
- Also handle tab restores (`tabManager.restore` and `tabManager.restoreOrCreate`).
- Check that a tab's `url` is immutable. If not, handle modification (remove from one weak set and add to another).
- Cleanup origins with no more tabs (otherwise this will leak memory). Avoid doing this too often, as we have to iterate the map values, which will be almost as large as the total number of tabs.

shared/js/background/classes/sw-tab.es6.js
sammacbeth: I think that would be strange, and add noise to the panel. I'll leave it how it is.

shared/js/background/before-request.es6.js
kzar: Alright, fair enough. Sounds good. (P.S. Don't worry about number four, I will be changing that logic very shortly.)

shared/js/background/classes/sw-tab.es6.js
kzar: Alright, fair enough. I would have liked a short comment to note the cost and potential future improvement, but I don't insist. Up to you :+1:

kzar (approved):
LGTM % one optional change

PR #1710, opened 2023-02-07
Remove legacy Click to Load configuration

kzar:
The migration away from the legacy social_ctp_configuration.json
configuration for Click to Load is complete. Let's use the new
configuration for the MV2 code paths and remove all references to the
legacy configuration.

While at it, let's also:
 - Deprecate the clickToLoadClicks Object that's no longer required
   and remove the corresponding logic.
 - Remove the imgToData function that was added for Click to Load but
   no longer used.
 - Remove a now unused webNavigation.onCommitted listener that was
   originally needed for Click to Load.
 - Rewrite the Click to Load unit tests that were passing
   incorrectly.
 - Fix the Click to Load Facebook integration tests, so that they
   aren't marked as pending when blocking fails.

**Reviewer:** @sammacbeth 
Cc: @ladamski, @franfaccin 

## Steps to test this PR:
1. Ensure the Click to Load integration tests are still passing (for MV2 and MV3).
2. Test that Facebook Click to Load works correctly on the Facebook test page (for MV2).

## Automated tests:
- [ ] Unit tests
- [x] Integration tests

###### Reviewer Checklist:

###### PR Author Checklist:


shared/js/background/dnr-click-to-load.js
```diff
@@ -69,6 +69,51 @@ async function generateDnrAllowingRules (tab, ruleAction) {
...
+
+    for (let [entity, { ruleActions, state }] of Object.entries(clickToLoadSettings)) {
+        // No rule actions, or entity is disabled.
+        if (!ruleActions || ruleActions.length === 0 || state !== 'enabled') {
+            continue
+        }
+
+        // TODO: Remove this workaround once the content-scope-scripts and
+        //       privacy-configuration repositories have been updated.
+        if (entity === 'Facebook') entity = 'Facebook, Inc.'
```
kzar, line 105: I changed this deliberately so that when update "Facebook" => "Facebook, Inc." in the config + content-scope-scripts the logic won't break. It's annoying doing the `=== 'Facebook'` comparison in the loop, but it seemed worth it.

shared/js/background/classes/tab-state.js
```diff
@@ -32,6 +32,8 @@ export class TabState {
         this.trackers = {}
         /** @type {null | import('../events/referrer-trimming').Referrer} */
         this.referrer = null
+        /** @type {Record<string, null>} */
```
kzar, line 35: I don't think I can use a Set, since they aren't JSON serialisable.

shared/js/background/dnr-click-to-load.js
```diff
@@ -69,6 +69,51 @@ async function generateDnrAllowingRules (tab, ruleAction) {
...
 }
 
+/**
+ * Find the enabled Click to Load rule actions for the given tab.
+ * Note: Take care to ensure wait for the extension configuration to be ready
+ *       first.
+ * @param {import('./classes/tab.es6')} tab
+ * @return {string[]}
+ */
+export function getDefaultEnabledClickToLoadRuleActionsForTab (tab) {
```
sammacbeth, line 79: This is not DNR related right? Why is it in the `dnr-click-to-load` file? Should we maybe just rename the whole file to `click-to-load` if it is manifest-version agnostic?

shared/js/background/classes/tab.es6.js
```diff
@@ -70,6 +70,14 @@ class Tab {
         return this._tabState.adClick
     }
 
+    set disabledClickToLoadRuleActions (value) {
+        this._tabState.setValue('disabledClickToLoadRuleActions', value)
+    }
+
+    get disabledClickToLoadRuleActions () {
+        return this._tabState.disabledClickToLoadRuleActions
+    }
```
sammacbeth, line 79: I think you're doing too much work just to have `disabledClickToLoadRuleActions` serializable. As you have the `Tab` wrapping the `_tabState`, can't you leverage accessors and class methods to make it seem like a set, but have a list in the background?

i.e.
Tabstate:
```javascript
/** @type {string[]} */
this.disabledClickToLoadRuleActions = []
```
Tab:
```javascript
get disabledClickToLoadRuleActions () {
   return new Set(this._tabState.disabledClickToLoadRuleActions)

// or an explicit action checker function
isClickToLoadRuleActionDisabled (action) {
  return this._tabState.disabledClickToLoadRuleActions.contains(action)
}

addDisabledClickToLoadRuleAction (action) {
  if (!this._tabState.disabledClickToLoadRuleActions.contains(action)) {
     this._tabState.disabledClickToLoadRuleActions.push(action)
  }
}
```

sammacbeth (commented):
Mostly looks good, but I think you've overcomplicated things with the serializable `disabledClickToLoadRuleActions`. I proposed an alternative.

shared/js/background/utils.es6.js
```diff
@@ -317,39 +317,6 @@ function isDomainCookieExcluded (domain) {
     return false
 }
 
-/**
- * Convert an image file to a base64 data:image file,
- * for use in injections where the extension URL may not be
- * accessible
- */
-export async function imgToData (imagePath) {
```
kzar, line 325: I noticed [imgToData was added with the original Click to Load commit](https://github.com/duckduckgo/duckduckgo-privacy-extension/commit/624274921b521c219183f23ad0482780e37a4fc6#diff-fa4895e8a5f9f011f418b7e04b16ec46ac30764db15368ae60a2d6820962df9b), and I can't see where else it's used.

shared/js/background/before-request.es6.js
```diff
@@ -214,39 +216,22 @@ function handleRequest (requestData) {
...
     const blockingEnabled = thisTab.site.isContentBlockingEnabled()
 
-    const tracker = trackers.getTrackerData(requestData.url, thisTab.site.url, requestData)
+    // Find the supported and enabled Click to Load rule actions for this tab.
+    const enabledRuleActions = new Set(
+        getDefaultEnabledClickToLoadRuleActionsForTab(thisTab).filter(ruleAction =>
+            !Object.prototype.hasOwnProperty.call(
+                thisTab.disabledClickToLoadRuleActions,
+                ruleAction
+            )
```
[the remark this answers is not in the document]
kzar: Yea, if `thisTab.disabledClickToLoadRuleActions` started off as `Object.create(null)` instead of `{}` then this would be equivalent. I would have changed the Object to have no prototype, but I figured when it was read back from session storage that work would be undone.

shared/js/background/dnr-click-to-load.js
kzar: Mind if I leave that for now? I thought the same, but if we rename the file some of the other exported function names are less obviously relating to the declarativeNetRequest API. I should probably rename those too, but this PR is already getting huge, and equally having a module just for this function seems like overkill.

shared/js/background/classes/tab.es6.js
kzar: I like the idea of having set/get wrap the Array in a Set, but how will that work with saving mutations. I don't know how this tabState storage magic works, but I'm assuming the mutations made to a Set based on the Array in tabState might be lost?

I don't mind using `Array.prototype.contains` when adding disabled rule actions, but I'm against using that when checking if a rule action is disabled - that happens quite often.

I'm struggling to see how to apply your suggestion with the above constraints, but I'm open to doing this differently. I don't particularly like having the Record<string, null> Object hack either. What do you think?

sammacbeth: What is the concern with `Array.prototype.contains`? My assumption was that `disabledClickToLoadRuleActions` would by mostly an empty list, and then rarely 1 or 2 elements long. In that case, then the overhead of using a `Set` is probably higher than just looking through the array. If that assumption is wrong, then maybe I need to reread the rest of this PR to understand how this works.

kzar: Yea, fair enough, you're right that it will usually be an empty list and even when not the list will be short. The lookup will be cheap enough either way.

I started making the change to use an Array instead, but I do wonder, is it worth it? It seems the change is basically that we can have

```javascript
        getDefaultEnabledClickToLoadRuleActionsForTab(thisTab).filter(
            ruleAction => !disabledClickToLoadRuleActions.includes(ruleAction)
        )
```

instead of

```javascript
        getDefaultEnabledClickToLoadRuleActionsForTab(thisTab).filter(ruleAction =>
            !Object.prototype.hasOwnProperty.call(
                thisTab.disabledClickToLoadRuleActions,
                ruleAction
            )
        )
```

in before-request.es6.js.

At the expense of needing to having to have

```javascript
    if (!tab.disabledClickToLoadRuleActions.includes(data.action)) {
        tab.disabledClickToLoadRuleActions.push(data.action)
    }
```

instead of

```javascript
    tab.disabledClickToLoadRuleActions[data.action] = null
```

in message-handlers.js.

kzar: I don't feel too strongly about it either way, so I've changed it now anyway. Done.

sammacbeth: Thanks - I think the readability gain is worth it.

sammacbeth (approved):

PR #1712, opened 2023-02-08
Playwright for integration tests

sammacbeth:
Demo of how we could replace our current test setup with Playwright.


playwright-test/atb.spec.ts
```diff
@@ -0,0 +1,110 @@
+import { test, expect, mockAtb } from "./harness";
+
```
kzar, line 2: I think we should rethink the directory structure a bit, so that it's clear that the playwright tests are integration tests, so that it's clear that the old integration tests are legacy ones, and so that it's clear which shared files (data and utilities) are shared between the Playwright and legacy integration tests.

Ideally this would be explained in a README or similar too, so folks writing new tests no where to put them and which examples to copy. Otherwise, new tests will be written by copying the legacy tests and we'll have more work to do if we want to eventually migrate them.

playwright-test/atb.spec.ts
```diff
@@ -0,0 +1,110 @@
+import { test, expect, mockAtb } from "./harness";
```
kzar, line 1: Shouldn't the legacy atb integration tests be removed?

playwright-test/atb.spec.ts
kzar, line 2: I don't feel strongly about it, but I know using TypeScript `.ts` files was kind of controversial at DuckDuckGo in the past. This is the first I've seen in the extension. My vote is to just go with it, I don't see the harm. But what do you think @jonathanKingston ?

playwright-test/harness.ts
```diff
@@ -0,0 +1,125 @@
...
+            args: [
+                `--disable-extensions-except=${pathToExtension}`,
+                `--load-extension=${pathToExtension}`,
+            ],
+        });
+        // intercept extension install page and use HAR
+        context.on("page", (page) => {
+            console.log("page", page.url());
+            if (page.url().includes("duckduckgo.com/extension-success")) {
+                page.routeFromHAR(path.join(__dirname, "data", "postinstall.har"), {
```
kzar, line 60: I'm curious, how are these HAR files generated? (Sorry, I've not looked through the docs yet.)

Also, perhaps the HAR files should live in a separate directory (e.g. `./data/har-responses`) if we're going to keep adding more? Also, perhaps we should follow a convention with their naming, like start with the URL and then have a description? I'm thinking something like `duckduckgo.com/extension-success - postinstall.har` might be good. WDYT?

playwright-test/harness.ts
```diff
@@ -0,0 +1,125 @@
...
+                    notFound: "abort",
+                });
+            }
+        });
+        //
+        await use(context);
+        await context.close();
+    },
+    backgroundPage: async ({ context, manifestVersion }, use) => {
```
jonathanKingston, line 69: We generally use the following structure:

```suggestion
    async backgroundPage ({ context, manifestVersion }, use) {
```

playwright-test/atb.spec.ts
jonathanKingston: Sam's rationale of the code samples of Playwright are default in TS doesn't to me warrant a departure from the existing codebase standard / direction.

The aim is to move to zero compilation and just because one tool does transpliation for you doesn't really justify it and may confuse new comers on the style.

playwright-test/harness.ts
jonathanKingston: Sorry, I dunno why I decided to give other feedback (I hadn't noticed it was a PoC 👍🏻 )

playwright-test/atb.spec.ts
```diff
@@ -0,0 +1,110 @@
...
+        // wait for post install page to open
+        // if it never does, jasmine timeout will kick in
+        while (!postInstallOpened) {
+            const urls = await Promise.all(
+                context.pages().map((target) => target.url())
+            );
+            postInstallOpened = urls.some((url) =>
+                url.includes("duckduckgo.com/extension-success")
+            );
+            await page.waitForTimeout(100);
```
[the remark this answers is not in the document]
sammacbeth: This is copied from the original test: https://github.com/duckduckgo/duckduckgo-privacy-extension/blob/develop/integration-test/background/atb.js#L35

playwright-test/atb.spec.ts
sammacbeth: How about we move playwright tests to the root of the `integration-test` folder, then put the old tests in a `legacy` subfolder of that?

playwright-test/atb.spec.ts
sammacbeth: Yes. Just left here for comparison.

playwright-test/atb.spec.ts
sammacbeth: I updated these files to `.js`

playwright-test/harness.ts
sammacbeth: Playwright has a built in generator tool. I added a comment for how to create this one: https://github.com/duckduckgo/duckduckgo-privacy-extension/blob/0296948f8e0c83b575b2a20f79e965e0877a0f12/playwright-test/harness.js#L49

Good suggestion on a naming convention.

playwright-test/atb.spec.ts
kzar: Sounds good to me :+1:

integration-test/background/atb.js
```diff
@@ -178,7 +48,7 @@ describe('search workflow', () => {
     let twoWeeksAgoAtb
 
     beforeAll(async () => {
```
kzar, line 50: How come we're not moving all the atb tests over?

playwright-test/atb.spec.ts
kzar: Cool, thanks :+1:

playwright-test/harness.ts
kzar: While it's a Nit, I do agree. Please could you use the method shorthand for these functions @sammacbeth?

playwright-test/harness.ts
kzar: Ace thanks :+1: Would you also mind adding a README.md to the har directory? Just to quickly explain what the files are for and to include that example command.

playwright-test/atb.spec.ts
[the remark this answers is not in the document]
kzar: Well, fair enough. But if you don't want to fix that here, could you at least add a comment that we should remove the arbitrary delay and make use of the pageWait helpers?

integration-test/background/atb.js
sammacbeth: I didn't want to spend too much time adding those for this PoC. Looking at those remaining tests, I think some additional thought would be needed for how to move them to an offline test (and if that makes sense), as they seem to depend on the SERP's behavior. I'm not familiar enough with the ATB dynamics to do that myself.

kzar: Ah OK, it's just I figure these questions are going to come up the first time someone needs to write a new integration test. If we're going to merge this, I'd like to figure out how we expect folks to write tests going forward.

Do the new Playwrite tests have to be 100% offline? I'm wondering if we could simply move the remaining atb tests over into the same file (maybe with a TODO comment), even though they require internet access.

integration-test/content-scripts/gpc.js
```diff
@@ -1,91 +0,0 @@
-/**
```
kzar, line 1: I think the remaining content-scripts integration tests (that as far as I can tell aren't ever used) should be either deleted or moved inside of the legacy directory. So something like `integration-test/legacy/content-scripts`, just so it's clear they're old cruft.

kzar (approved):
LGTM % One comment about the remaining (unused?) legacy content-script tests.

PR #1716, opened 2023-02-13
Do not create release as draft

sammacbeth:
**Reviewer:**

## Description:

Make a full release when running the release pipeline.

This change is because, when we create a draft release, the tag is only created when someone publishes that draft. This means that, unless the draft is published immediately, the tag will be incorrect. This is especially problematic when we use the Github release to download source for extension release, as if the tag commit is wrong, it will not reproduce the same build.


jonathanKingston (approved):

jonathanKingston:
Thanks for the explanation :)

PR #1733, opened 2023-02-21
Remove .es6 transform and filenames

jonathanKingston:
Related to https://app.asana.com/0/892838074342800/1203977997399511/f


sammacbeth (approved):

Gruntfile.js
```diff
@@ -217,10 +217,6 @@ module.exports = function (grunt) {
                 files: watch.contentScope,
                 tasks: ['exec:buildContentScope', 'copy:contentScope']
             },
-            backgroundES6JS: {
-                files: watch.background,
-                tasks: ['browserify:background']
```
kzar, line 222: Could we drop the browserify step above for the UI code too?

kzar (approved):
LGTM % The comment about removing browserify completely.

Gruntfile.js
jonathanKingston: I just noticed I was missing the other 'browserify:background' I don't think this had a functional change tbh, the bit I removed is just the watcher.

I've pushed to CI to see what explodes; I expect everything :D

jonathanKingston: Simple answer is no, we should fast follow with that change though I think.

PR #1735, opened 2023-02-24
PoC: Replace grunt with make

sammacbeth:
**Reviewer:**

## Automated tests:
- [ ] Unit tests
- [ ] Integration tests

###### Reviewer Checklist:

###### PR Author Checklist:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -96,6 +96,10 @@ BUILD_FOLDERS = $(BUILD_DIR)/public/js/content-scripts $(BUILD_DIR)/public/css
 BROWSERIFY = node_modules/.bin/browserify
 DASHBOARD_DIR = node_modules/@duckduckgo/privacy-dashboard/build/app/
 SURROGATES_DIR = node_modules/@duckduckgo/tracker-surrogates/surrogates
+BROWSER_TYPE = $(browser)
+ifeq ($(browser), chrome-mv3)
+	BROWSER_TYPE := chrome
+endif
 
 # create build dir
 prepare-build-dir:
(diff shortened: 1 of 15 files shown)


Makefile
```diff
@@ -78,11 +94,144 @@ remove-firefox-id:
...
 	curl https://staticcdn.duckduckgo.com/https/smarter_encryption.txt.gz | gunzip -c > shared/data/smarter_encryption.txt
 
+# build SE rules for MV3
 shared/data/bundled/smarter-encryption-rules.json: shared/data/smarter_encryption.txt
 	npm run bundle-se
 
+# fetch integration test data
+integration-test/artifacts/attribution.json: node_modules/privacy-test-pages/adClickFlow/shared/testCases.json setup-artifacts-dir
+	mkdir -p integration-test/artifacts
+	cp $< $@
```
sammacbeth, line 124: @jonathanKingston this replaces `attributionTestCases.mjs`, and moves to a local copy rather than fetching from glitch.

scripts/buildTests.js
```diff
@@ -0,0 +1,37 @@
...
+
+function listSourceFiles (dir) {
+    return fs.readdirSync(dir).filter(f => f.endsWith('.js')).map(f => path.join(dir, f))
+}
+const input = listSourceFiles(path.join('unit-test', 'background'))
+    .concat(listSourceFiles(path.join('unit-test', 'background', 'classes')))
+    .concat(listSourceFiles(path.join('unit-test', 'background', 'events')))
+    .concat(listSourceFiles(path.join('unit-test', 'background', 'reference-tests')))
+    .concat(listSourceFiles(path.join('unit-test', 'background', 'storage')))
+const b = browserify(input, {
```
kzar, line 16: Nit: Think it would be easier to read if you added a newline and skipped the `b` variable.
```suggestion

browserify(input, {
...
}).bundle(err, bundle) => {
...
})
```
```

File: scripts/bundleContentScopeScripts.mjs [hunk 3]
```diff
@@ -0,0 +1,13 @@
+import fs from 'fs/promises'
+const [targetPath, sourcePath, trackerLookupPath, configPath] = process.argv.slice(2)
+const utf8 = { encoding: 'utf-8' };
+
+(async () => {
+    const source = await fs.readFile(sourcePath, utf8)
```
kzar, line 6: Nit: If we're waiting for each file to read in turn anyway, maybe just use the synchronous file APIs, which IIRC are quicker anyway + would mean we could skip wrapping in this async function.

scripts/bundleContentScopeScripts.mjs
```diff
@@ -0,0 +1,13 @@
+import fs from 'fs/promises'
```
kzar, line 1: How come this is a `.mjs` (instead of `.js`) file?

shared/js/background/before-request.js
```diff
@@ -1,6 +1,6 @@
 import browser from 'webextension-polyfill'
 import EventEmitter2 from 'eventemitter2'
-const tldts = require('tldts')
```
kzar, line 3: These changes don't seem to generally change require to import, so I'm wondering if there's a reason to change this one? If not, maybe save this for a follow-up where we adjust all the requires?

shared/js/ui/base/ui-wrapper.js
```diff
@@ -23,7 +23,7 @@ export const openExtensionPage = (path) => {
     browser.tabs.create({ url: getExtensionURL(path) })
 }
 
-module.exports = {
+export default {
```
kzar, line 26: Likewise for this one, do we need to do this here, or can we save for the follow-up?

kzar (approved):
Looks great @sammacbeth.
- We already discussed the overall direction in Asana, thanks for getting rid of some of those "copy-modules" steps.
- I left a few small NIT/question comments, but I can't see anything major. So LGTM % those.
- I should say though, I'm not too experienced with Makefiles and so I only looked through the Makefile changes quite briefly. If possible, maybe get another pair of eyes on that file, ideally someone who's written more Makefiles. If not, perhaps we should just go ahead with the changes to avoid bitrot.

scripts/bundleContentScopeScripts.mjs
sammacbeth: So we can use es6 imports with `node`.

shared/js/background/before-request.js
sammacbeth: Thanks for pointing that out - there were some changes for `esbuild` which are not needed now I decided to leave that out of the scope of this PR.

PR #1739, opened 2023-02-28
Merge privacy-grade

sammacbeth:
**Reviewer:**

## Description:

- Merges the privacy-grade repo into `/packages` as an npm workspace.
- Adds a workflow to run lint and tests for privacy grade.


packages/privacy-grade/src/classes/trackers.js
```diff
@@ -0,0 +1,593 @@
...
+            'about:',
+            'moz-extension:',
+
+            // Special web protocols
+            'file:',
+            'javascript:',
+            'data:',
+            'blob:',
+            'view-source:',
+            'vbscript:',
```
jonathanKingston, line 261: Is this even supported in any browsers?

sammacbeth: This was triggered by our code scanning. No idea what the attack vector would be here.
Screenshot 2023-03-08 at 13 41 30

jonathanKingston: Ah, I mean we can leave it in but yeah this iiuc isn't relevant to our use-case at all.

jonathanKingston (approved):
LGTM, pull the trigger and see if it explodes :D

PR #1766, opened 2023-03-08
Fix C-S-S copy step to avoid copying unnecessary files.

sammacbeth:
**Reviewer:**

## Description:
Fixes an issue since #1735 when developing with content-scope-scripts. This change limits what gets copied to `shared/content-scope/scripts` so this doesn't pollute the workspace and break other build targets.

Note, you may need to run `make clean` to remove your existing `shared/content-scope-scripts` folder for this change to fix your local copy.

## Steps to test this PR:
1. `npm ci`
2. `cd /path/to/content-scope-scripts && npm ci && npm link`
3. `cd /path/to/extension`
4. `npm link @duckduckgo/content-scope-scripts`
5. `npm test`
6. `npm run dev-chrome`

These steps should complete as expected with no errors.

## Automated tests:
- [ ] Unit tests
- [ ] Integration tests

###### Reviewer Checklist:

###### PR Author Checklist:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -34,8 +34,8 @@ watch:
 unit-test: build/test/background.js build/test/ui.js build/test/shared-utils.js
 	$(KARMA) start karma.conf.js
 
-shared/content-scope-scripts: node_modules/@duckduckgo/content-scope-scripts
-	cp -r node_modules/@duckduckgo/content-scope-scripts shared/
+shared/content-scope-scripts: node_modules/@duckduckgo/content-scope-scripts $(shell find node_modules/@duckduckgo/content-scope-scripts/src -type f)
+	rsync -av node_modules/@duckduckgo/content-scope-scripts/ shared/content-scope-scripts --include="lib/***" --include="src/***" --exclude="*"
 
 .PHONY: unit-test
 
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -34,8 +34,11 @@ watch:
 unit-test: build/test/background.js build/test/ui.js build/test/shared-utils.js
 	$(KARMA) start karma.conf.js
 
-shared/content-scope-scripts: node_modules/@duckduckgo/content-scope-scripts
-	cp -r node_modules/@duckduckgo/content-scope-scripts shared/
+shared/content-scope-scripts: node_modules/@duckduckgo/content-scope-scripts $(shell find node_modules/@duckduckgo/content-scope-scripts/src -type f)
```
kzar, line 37: Can we not remove this copying entirely? (I thought we already had.)

sammacbeth: I did try to update imports in the unit-test to get the files direct from node_modules, but this breaks due to the mixing of `require` and `import` statements. Ideally, we can fix the build for the tests and switch to `import` everywhere, then this copying can be removed.

Note that this is the only place where we need to copy, I removed the ones during the grunt deprecation.

jonathanKingston (approved):

PR #1777, opened 2023-03-10
Run build checks on multiple OSes

sammacbeth:
## Description:
This should help us catch cases where the build gets broken on a specific platform.


jonathanKingston (approved):
Great! Thanks for this!

PR #1779, opened 2023-03-10
Reload extension on change

kzar:
During development, it's kind of a hassle to click to reload the
extension every time you have made a change. Why not have the
extension reload automatically instead?

1. Create a buildtime.txt file every time a new development build is
  made (meaning every time the dev target runs, except when in
  --question mode).
2. Have the extension poll for that file every five seconds,
   triggering a reload if the file contents have changed.

**Reviewer:** @sammacbeth, @jdorweiler 

## Steps to test this PR:
1. Test that development builds of the extension are reloaded every time the extension is rebuilt. Test with Firefox, Chrome and Chrome MV3.
2. Test that when `reloader=0` is passed (to either the npm or make commands)  that the extension is _not_ automatically reloaded.

Two testing notes:
- If a reloading build is installed and running, but then you remake the build as non-reloading... it will still reload the first time.
- If you make a build as non-reloading/reloading  and then remake the build as the opposite without making any other changes, the build will not be remade since none of the source files have changed.

## Automated tests:
- [ ] Unit tests
- [ ] Integration tests

###### Reviewer Checklist:

###### PR Author Checklist:


Makefile
```diff
@@ -25,13 +25,16 @@ release: clean npm $(BUILD_DIR)/public/js copy sass js
 ## dev: create a debug build for a platform in build/$BROWSER/dev
 ## specify browser=(chrome|chrome-mv3|firefox) type=dev
 dev: $(BUILD_DIR)/public/js copy sass js
+ifndef question
+	echo $(shell date "+%s") > $(BUILD_DIR)/buildtime.txt
```
kzar, line 29: I wanted to make `$(BUILD_DIR)/buildtime.txt` a proper build target, then give that the same dependencies as the `dev` target, then add `$(BUILD_DIR)/buildtime.txt` as a dependency to `dev`. But it seems that since the `copy`, `sass` and `js` targets are marked `.PHONY` they always run, so the `buildtime.txt` would always be updated.

I'm thinking that perhaps we should assign the dependencies for `sass`, `copy` and `js` to variables, then use those variables for the dependencies of `dev` + `release` instead. I think that way, the dependencies wouldn't be phony, and we could use those for the dependencies of the `$(BUILD_DIR)/buildtime.txt` target too.

What do you think?

jdorweiler: Sounds fine to me as long we're not using them outside of the makefile.

Makefile
```diff
@@ -47,14 +47,14 @@ beta-firefox: release beta-firefox-zip
...
-dev: copy build
+## specify browser=(chrome|chrome-mv3|firefox) type=dev [reloader=1]
+dev: copy build $(BUILD_DIR)/buildtime.txt
 
 .PHONY: dev
 
 ## watch: Create a debug build for a platform in build/$(browser)/dev, and keep
 ##        it up to date as files are changed.
-## specify browser=(chrome|chrome-mv3|firefox) type=dev
+## specify browser=(chrome|chrome-mv3|firefox) type=dev [reloader=1]
```
sammacbeth, line 57: `reloader=1` seems to contradict below, as we're saying here it is optional, but it's also the default. We should instead document that `reloaded=0` is the only option here that makes a difference.

Makefile
```diff
@@ -199,7 +199,12 @@ endif
 ## Extension background/serviceworker script.
 BACKGROUND_JS = shared/js/background/background.js
 ifeq ($(type), dev)
-  BACKGROUND_JS := shared/js/background/debug.js $(BACKGROUND_JS)
+  # Developer builds include the devbuilds module for debugging.
+  BACKGROUND_JS += shared/js/background/devbuild.js
+  # Unless reloader=0 is passed, they also contain an auto-reload module.
+  ifneq ($(reloader),0)
+    BACKGROUND_JS += shared/js/background/devbuild-reloader.js
```
sammacbeth, line 206: I'd prefer a solution here that avoids adding more entrypoints as this could be a blocker to switching bundlers. For example, `esbuild` does a 1-1 mapping from entry point files to bundle files. Therefore, this solution (an array of input files), would not be compatible.

shared/js/background/devbuild-reloader.js
```diff
@@ -0,0 +1,33 @@
...
+const browserWrapper = require('./wrapper')
+
+function createAlarm () {
+    browserWrapper.createAlarm('checkBuildTime', { when: Date.now() + 5000 })
+}
+
+browser.alarms.onAlarm.addListener(async alarmEvent => {
+    if (alarmEvent.name !== 'checkBuildTime') {
+        return
+    }
```
sammacbeth, line 16: Could this conflict with the centralized alarm listener in events.js? Does the alarm trigger for all listeners?

kzar: No, since the alarm name is checked in all the listeners.

Makefile
kzar: The background and some of the unit test bundles already have multiple entry points. That is something we will need to resolve anyway, if we switch to a bundler that doesn't support that.

To do that, I think we could have the Makefile generate a intermediate file that imports the modules we need, if there's not another way to do it. I think Make has support for those types of files, and even knows to clear them away again.

Doing that is IMO out of scope for this change, we should worry about it at the point we switch bundler. Happy to help with it when we get there though!

Makefile
kzar: Well, the API is that `reloader=1` enables the reloader and `reloader=0` disables it, and that by default it's enabled. As an implementation detail, we don't actually set `reloader=1` because that doesn't do anything.

I didn't add a comment to explain that, since we don't explain browser/type parameters either. But perhaps that was a mistake, I've added a comment to better explain this now. Hopefully that helps clarify?

sammacbeth (approved):

PR #1784, opened 2023-03-13
Improve Makefile handling of locally linked copies of our repositories

kzar:
- Ensure the SOURCE_FILES variable includes node_modules/@duckduckgo/*
  packages, even when they are symlinked for development.
- Ensure .git/ and node_modules/ subdirectories of symlinked
  node_modules/@duckduckgo/* packages aren't included.
- Check node_modules/@duckduckgo/ exists, before looking for its
  files. Otherwise, `make clean` fails when node_modules/ is empty.
- Avoid watching Emacs temporary files (files ending with "~") in
  SOURCE_FILES or the other similar file lists.
- Have content-scope-scripts rebuild automatically on source change,
  but only when the dependency is symlinked (when .git/) exists. That
  is important, since it turns out that the content-scope-scripts
  releases can contain source files that are newer than the included
  build files - we want to use the released build files in that case.

Notes:
 - We don't yet include recursive node_modules/@duckduckgo/*
   dependencies in either symlinked node_modules or packages/. That's
   not an issue for now, but we might need to revisit that in the
   future.
 - Editing content-scope-scripts source files will trigger some of the
   extension bundling tasks, even when we are not rebuilding
   content-scope-scripts. That isn't ideal, but a side-effect of
   getting changes to other node_modules/@duckduckgo/* packages to be
   noticed.

**Reviewer:** @jdorweiler 

## Steps to test this PR:
1. Remove node_modules/ entirely and ensure `make clean` doesn't fail.
2. `npm install && npm run dev-chrome`
3. Edit a file in `node_modules/@duckduckgo/*/*`, for example a JavaScript file in ddg2dnr and ensure that triggers Make to rebuild the extension and include your changes.
4. Edit a file in `node_modules/@duckduckgo/content-scope-scripts/src/*` and ensure that while the extension is rebuilt, content-scope-scripts wasn't rebuilt (so your changes didn't end up in the extension).
5. Kill the build watcher.
6. Edit packages.json to use a locally checked out copy of content-scope-scripts.
7. `npm install && npm run dev-chrome`
8. Edit a file in `node_modules/@duckduckgo/content-scope-scripts/src/*` and ensure that content-scope-scripts _is_ rebuilt this time. Ensure your changes to content-scope-scripts now show up in the extension.
9. Stop Ensure that changes to files ending with a "~" symbol are ignored, so do not trigger the extension to rebuild. 
10. Ensure that the output of the commands is otherwise the same.

## Automated tests:
- [ ] Unit tests
- [ ] Integration tests

###### Reviewer Checklist:

###### PR Author Checklist:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -14,8 +14,15 @@ BUILD_DIR = build/$(browser)/$(type)
 ifeq ($(browser),test)
 	BUILD_DIR := build/test
 endif
-SOURCE_FILES = $(shell find shared/ packages/ node_modules/@duckduckgo/ -type f -not -path "packages/*/node_modules/*")
-TEST_FILES = $(shell find unit-test/ -type f)
+
+SOURCE_FILES = $(shell find -L shared/ packages/ -type f -not -path "packages/*/node_modules/*" -not -name "*~")
+# If the node_modules/@duckduckgo/ directory exists, include those source files
+# in the list too.
+ifneq ("$(wildcard node_modules/@duckduckgo/)","")
+	SOURCE_FILES += $(shell find -L node_modules/@duckduckgo/ -type f -not -path "node_modules/@duckduckgo/*/.git/*" -not -path "node_modules/@duckduckgo/*/node_modules/*" -not -name "*~")
+endif
+
+TEST_FILES = $(shell find unit-test/ -type f -not -name "*~")
 
 ###--- Top level targets ---###
 ## release: create a release build for a platform in build/$BROWSER/release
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -223,6 +228,17 @@ $(BUILD_DIR)/public/js/newtab.js: $(SOURCE_FILES)
...
+# Rebuild content-scope-scripts if it's a local checkout (.git is present), but
+# not otherwise. That is important, since content-scope-scripts releases often
+# have newer source files than build files.
+CONTENT_SCOPE_SCRIPTS_SOURCE =
+ifneq ("$(wildcard node_modules/@duckduckgo/content-scope-scripts/.git/)","")
+	CONTENT_SCOPE_SCRIPTS_SOURCE += $(shell find node_modules/@duckduckgo/content-scope-scripts/src/ node_modules/@duckduckgo/content-scope-scripts/inject/ node_modules/@duckduckgo/content-scope-scripts/package.json -type f -not -name "*~")
+endif
+
+node_modules/@duckduckgo/content-scope-scripts/build/$(browser)/inject.js: $(CONTENT_SCOPE_SCRIPTS_SOURCE)
+	cd node_modules/@duckduckgo/content-scope-scripts; npm install; npm run build
```
sammacbeth, line 240: Shouldn't we split out the install and build steps here?

Makefile
```diff
@@ -14,8 +14,13 @@ BUILD_DIR = build/$(browser)/$(type)
 ifeq ($(browser),test)
 	BUILD_DIR := build/test
 endif
-SOURCE_FILES = $(shell find shared/ packages/ node_modules/@duckduckgo/ -type f -not -path "packages/*/node_modules/*")
-TEST_FILES = $(shell find unit-test/ -type f)
+
+SOURCE_FILES = $(shell find -L shared/ packages/ -type f -not -path "packages/*/node_modules/*" -not -name "*~")
+ifneq ("$(wildcard node_modules/@duckduckgo/)","")
```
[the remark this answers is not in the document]
kzar: It's checking that the directory exists. Without that, if we clear node_modules/ and then run `make clean` it fails.

Makefile
kzar: No, I think this is good enough. Also, Make goes back to the PWD after each line, so we'd have to prepend each line with the `cd node_modules/@duckduckgo/content-scope-scripts;`.

sammacbeth: I mean that `npm ci` should be a separate target, so we don't have to re-run `npm install` on every file change in that dir.

Makefile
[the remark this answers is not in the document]
sammacbeth: Ok. Could you add a comment to explain the check, as I think most readers won't know when this if is entered.

jdorweiler (approved):

Makefile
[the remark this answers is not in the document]
kzar: Done.

Makefile
kzar: Ah right, good point. Done.

PR #1802, opened 2023-03-16
Tidy up the Makefile and ensure updated files are copied again

kzar:
The Makefile had an issue whereby copied files were not re-copied when
they were updated. This could cause builds to contain out of date
files.

While I was fixing that, I decided to refactor/tidy the Makefile:
- Grouped all top-level targets together at the start of the Makefile.
- Added some missing comments.
- Used whitespace more consistently.
- Moved some out of place targets into the correct section.
- Added missing .PHONY target dependencies.

There is more work to do, in particular it would be nice to use
automatic dependency generation (e.g. `browserify --list`) in order to
reduce regeneration of the bundles.

**Reviewer:** @jdorweiler 
**CC:** @sammacbeth 

## Steps to test this PR:
1. Check build outputs are identical.
2. Check when files are changed (e.g. shared/data/constants.json) they are copied again with new builds.
3. Check that bundles are updated when files are changed too.
4. Check that files are not recopied/bundled when files _haven't_ changed.
5. Check unit and integration tests still pass.

## Automated tests:
- [x] Unit tests
- [x] Integration tests

###### Reviewer Checklist:

###### PR Author Checklist:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,89 +1,98 @@
-ITEMS   := shared/html shared/data shared/img
-
-# Workaround Browserify not following symlinks in --only.
-# TODO: Remove this once Browserify has been replaced with a different bundler.
-BROWSERIFY_GLOBAL_TARGETS = ./node_modules/@duckduckgo
-BROWSERIFY_GLOBAL_TARGETS += $(shell find node_modules/@duckduckgo/ -maxdepth 1 -type l | xargs -n1 readlink -f)
-
-###--- Binaries ---###
-SASS = node_modules/.bin/sass
-BROWSERIFY_BIN = node_modules/.bin/browserify
-BROWSERIFY = $(BROWSERIFY_BIN) -t babelify -t [ babelify --global  --only [ $(BROWSERIFY_GLOBAL_TARGETS) ] --presets [ @babel/preset-env ] ]
-ifeq ($(type),dev)
-	BROWSERIFY += -d
+###--- Shared variables ---###
+# Browser type (browser, but "chrome-mv3" becomes "chrome").
+BROWSER_TYPE = $(browser)
+ifeq ('$(browser)','chrome-mv3')
+  BROWSER_TYPE = chrome
 endif
-KARMA = node_modules/.bin/karma
 
-###--- Variables ---###
+# Output directory for builds.
 BUILD_DIR = build/$(browser)/$(type)
 ifeq ($(browser),test)
-	BUILD_DIR := build/test
+  BUILD_DIR := build/test
 endif
 
-SOURCE_FILES = $(shell find -L shared/ packages/ -type f -not -path "packages/*/node_modules/*" -not -name "*~")
+## All source files that potentially need to be bundled or copied.
+# TODO: Use automatic dependency generation (e.g. `browserify --list`) for
+#       the bundling targets instead?
+WATCHED_FILES = $(shell find -L browsers/ shared/ packages/ unit-test/ -type f -not -path "packages/*/node_modules/*" -not -name "*~")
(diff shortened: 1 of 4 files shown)


browsers/chrome-mv3/managed-schema.json
```diff
@@ -0,0 +1,10 @@
+{
```
kzar, line 1: Rather than have special cases for chrome-mv3 builds, I just added this small file from `browsers/chrome` to `browsers/chrome-mv3`.

Makefile
```diff
@@ -1,89 +1,97 @@
...
-###--- Variables ---###
+# Output directory for builds.
 BUILD_DIR = build/$(browser)/$(type)
 ifeq ($(browser),test)
-	BUILD_DIR := build/test
+  BUILD_DIR := build/test
 endif
 
-SOURCE_FILES = $(shell find -L shared/ packages/ -type f -not -path "packages/*/node_modules/*" -not -name "*~")
+## All source files that potentially need to be bundled or copied.
```
kzar, line 14: It remains to be seen if being clever with the watch dependencies would help in practice, and it would certainly add complexity + room for mistakes. I figured the simple option was better here, especially if we end up moving to a (much) faster bundler like esbuild.

Makefile
```diff
@@ -98,147 +106,129 @@ prepare-chrome-beta:
...
+# See https://www.gnu.org/software/make/manual/html_node/Empty-Targets.html
+LAST_COPY = build/.last-copy-$(browser)-$(type)
 
-$(BUILD_DIR)/dashboard: $(DASHBOARD_DIR)/
-	cp -r $< $@
+RSYNC = rsync -ra --delete --exclude="*~"
 
-$(BUILD_DIR)/web_accessible_resources: $(SURROGATES_DIR)/
-	cp -r $< $@
+$(LAST_COPY): $(WATCHED_FILES) | $(MKDIR_TARGETS)
```
kzar, line 155: The old file copying approach didn't really work. Updated files often weren't updated for example. I spent some time trying different approaches, but I found that having "proper" Makefile targets for all the copied files was a pain in practice.

Far easier to use rsync to copy all the files, then take a note (via touch) of the last copy time. That way, files are only re-copied if the last copy time is older than some of the files. Rsync takes care of updated the right files, and we do anything fancy.

package.json
```diff
@@ -9,7 +9,6 @@
     "scripts": {
         "install-ci": "make npm",
         "bundle-config": "node scripts/bundleConfig.mjs && node scripts/bundleTrackers.mjs",
-        "bundle-se": "ddg2dnr smarter-encryption shared/data/smarter_encryption.txt shared/data/bundled/smarter-encryption-rules.json",
```
jdorweiler, line 12: This was intentional?

Makefile
jdorweiler: I think `-a` for rsync handles this all for you, and only copies changed/new files over. I could be wrong though.

jdorweiler (approved):

package.json
kzar: Yea, we are turning `shared/data/smarter_encryption.txt` into `$(BUILD_DIR)/data/bundled/smarter-encryption-rules.json`. Since we can't do that directly from the package.json task, we had to create an intermediate file (which then ended up in the builds, which wasn't great). Better to just call ddg2dnr directly from the Makefile and create the file we want directly.

Makefile
kzar: Yea, and we also needed a way to know if copied files needed to be copied again. Like you say, rsync knows to only copy changed files, but we didn't want the watcher to constantly call rsync on every loop.

PR #1805, opened 2023-03-17
Use parent puppeteer version in privacy-grade

sammacbeth:
The privacy-grade tests were using an ancient puppeteer version, but we can't upgrade to the latest without breaking our extension integration tests. We can just remove puppeteer from the privacy-grade `package.json` so it will inherit the version that the extension uses.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -14836,7 +14836,6 @@
                 "grunt-exec": "3.0.0",
                 "grunt-karma": "4.0.2",
                 "load-grunt-tasks": "3.5.2",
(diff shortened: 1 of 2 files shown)


kzar (approved):
Assuming node modules are inherited for packages/ like you describe (I've not checked that.) then sounds good to me.

sammacbeth:
@kzar Yes, that's a nice feature of the monorepo setup. This is also covered by the Privacy Grade CI checks (which use puppeteer), and have passed.

PR #1806, opened 2023-03-20
Add ddg2dnr as a sub-package

sammacbeth:
Merges https://github.com/duckduckgo/ddg2dnr as a subproject of this repo.


packages/ddg2dnr/tsconfig.json
shakyShane, line 1: @sammacbeth  I would double-check that the Typescript setup is working as you expect. 

There tends to be 2 paths to go down during a migration like this

1) You can exclude type-checking the `packages/` part from the root tsconfig.json, and then ensure that `tsc` is also run for the workspace

or 

2) You configure the root-level tsconfig to also check child packages (via regular `include: ['...']`)

The reason I mention this, is that from the current PR I don't think the child `tsconfig.json` will work exactly as you imagine (eg: it's rules won't override the top level ones).

--

A 3rd option, which might not make sense here, is to use the recommended 'project references' setup, but that brings a whole heap of other issues too, like needing to compile `.d.ts` files for each project etc. (I've just beedn down this rabbit hole)

shakyShane (changes requested):

kzar (approved):

packages/ddg2dnr/tsconfig.json
```diff
@@ -0,0 +1,25 @@
...
+    "maxNodeModuleJsDepth": 0
+  },
+  "include": [
+      "*.js",
+      "**/*.js",
+      "node_modules/@duckduckgo/privacy-grade/src/classes/trackers.js"
+  ],
+  "types": [
+      "chrome"
+  ]
```
shakyShane, line 24: This is incorrect (should be inside compilerOptions) - but actually it's not needed at all - TS will auto-include npm dependencies that are in `node_modules`
```suggestion
```

tsconfig.json
```diff
@@ -32,7 +32,7 @@
       "shared/js/background/events.js",
       "shared/js/background/onboarding.js",
       "shared/js/devtools/list-editor.js",
-      "shared/js/devtools/panel.js"
+      "shared/js/devtools/panel.js",
   ],
     "types": [
         "chrome"
```
shakyShane, line 38: as before, types: ["chrome"] can be removed

packages/ddg2dnr/tsconfig.json
```diff
@@ -0,0 +1,25 @@
...
+    "checkJs": true,
+    "noEmit": true,
+    "allowSyntheticDefaultImports": true,
+    "resolveJsonModule": true,
+    "declaration": true,
+    "strict": true,
+    "noImplicitAny": false,
+    "strictNullChecks": true,
+    "maxNodeModuleJsDepth": 0
+  },
```
shakyShane, line 16:
```suggestion
  "extends": "../../tsconfig.json",
```

shakyShane (changes requested):

packages/ddg2dnr/tsconfig.json
shakyShane: @sammacbeth we can discourage config drift by sharing the compilerOptions from the root

packages/ddg2dnr/tsconfig.json
shakyShane: @sammacbeth Sorry if this wasn't part of the PR, but we might as well fix now :)

shakyShane (approved):

PR #1807, opened 2023-03-20
Replace browserify with esbuild

sammacbeth:
Replaces browserify with esbuild for all extension bundles.


packages/privacy-grade/src/classes/trackers.js
```diff
@@ -162,7 +162,7 @@ class Trackers {
 
             // take identifier from first line
             const pattern = firstLine.split(' ')[0].split('/')[1]
-            const b64surrogate = Buffer.from(lines.join('\n').toString(), 'binary').toString('base64')
```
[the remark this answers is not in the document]
sammacbeth: I think this is basically unused, as `lines` always comes through as an empty list. The `surrogates.txt` file that is being parsed here now only has the surrogate names, and the script content is now bundled via the `tracker-surrogates` dependency.

shared/js/background/devbuild.js
```diff
@@ -17,26 +17,31 @@ const { TabState } = require('./classes/tab-state')
...
-    utils,
-    setListContents,
-    getListContents,
-    companies: Companies,
-    ntts: createNewtabTrackerStatsDebugApi()
-}
+// @ts-ignore
+// RELOADER && require('./devbuild-reloader')
+
+export default function setUpDevbuild () {
```
jonathanKingston, line 23: How does this get called?

sammacbeth: In background.js - I've updated the naming to match the usage there, so it's clearer.

kzar (approved):

unit-test/background/classes/tab.js
```diff
@@ -15,7 +14,7 @@ let tab
 describe('Tab', () => {
     describe('updateSite()', () => {
         beforeEach(() => {
-            spyOn(browserWrapper, 'getExtensionId').and.returnValue('sdf')
+            // spyOn(browserWrapper, 'getExtensionId').and.returnValue('sdf')
```
kzar, line 17: Is this a typo? Or can we just remove this?

Makefile
```diff
@@ -67,7 +67,9 @@ watch:
 .PHONY: watch
 
 ## unit-test: Run the unit tests.
-unit-test: build/test/background.js build/test/ui.js build/test/shared-utils.js
+ESBUILD_TESTS = unit-test/background/*.js unit-test/background/**/*.js unit-test/ui/**/*.js unit-test/shared-utils/*.js
```
kzar, line 70: (Random line, GitHub isn't great for reviews.) Can you try adjusting the Make command above to remove `j4`? That causes problems, but was useful with Browserify due to speed. I'm thinking ESBuild might be fast enough we can just remove that.

unit-test/background/classes/tab.js
sammacbeth: Not a typo - commenting it out didn't affect the test. I'll delete the line fully.

Makefile
sammacbeth: Sure. Using -j4 does reduce the time for a clean build from 5s to 3s, but maybe that's fast enough?

PR #1812, opened 2023-03-21
Remove mixed import patterns

sammacbeth:
Fixes each case where we were mixing `export` statements and `module.exports` in a single file.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -2,7 +2,14 @@
     "extends": ["standard", "eslint:recommended"],
     "rules": {
         "indent": ["error", 4],
-        "no-shadow": ["error"]
+        "no-shadow": ["error"],
+        "no-restricted-syntax": [
+            "warn",
+            {
+              "selector": ":matches(ImportNamespaceSpecifier, ExportAllDeclaration, ExportNamespaceSpecifier)",
+              "message": "Prefer explicit named imports over wildcard (import * as x)"
+            }
+          ]
     },
     "env": {
         "webextensions": true,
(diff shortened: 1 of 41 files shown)


shared/js/background/devbuild-reloader.js
```diff
@@ -4,10 +4,10 @@
  */
 
 import browser from 'webextension-polyfill'
-const browserWrapper = require('./wrapper')
+import { createAlarm, getFromSessionStorage, setToSessionStorage } from './wrapper'
```
kzar, line 7: Nit: We have a convention of importing this one as browserWrapper. Might as well do that here too?
```suggestion
import * as browserWrapper from './wrapper'
```

kzar (changes requested):

shared/js/background/devbuild-reloader.js
sammacbeth: I would suggest we stop that convention and start only importing the individual library functions we need. In the long run, this enables the bundler to tree shake as it knows which functions are being used.

.eslintrc
```diff
@@ -2,7 +2,14 @@
     "extends": ["standard", "eslint:recommended"],
     "rules": {
         "indent": ["error", 4],
-        "no-shadow": ["error"]
+        "no-shadow": ["error"],
+        "no-restricted-syntax": [
```
kzar, line 6: Very cool, but please can you split this out as a separate change. It's unrelated to this PR. That way we can also change "warn" to "error" and fix all the failures at the same time.

kzar (approved):

PR #1838, opened 2023-03-29
Avoid triggering Chrome MV3 builds twice

kzar:
For Chrome MV3 builds, we need to fetch a list of domains that are
then used to generate a "Smarter Encryption" (HTTP -> HTTPS upgrade)
declarativeNetRequest ruleset. We were writing that list of domains as
a text file in the source tree, but that was confusing and could cause
a second incremental build to be triggered during development.

Let's save that text file in the build directory instead, that way
it's clearer, we avoid triggering builds twice and we can remove
the rsync exception.

**Reviewer:** @sammacbeth 

## Steps to test this PR:
1. Run `make clean`
2. Run `make dev browser=chrome-mv3 type=dev` and ensure `build/chrome-mv3/dev/data/bundled/smarter-encryption-rules.json` was created OK.
3. Run `make dev browser=chrome-mv3 type=dev` and ensure it reports "make: Nothing to be done for 'dev'."

## Automated tests:
- [ ] Unit tests
- [ ] Integration tests

###### Reviewer Checklist:

###### PR Author Checklist:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -94,7 +94,7 @@ npm:
 
 ## clean: Clear the builds and temporary files.
 clean:
-	rm -f shared/data/smarter_encryption.txt shared/data/bundled/smarter-encryption-rules.json integration-test/artifacts/attribution.json:
+	rm -f build/.smarter_encryption.txt integration-test/artifacts/attribution.json:
 	rm -rf $(BUILD_DIR)
 
 .PHONY: clean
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -94,7 +94,7 @@ npm:
 
 ## clean: Clear the builds and temporary files.
 clean:
-	rm -f shared/data/smarter_encryption.txt shared/data/bundled/smarter-encryption-rules.json integration-test/artifacts/attribution.json:
+	rm -f build/.smarter_encryption.txt integration-test/artifacts/attribution.json:
```
sammacbeth, line 97: Is the `.` here intentional? I'm not sure we need to make this file hidden, I see a use-case for having files in the `build` dir which are used as part of the build, but not included in the final output. Perhaps we could even make that more explicit with a `build/intermediates` folder that this file could be put in. WDYT?

kzar: Yea, it was intentional. I think the name makes sense, since folks who need to edit the file will know where to look, and other folks won't have extra noise in their build dir. I would be OK with a directory for such temporary files, if we keep it hidden too.

sammacbeth (approved):

PR #1847, opened 2023-04-02
Cookie stack trace

jonathanKingston:
## Description:

Goes with https://github.com/duckduckgo/content-scope-scripts/pull/349
See https://app.asana.com/0/0/1204257980027287/f

## Automated tests:
- [ ] Unit tests
- [ ] Integration tests

###### Reviewer Checklist:

###### PR Author Checklist:


Makefile
```diff
@@ -95,6 +95,7 @@ npm:
 
 ## clean: Clear the builds and temporary files.
 clean:
+	rm -rf shared/data/bundled/tracker-lookup.json
```
sammacbeth, line 98: If we're committing this to source control, why are we deleting on clean?

Either this is source control, then we don't need these makefile changes. Then we update it with `npm run bundle-config`.

Alternatively, it's not in source control, and generated with the build and not by `bundle-config`.

shared/data/etags.json
```diff
@@ -1 +1 @@
-{"config-etag":"W/\"260cf6ba6d4ac2cf2e2a5ef94c9edcdc\""}
\ No newline at end of file
+{"config-etag":"W/\"f1fa2ea9d9d27db1c7a68d7a81f6242d\""}
```
sammacbeth, line 1: Why has the etag changed when the content has not? Can we just drop these data changes from this diff?

sammacbeth (changes requested):
We're mixing two methods of updating tracker-lookup here. We should pick one. IMO it's file to continue with `tracker-lookup.json` in version control, meaning you can drop the makefile changes here (just remove the `shared/data/bundled/tracker-lookup.json` target).

Makefile
jonathanKingston: As I was getting the following with Make without it when running `npm run bundle-confiug`:
```

> bundle-config
> node scripts/bundleConfig.mjs && make shared/data/bundled/tracker-lookup.json

Config imported successfully
make: `shared/data/bundled/tracker-lookup.json' is up to date.
```

shared/data/etags.json
jonathanKingston: Me messing about, removed.

jonathanKingston:
> We're mixing two methods of updating tracker-lookup here. We should pick one. IMO it's file to continue with `tracker-lookup.json` in version control, meaning you can drop the makefile changes here (just remove the `shared/data/bundled/tracker-lookup.json` target).

Ok, moved just to be npm only.

sammacbeth (approved):

PR #1495, opened 2022-10-31
localize options page

shakyShane:
**Reviewer:** @ladamski  ?

## Description:
This was originally added to the Dashboard PR, but I think it can land separately and prevent the Dashboard PR being too large.

## Automated tests:
- [ ] Unit tests
- [ ] Integration tests

###### Reviewer Checklist:
- [ ] **Ensure the PR solves the problem**
- [ ] **Review every line of code**
- [ ] **Ensure the PR does no harm by testing the changes thoroughly**
- [ ] **Get help if you're uncomfortable with any of the above!**
- [ ] Determine if there are any quick wins that improve the implementation

###### PR Author Checklist:
- [ ] Get advice or leverage existing code
- [ ] Agree on technical approach with reviewer (if the changes are nuanced)
- [ ] Ensure that there is a testing strategy (and documented non-automated tests)
- [ ] Ensure there is a documented monitoring strategy (if necessary)
- [ ] Consider systems implications

PR #1892, opened 2023-04-12
Enable localization

sammacbeth:
Enables localization of:
- Privacy dashboard
- Settings page (supercedes #1495)
- Click to load.


shared/locales/en/feedback.json
```diff
@@ -0,0 +1,54 @@
...
+        "string_format" : "icu",
+        "translate_paths": [{
+            "path": "*/title",
+            "key": "{*}/title",
+            "instruction": "*/note"
+        }]
+    },
+    "brokenSiteLabel": {
+        "title": "Which website is broken?",
+        "note": "Input label for broken site URL"
```
GioSensation, line 12: "We're asking feedback from users. Here they are reporting a website url that is not working correctly."

shared/locales/en/feedback.json
```diff
@@ -0,0 +1,54 @@
...
+    },
+    "brokenSitePlaceholder": {
+        "title": "Copy and paste your URL",
+        "note": "Placeholder text on URL input"
```
GioSensation, line 16: "Instructions to fill an input field with a URL. Here "your URL" doesn't mean that they own the website, just that they're reporting it."

shared/locales/en/feedback.json
```diff
@@ -0,0 +1,54 @@
...
+    },
+    "feedbackHeaderLabel": {
+        "title": "What do you love? What isn't working? How could the extension be improved?",
+        "note": "Label over feedback textarea"
```
GioSensation, line 20: "Label over a feedback text area where the user can tell us about their experience with our browser extension."

shared/locales/en/feedback.json
```diff
@@ -0,0 +1,54 @@
...
+        "title": "Which features or functionality does your feedback refer to? Please be as specific as possible.",
+        "note": "Placeholder text on feedback form input."
+    },
+    "submittingFeedbackHelps": {
+        "title": "Submitting anonymous feedback helps us improve DuckDuckGo Privacy Essentials.",
+        "note": "Description at top of feedback form."
+    },
+    "reportBrokenSite": {
+        "title": "I want to report a broken site",
+        "note": "Label for broken site report toggle"
```
GioSensation, line 32: "This opens a feedback form to report a broken site. Broken here means 'not working as expected'."

GioSensation (commented):
Suggestions for better translation instructions.

shared/js/background/i18n.js
```diff
@@ -0,0 +1,9 @@
+import browser from 'webextension-polyfill'
+
+export function getUserLocale () {
+    if (!browser?.i18n) {
+        return 'en'
+    }
+    // returns browser locale with country suffix removed
+    return browser.i18n.getUILanguage().slice(0, 2)
```
ladamski, line 8: This doesn't handle the `nn` Norwegian case (it defaults to `en`), so perhaps we should map `nn` to `nb`?  @GioSensation might have a better sense of whether this make sense.

shared/js/ui/base/localize.js
```diff
@@ -1,37 +1,26 @@
...
+    mapping[locale] = mapping[locale] || {}
+    mapping[locale][namespace] = module
+    return mapping
+}, {})
 
 i18next
     .use(ICU)
     .init({
-        // debug: true,
+        debug: true,
```
ladamski, line 17: This should be removed

ladamski (changes requested):

unit-test/background/reference-tests/3p-cookies-tests.js
```diff
@@ -7,7 +7,7 @@ const tdsStorage = require('../../../shared/js/background/storage/tds').default
 const tabManager = require('../../../shared/js/background/tab-manager')
 const browserWrapper = require('../../../shared/js/background/wrapper')
 const { dropTracking3pCookiesFromResponse, dropTracking3pCookiesFromRequest } = require('../../../shared/js/background/events/3p-tracking-cookie-blocking')
-const getArgumentsObject = require('../../../shared/js/background/helpers/arguments-object')
+const { getArgumentsObject } = require('../../../shared/js/background/helpers/arguments-object')
```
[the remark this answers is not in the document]
sammacbeth: For unit-tests, `getUserLocale` will fall back to `en`, as the `browser.i18n` API is missing in the test context.

For integration tests, I think Playwright runs an `en-US` locale by default (this can be [overridden](https://playwright.dev/docs/emulation#locale--timezone) in the test config). 

Therefore, I don't think this would be an issue.

shared/js/background/i18n.js
sammacbeth: From this [Stack Overflow answer](https://linguistics.stackexchange.com/questions/36784/norwegian-translation-codes-no-nn-nb-which-to-use-on-a-website) it looks like mapping `nn` and `no` to `nb` is reasonable. I added special-case handling to do that now.

ladamski (approved):

PR #1936, opened 2023-04-27
Re-add fonts for autofill

sammacbeth:
**Reviewer:**

## Description:
In #1903 I missed that we need `woff2` fonts for autofill. This PR re-adds those fonts.

## Steps to test this PR:
1. Load a page where autofill is triggered.
2. Check that the ProximaNova font is loaded correctly in the popup.


GioSensation (approved):
I'm a little unclear why we need woff, given that woff2 is [everywhere](https://caniuse.com/woff2), but this seems fine. Haven't loaded it in the browser, though.

PR #1939, opened 2023-04-27
Migrate more tests to playwright

sammacbeth:
- AMP protection tests
- onboarding tests
- fingerprinting tests

PR #1940, opened 2023-04-27
Migrate click attribution tests to Playwright

sammacbeth:
Seperated from #1939 because this migration was a bit more complex.

### DIFF
--- a/integration-test/click-attribution.spec.js
+++ b/integration-test/click-attribution.spec.js
@@ -26,19 +26,7 @@ test.describe('Ad click blocking', () => {
     }
 
     /**
-     * @param {Puppeteer.page} page
-     * @param {string} variableName
-     * @returns {Promise<*>}
-     */
-    async function waitForVariable (page, variableName) {
-        const callback = (variableNameInstance) => globalThis[variableNameInstance]
-        await page.waitForFunction(callback, variableName, { timeout: 2000 })
-        return page.evaluate(callback, variableName)
-    }
-
-    /**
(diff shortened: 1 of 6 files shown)


integration-test/click-attribution.spec.js
```diff
@@ -39,48 +32,43 @@ describe('Ad click blocking', () => {
...
-        await expectAsync(page.waitForFunction(callback, { timeout: 2000 }, variableName))
-            .withContext(`waitForVariable for '${variableName}'`)
-            .not.toBeRejected()
+        await page.waitForFunction(callback, variableName, { timeout: 2000 })
         return page.evaluate(callback, variableName)
     }
 
     /**
      * Clicks on an item and awaits a new tab to load.
      * https://github.com/puppeteer/puppeteer/issues/3667 without this intercepting new tab requests is too late
```
jonathanKingston, line 41: Is this issue still relevant?

integration-test/click-attribution.spec.js
jonathanKingston, line 1: There's a `Puppeteer.page` type in this file still (in a comment so not important etc).

integration-test/click-attribution.spec.js
jonathanKingston: Also hanging brace here.

jonathanKingston (approved):
Looks good some comment fixing to happen :)

sammacbeth:
Merging this with a failing GPC test. I think this is failing because of config changes I made as part of #1951, and I'll fix it there.

PR #1943, opened 2023-04-28
[WIP] Send incontext_eligible pixel

GioSensation:
*Reviewer:**
**Asana:** https://app.asana.com/0/0/1204486716187197/f

## Description:
Accepts the `incontext_eligible` **temporary** pixel and sends it when needed.

- Only send for new installations
- Only send once per installation
- Add days since installation as a parameter

## Steps to test this PR:
If your dev extension was already installed no pixel should fire when focusing on a field.

After that, try removing and installing again. Now visit an Email Protection-eligible field and click on it. You should see a pixel being fired from the background page devtools. The pixel should include the `days_since_installation` parameter.

## Automated tests:
- [ ] Unit tests
- [ ] Integration tests

###### Reviewer Checklist:

###### PR Author Checklist:

### DIFF
--- a/shared/js/background/email-utils.js
+++ b/shared/js/background/email-utils.js
@@ -2,6 +2,7 @@ import browser from 'webextension-polyfill'
 
 import { getURL } from './pixels'
 import load from './load'
+import { daysInstalled } from './utils'
 const { getSetting, updateSetting } = require('./settings')
 const browserWrapper = require('./wrapper')
 const utils = require('./utils')
@@ -124,11 +125,11 @@ const fireAutofillPixel = (pixel, shouldUpdateLastUsed = false) => {
     }
 }
 
-const fireIncontextSignupPixel = (pixel) => {
+const fireIncontextSignupPixel = (pixel, params) => {
     const browserName = utils.getBrowserName() ?? 'unknown'
     if (!pixelsEnabled) return
 
-    sendPixelRequest(getFullPixelName(pixel, browserName))
+    sendPixelRequest(getFullPixelName(pixel, browserName), params)
 }
 
 /**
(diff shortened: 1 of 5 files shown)


shared/js/background/email-utils.js
```diff
@@ -165,6 +166,15 @@ export const sendJSPixel = (options) => {
     case 'incontext_close_x':
         fireIncontextSignupPixel('incontext_close_x_extension')
         break
+    case 'incontext_eligible': {
+        const shouldFireIncontextEligibilityPixel = getSetting('shouldFireIncontextEligibilityPixel')
+        if (!shouldFireIncontextEligibilityPixel) return
+
+        const daysSinceInstallation = Math.ceil(daysInstalled())
```
GioSensation, line 173: We do `Math.ceil` because this will be used for a cutoff, so we're interested in the upper bound, not the usual rounding.

shared/js/background/events.js
```diff
@@ -40,6 +40,7 @@ async function onInstalled (details) {
         settings.updateSetting('showWelcomeBanner', true)
         if (browserName === 'chrome') {
             settings.updateSetting('showCounterMessaging', true)
+            settings.updateSetting('shouldFireIncontextEligibilityPixel', true)
```
GioSensation, line 43: Set on first installation so we avoid firing this for existing users.

shared/js/background/utils.js
```diff
@@ -440,12 +440,21 @@ export function getInstallTimestamp (atb) {
...
+    return daysInstalled(fromDate, atb) <= numberOfDays
+}
+
+/**
+ * Returns the days since installed using atb
+ * @param {number} [fromDate]
+ * @param {string} [atb]
+ * @returns {number}
+ */
+export function daysInstalled (fromDate = Date.now(), atb = settings.getSetting('atb')) {
```
GioSensation, line 452: Just moved this out so I can re-use it.

shared/js/background/utils.js
```diff
@@ -440,12 +440,21 @@ export function getInstallTimestamp (atb) {
...
+    if (!atb) return NaN
 
     const installTimestamp = getInstallTimestamp(atb)
     // If we can't get the install date, assume it wasn't installed in time period
-    if (!installTimestamp) return false
+    if (!installTimestamp) return NaN
```
GioSensation, line 457: I'm returning `NaN` here and above so I can keep the type consistency while also ensuring maximum flexibility. `NaN` is both falsy and returns false for all math comparisons, so if you do `5 < NaN` or `5 > NaN` they're both false, which match what you would expect in these cases.

jonathanKingston: Hot new type just dropped.

alistairjcbrown (approved):
LGTM :shipit:  (once autofill update has been released and bumped in `package.json)

GioSensation:
Autofill reference updated to exact version. Merging now.

PR #1951, opened 2023-05-02
Disable all protections for 'special' pages (e.g. localhost)

sammacbeth:
**Reviewer:**

## Description:
Ensure protections are fully disabled on localhost.


integration-test/request-blocking.spec.js
```diff
@@ -20,14 +21,14 @@ async function runRequestBlockingTest (page) {
         }
         pageRequests.push({
             url: req.url(),
-            mathod: req.method(),
```
jonathanKingston, line 23: Has this always been broken?

sammacbeth: Yep. I'm not sure we're checking this in the tests

jonathanKingston (approved):
looks good, thanks for fixing!

PR #1992, opened 2023-05-22
Allow message handlers to be registered dynamically

sammacbeth:
## Description:
Adds an explicit default export from `message-handlers.js` that contains the default set of message handlers. We then also export a function that allows new handlers to be registered.

This will enable us to move handlers out of the central message handler file, and register handlers per-feature.

### DIFF
--- a/shared/js/background/events.js
+++ b/shared/js/background/events.js
@@ -5,7 +5,7 @@
  * if we do too much before adding it
  */
 import browser from 'webextension-polyfill'
-import * as messageHandlers from './message-handlers'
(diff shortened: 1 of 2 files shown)


kzar (approved):
Looks good, splitting the message handlers out of this file is something I hoped we'd do and this approach seems pragmatic.

Perhaps add a comment at the top of message-handlers.js (or above the list of default handlers) to advise folks to use `registerMessageHandler()` in their feature's code instead though. (And maybe to explain that we're migrating away from these default handlers all living in this file.)

PR #2053, opened 2023-06-14
Do not strip subdomains when referrer trimming

sammacbeth:
We're removing the eTLD+1 referrer trimming feature for breakage and party reasons.


SlayterDev (approved):
Looks good to me but unit tests are failing here.

PR #2067, opened 2023-06-22
Fire button in the extension

sammacbeth:
Depends on https://github.com/duckduckgo/privacy-dashboard/pull/43


shared/js/background/email-utils.js
```diff
@@ -91,13 +89,6 @@ export const getAddresses = () => {
     }
 }
 
-function sendPixelRequest (pixelName, params = {}) {
```
kzar, line 94: Would be nice to split this part of the change out, but I don't insist if you'd rather not.

shared/js/background/features/fire-button.js
```diff
@@ -0,0 +1,305 @@
+import browser from 'webextension-polyfill'
+import { registerMessageHandler } from '../message-handlers'
+import { getCurrentTab } from '../utils'
+import { getExtensionURL } from '../wrapper'
+import { getDomain, parse } from 'tldts'
+import { sendPixelRequest } from '../pixels'
+
+/**
+ * @typedef {object} BurnConfig
```
kzar, line 9: Could you add some more context to this type comment? In particular the since and origins variables.

shared/js/background/features/fire-button.js
```diff
@@ -0,0 +1,305 @@
...
+     */
+    async clearTabs (closeTabs = true, origins) {
+        // gather all non-pinned tabs
+        const openTabs = (await browser.tabs.query({
+            pinned: false
+        })).filter(tabMatchesHostFilter(origins))
+        const removeTabIds = openTabs.map(t => t.id || 0)
+        // clear adclick attribution data
+        removeTabIds.forEach((tabId) => {
+            const tab = this.tabManager.tabContainer[tabId]
```
kzar, line 128: Nit:
```suggestion
            this.tabManager.tabContainer[tabId]?.adClick = null
```

shared/js/background/features/fire-button.js
```diff
@@ -0,0 +1,305 @@
...
+            url: getExtensionURL('/html/fire.html')
+        })
+    }
+
+    /**
+     * Get options to display in the burn modal
+     * @returns {Promise<import('@duckduckgo/privacy-dashboard/schema/__generated__/schema.types').FireButtonData>}
+     */
+    async getBurnOptions () {
+        const ONE_HOUR = 60 * 60 * 1000
```
kzar, line 153: Nit: I'd rename ONE_HOUR_MS and put it at the top of the file personally.

shared/js/background/features/fire-button.js
```diff
@@ -0,0 +1,305 @@
...
+            },
+            descriptionStats: {
+                ...defaultStats,
+                duration: 'hour'
+            }
+        })
+        options.push({
+            name: 'Last24Hour',
+            options: {
+                since: Date.now() - (24 * ONE_HOUR)
```
kzar, line 212: Nit: Perhaps add ONE_DAY_MS, ONE_WEEK_MS etc constants at the top of the file too?

shared/js/background/features/fire-button.js
```diff
@@ -0,0 +1,305 @@
...
+    }
+}
+
+/**
+ * Given a URL, return a list of origins for that URL, including both secure and insecure contexts,
+ * and using all subdomains up to eTLD+1.
+ * @param {string} url
+ * @returns {string[]} List of origins
+ */
+export function getOriginsForUrl (url) {
```
kzar, line 274: Does the tldts library not have a similar utility?

shared/js/background/background.js
```diff
@@ -25,3 +28,16 @@ require('./script-injection')
...
+ *  fireButton?: FireButton;
+ * }}
+ */
+const features = {}
+
+if (getBrowserName() !== 'moz') {
+    features.fireButton = new FireButton({ settings, tabManager })
+}
+console.log('Loaded features:', features)
+self.features = features
```
kzar, line 43: How come we assign it to `self.features`?

kzar (commented):
I've not tested the changes, but they look great overall. Just some small questions + suggestions.

browsers/chrome/manifest.json
```diff
@@ -84,7 +84,11 @@
         "tabs",
         "storage",
         "<all_urls>",
-        "alarms"
+        "alarms",
+        "cookies"
```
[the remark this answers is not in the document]
sammacbeth: Yes. We already added that permission in MV3 with no issues. In the documentation, it is stated that this permission can be added without a prompt.

integration-test/fire-button.spec.js
```diff
@@ -0,0 +1,292 @@
...
+        })
+        {
+            const { options } = await fireButton.evaluate(f => f.getBurnOptions())
+            expect(options.every(o => o.descriptionStats.openTabs === 0)).toBeTruthy()
+            expect(options.every(o => o.descriptionStats.pinnedTabs === 0)).toBeTruthy()
+        }
+    })
+
+    test.describe('burn', () => {
+        // Skip these tests on MV3, as we don't get browsingData permissions automatically granted.
```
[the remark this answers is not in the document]
sammacbeth: We need to call `chrome.permissions.request` in the test setup to be granted the optional `browsingData` permission. In normal circumstances, this API must be called as a result of a user action. However, calling this in the MV2 extension test 'just-works'. With the MV3 extension, this request is rejected. This means we can't run these tests in MV3, without a way of simulating a user action.

package.json
```diff
@@ -73,7 +73,7 @@
         "@duckduckgo/content-scope-scripts": "github:duckduckgo/content-scope-scripts#4.22.0",
         "@duckduckgo/ddg2dnr": "file:packages/ddg2dnr",
         "@duckduckgo/jsbloom": "^1.0.2",
-        "@duckduckgo/privacy-dashboard": "github:duckduckgo/privacy-dashboard#1.5.1",
+        "@duckduckgo/privacy-dashboard": "github:duckduckgo/privacy-dashboard#45c812553e38d6d2cd9ceffc11d2cb53f53f3512",
```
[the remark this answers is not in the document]
sammacbeth: No, it will be updated once the dashboard PR is merged.

package.json
```diff
@@ -85,6 +85,7 @@
         "i18next-icu": "^2.3.0",
         "is-plain-object": "5.0.0",
         "jquery": "3.7.0",
+        "lottie-web": "^5.12.2",
         "normalize.scss": "0.1.0",
         "privacy-test-pages": "github:duckduckgo/privacy-test-pages#1.2.0",
```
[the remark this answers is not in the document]
sammacbeth: This is the lottie animation data file for the burn animation. I pulled it out of the macOS repo where it had this name. I'll rename it to something a bit more normal.

shared/js/background/features/fire-button.js
sammacbeth: You can't use optional chaining on the left-hand side of an expression in JS: `SyntaxError: Invalid left-hand side in assignment`

shared/js/background/features/fire-button.js
sammacbeth: These values are only used in a single function call, so I think keeping them in there makes sense, so everything is self-contained. I did move them all to the top of the function though, so we define all the durations up-front.

shared/js/background/features/fire-button.js
sammacbeth: No, though we use their extraction of subdomains as part of this function.

shared/js/background/background.js
sammacbeth: This is for debugging (so the class instance can be accessed in the extension debug console), and for integration tests, where we call the class functions.

kzar (approved):

PR #2093, opened 2023-07-05
Add type declarations for build flags injected by ESBuild

sammacbeth:
**Reviewer:**

## Description:
- Adds a `.d.ts` file to declare injected build-flag variables.
- Added a flag for the build target.
- Added an if block using this flag to exclude some imports for non-MV3 builds.
- Added `--minify-syntax` option for release builds to ensure `if(false)` statements get cleaned from the build output.

Screenshot 2023-07-05 at 14 18 12


shared/js/background/background.js
```diff
@@ -22,8 +22,10 @@ import initReloader from './devbuild-reloader'
 // otherwise FF might miss the onInstalled event
 require('./events')
 const settings = require('./settings')
-require('./dnr-config-rulesets')
-require('./script-injection')
+if (BUILD_TARGET === 'chrome-mv3') {
+    require('./dnr-config-rulesets')
+    require('./script-injection')
```
shakyShane, line 27: @sammacbeth is esbuild stripping these then?

sammacbeth: Yes. These modules will be removed from the build output.

shakyShane (approved):

PR #2096, opened 2023-07-07
Cleanup unused modules

sammacbeth:
**Reviewer:**

## Description:
 - `brfs`: removing this from legacy unit-test build doesn't have an effect, so we can safely remove.
 - `jasmine-core`: Moved to be a transitive dependency of `jasmine`.
 - `load-grunt-tasks`: Only used in privacy-grade submodule.
 - `is-plain-object`: Unused.
 - `seedrandom`: Unused
 - `underscore`: Unused.


shakyShane (approved):
❤️

PR #2110, opened 2023-07-13
Only ship sourcemaps in dev builds

sammacbeth:
Sourcemaps tip the size of `background.js` to over 4mb, which the Firefox addon store refuses to analyze. Let's revert to only including source maps in dev builds.


jonathanKingston (approved):
SGTM can we explore lighter minification instead in the future?

sammacbeth:
@jonathanKingston The sourcemaps are 2.8M of the bundle size. We could use `--minify` to get the bundle, with sourcemaps, down to 3.3M as an alternative.

PR #2115, opened 2023-07-18
Fire button translations

sammacbeth:
**Reviewer:**

## Description:

- Adds options page translations for the fire button.
- Reverts #2111 - reenables the feature for all Chrome builds.
- Depends on https://github.com/duckduckgo/privacy-dashboard/pull/44


shared/locales/pl/options.json
```diff
@@ -87,5 +87,29 @@
   "invalidURL" : {
     "title" : "Nieprawidłowy adres URL",
     "note" : "The URL the user entered is not a valid website address"
+  },
+  "fireButtonHeading" : {
+    "title" : "Przycisk zabezpieczenia",
```
kdzwinel, line 92: 😬 In many places they called Fire Button something else (e.g. "Protection Button" here), while they are referring to it in other descriptions as "Fire Button" (untranslated, see the next section). Not sure if that's OK?

GioSensation: FYI, I've rejected the strings where Fire Button was translated asking if it was intended for some reason. We'll check back tomorrow.

GioSensation (approved):
Looking good. I've tested a bunch of combinations of states and languages and now we should be fine. Still unsure what to do with some of those Fire Button strings being translated.

PR #2119, opened 2023-07-20
Also treat new tab page with ATB params as 'special'

sammacbeth:
## Description:
- After clearing tabs from with the fire button, the privacy dashboard was showing a 'normal' view, rather than the 'special page' view we expect for the new tab page.
- Due to the way the redirect to the NTP happens in this case, we were picking up the URL with ATB params included.
- Our 'special page' detection code expected an exact match, so this URL wasn't detected as 'special'.
- This PR makes the matching accept parameters on the NTP url.


GioSensation (approved):
👍

PR #2120, opened 2023-07-20
Fix checksum code for browsers

sammacbeth:
## Description:
- The code to check the checksum of HTTPs bloom filters was disabled after we stopped including a `Buffer` polyfill.
- This also prevented us migrating the test for this to ESBuild.
- This PR replaces the `Buffer` usage for browser-compatible JS, so we can reenabe this check, and move the https test out of the legacy folder.


jdorweiler (approved):
👍

PR #2123, opened 2023-07-24
Extract tab lifecycle events 

sammacbeth:
**Reviewer:** @kzar 
## Description:

- Currently, events that trigger updates to the data in `tabManger` global are strewn about several parts of the codebase.
- This PR moves all these events into a single 'component', which is initialized on startup. This keeps all this code together in a single, short(ish) initialization function.
- This also updates `tabManager` initialization, to load existing tabs ASAP (it was previously delayed behind several async startup processes), which should limit startup race-conditions.
- This refactor will also help us make this code compatible with Safari in the feature, where the unsupported events here will have to be replaced.


shared/js/background/components/tab-tracking.js
```diff
@@ -0,0 +1,119 @@
...
+    /**
+     * @param {{
+    *  tabManager: TabManager;
+    * }} options
+    */
+    constructor ({ tabManager }) {
+        this.tabManager = tabManager
+        this.createdTargets = new Map()
+
+        browser.webRequest.onHeadersReceived.addListener((request) => {
```
kzar, line 23: Are we moving away from the pattern of putting all extension event listeners in events.js? Fine by me if so, but perhaps we should make that clear by adding a comment in events.js or something?

jonathanKingston: Is there a notable cost to multiple listeners? I've always avoided it; perhaps prematurely.

Either way we would now have more places we're binding to the extension APIs, would it make more sense for events.js to fire events that components can listen to?

kimmillet (approved):

shared/js/background/components/tab-tracking.js
sammacbeth: Yeah, I'd prefer to put all the code for a feature together in a single file, rather than having events in `event.js`, messages in `message-handlers.js` and so on. This makes it easier to track the surface area of each feature.

Regarding registering event handlers in multiple places - my assumption is that this shouldn't degrade performance, except perhaps in the sync case that Dave mentioned. It would be surprising if this assumption was incorrect.

jonathanKingston:
I think this is fine, just going to leave it out the latest build to reduce risk as people are away etc.

shared/js/background/events.js
```diff
@@ -284,80 +269,9 @@ browser.webNavigation.onCreatedNavigationTarget.addListener(details => {
     createdTargets.set(details.tabId, details.sourceTabId)
```
jonathanKingston, line 269: Doesn't this need to be removed?

sammacbeth: Good catch. Removed.

jonathanKingston (approved):
Cheers for the changes :)

PR #2140, opened 2023-07-25
Move last legacy unit-test to esbuild

sammacbeth:
## Description:
- The declarative-net-request unit-test was the last test stuck on browserify.
- ESbuild prevented us from stubbing the `startup.ready` function, which is used by our DNR code to wait for certain lists to be loaded.
- We can refactor that code to wait specifically for the surrogates list to be loaded, and remove the startup dependency.
-


kzar:
I also notice there are references to Browserify in packages/privacy-grade. Can those (and the  corresponding Gruntfile) be removed? If not, would removing Browserify as a dependency here break the privacy-grade package somehow?

package.json
```diff
@@ -32,19 +32,12 @@
...
-        "@babel/polyfill": "^7.10.4",
-        "@babel/preset-env": "^7.22.9",
         "@fingerprintjs/fingerprintjs": "^3.4.2",
         "@playwright/test": "^1.36.1",
         "@types/chrome": "0.0.241",
         "@types/jasmine": "^4.3.5",
         "@types/webextension-polyfill": "^0.10.1",
         "asana": "github:Asana/node-asana",
-        "babelify": "10.0.0",
-        "browserify": "^17.0.0",
```
kzar, line 47: (Random line because GitHub doesn't let me add inline comments to unchanged lines.)

Please could you also remove the "babel" section below?

sammacbeth:
Thanks @kzar. Regarding `privacy-grade`, that module transitively depends on browserify via `grunt-browserify`, so it won't get broken by this change. I'll make a follow-up PR to update the build system for that module.

kzar (approved):

PR #2141, opened 2023-07-25
Generate static MV3 rulesets for tracker blocking

kzar:
Generate static MV3 rulesets for tracker blocking

So far, we have managed to keep the necessary declarativeNetRequest
rules for tracker blocking and related features to below the 5,000
dynamic rule limit[1]. However, as we expand the block list and our
protections, the limit is approaching. At the time of writing we're up
to around 4,000 dynamic rules, and that is trending upwards.

Furthermore, for our Safari MV3 port, we will need even more rules for
our protections, since rule conditions such as requestDomains and
excludedRequestDomains are not yet supported.

Therefore, we're going to move to static declarativeNetRequest
rulesets for tracker blocking. As a first step, let's set up the
instrumentation to fetch and update the current/fallback[2] blocklists,
and to generate the corresponding static rulesets.

1 - https://developer.chrome.com/docs/extensions/reference/declarativeNetRequest/#property-MAX_NUMBER_OF_DYNAMIC_AND_SESSION_RULES
2 - During the technical design phase, we decided to add a fallback
    blocklist to MV3 builds as well as the current blocklist. That
    way, if there is a catastrophic problem, we should have a way to
    revert to an older and hopefully less broken blocklist.

**Reviewer:** @sammacbeth 
**CC:** @jdorweiler 

## Steps to test this PR:
- Check the `npm run bundle-config` command works as before.
- Check that the `npm run update-mv3-blocklists` successfully updates etags.json with the latest current/fallback blocklist etag. If you tweak one of the etags in `shared/data/etags.json` before running the command you should see the command updating it back, but otherwise it should say that the lists are up to date already.
- Check that building the extension for Chrome MV3 (e.g. `npm run dev-chrome-mv3`) successfully fetches the blocklists and generates the ruleset files in `build/chrome-mv3/dev/data/bundled`:
  - `current-mv3-tds.json` - Current blocklist
  - `fallback-mv3-tds.json` - Fallback blocklist
  - `current-ctl-allowing-rules.json`, `fallback-ctl-allowing-rules.json` - CTL allowing declarativeNetRequest rules by rule action for the corresponding blocklist.
  - current-tds-rules.json`, `fallback-tds-rules.json` - declarativeNetRequest rules for tracker blocking for the corresponding blocklist.  

## Automated tests:
- [ ] Unit tests
- [ ] Integration tests

###### Reviewer Checklist:

###### PR Author Checklist:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -296,8 +296,48 @@ ifeq ('$(browser)','chrome-mv3')
   BUILD_TARGETS += $(BUILD_DIR)/data/bundled/smarter-encryption-rules.json
 endif
 
+## Generate Tracker Blocking declarativeNetRequest rulesets for MV3 builds.
+
+# Necessary since "&:" grouped target syntax is not supported currently with
+# versions (< 4.3) of Make shipped with macOS. See $(LAST_COPY) comment above
+# for more context.
+LAST_MV3_BLOCKLIST_FETCH = build/.last-mv3-blocklist-fetch
+LAST_MV3_RULESET_BUILD = build/.last-mv3-ruleset-build-$(browser)-$(type)
+
+$(LAST_MV3_BLOCKLIST_FETCH): shared/data/etags.json
+	node scripts/fetchMV3Blocklists.mjs
+	touch $@
+
+.SECONDARY:
+$(INTERMEDIATES_DIR)/current-mv3-tds.json $(INTERMEDIATES_DIR)/fallback-mv3-tds.json: $(LAST_MV3_BLOCKLIST_FETCH)
+	touch $@
+
+$(BUILD_DIR)/data/bundled/current-mv3-tds.json: $(INTERMEDIATES_DIR)/current-mv3-tds.json
+	cp $< $@
+
+$(BUILD_DIR)/data/bundled/fallback-mv3-tds.json: $(INTERMEDIATES_DIR)/fallback-mv3-tds.json
+	cp $< $@
+
+$(LAST_MV3_RULESET_BUILD): $(INTERMEDIATES_DIR)/current-mv3-tds.json $(INTERMEDIATES_DIR)/fallback-mv3-tds.json $(INTERMEDIATES_DIR)/surrogates.json
+	npx ddg2dnr tds $(INTERMEDIATES_DIR)/current-mv3-tds.json $(INTERMEDIATES_DIR)/surrogates.json \
+	  $(BUILD_DIR)/data/bundled/current-tds-rules.json $(BUILD_DIR)/data/bundled/current-ctl-allowing-rules.json
+	npx ddg2dnr tds $(INTERMEDIATES_DIR)/fallback-mv3-tds.json $(INTERMEDIATES_DIR)/surrogates.json \
+	  $(BUILD_DIR)/data/bundled/fallback-tds-rules.json $(BUILD_DIR)/data/bundled/fallback-ctl-allowing-rules.json
+	touch $@
+
+# TODO: Uncomment this for the switch to static tracker blocking declarativeNetRequest rulesets.
+# ifeq ('$(browser)','chrome-mv3')
+#   BUILD_TARGETS += $(LAST_MV3_RULESET_BUILD)
+#   BUILD_TARGETS += $(BUILD_DIR)/data/bundled/current-mv3-tds.json $(BUILD_DIR)/data/bundled/fallback-mv3-tds.json
+# endif
+
+# Generate the list of "surrogate" (stub) scripts.
 $(BUILD_DIR)/data/surrogates.txt: $(BUILD_DIR)/web_accessible_resources $(LAST_COPY)
 	node scripts/generateListOfSurrogates.js -i $</ > $@
+
+.SECONDARY:
+$(INTERMEDIATES_DIR)/surrogates.json: $(BUILD_DIR)/web_accessible_resources $(LAST_COPY)
+	node scripts/generateListOfSurrogates.js --json -i $</ > $@
 
 BUILD_TARGETS += $(BUILD_DIR)/data/surrogates.txt
 
--- a/packages/ddg2dnr/README.md
+++ b/packages/ddg2dnr/README.md
@@ -28,7 +28,7 @@ Generate the TDS ruleset:
 
 ```bash
 npm run tds ../tds-input.json ../supported-surrogates-input.json ../tds-ruleset-output.json \
-        [../match-details-by-rule-id-output.json]
+        [../allowing-rules-by-ctl-action-output.json] [../match-details-by-rule-id-output.json]
 ```
 
 Note:
--- a/scripts/updateMV3Blocklists.mjs
+++ b/scripts/updateMV3Blocklists.mjs
@@ -1,0 +1,52 @@
+import { dirname } from 'path'
+
+import { readEtags, writeEtags } from './utils.mjs'
+
+function listConstants (listName) {
+    return {
+        url: `https://staticcdn.duckduckgo.com/trackerblocking/v6/${listName}/extension-mv3-tds.json`,
+        path: `./build/.intermediates/${listName}-mv3-tds.json`,
+        etagKey: `${listName}-mv3-tds-etag`
+    }
+}
+
+async function fetchBlocklistEtag (url) {
+    const response = await fetch(url)
+    if (!response.ok) {
+        throw new Error('Failed to fetch MV3 blocklist from ' + url)
+    }
+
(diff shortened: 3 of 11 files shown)


packages/ddg2dnr/README.md
```diff
@@ -28,7 +28,7 @@ Generate the TDS ruleset:
 
 ```bash
 npm run tds ../tds-input.json ../supported-surrogates-input.json ../tds-ruleset-output.json \
-        [../match-details-by-rule-id-output.json]
+        [../allowing-rules-by-ctl-action-output.json] [../match-details-by-rule-id-output.json]
```
sammacbeth, line 31: Are these both individually optional, or you can either specify none or both? If the latter, I think it makes more sense to use a single set of square brackets over both terms. 

If they are individually optional, it's not actually possible to provide `match-details-by-rule-id-output` without `allowing-rules-by-ctl-action-output`.

scripts/updateMV3Blocklists.mjs
```diff
@@ -0,0 +1,64 @@
...
+import { mkdirSync, writeFileSync } from 'fs'
+
+import { readEtags, writeEtags } from './utils.mjs'
+
+const currentTdsUrl = 'https://staticcdn.duckduckgo.com/trackerblocking/v6/current/extension-mv3-tds.json'
+const fallbackTdsUrl = 'https://staticcdn.duckduckgo.com/trackerblocking/v6/fallback/extension-mv3-tds.json'
+const currentTdsPath = './build/.intermediates/current-mv3-tds.json'
+const fallbackTdsPath = './build/.intermediates/fallback-mv3-tds.json'
+const currentTdsEtagKey = 'current-mv3-tds-etag'
+const fallbackTdsEtagKey = 'fallback-mv3-tds-etag'
```
sammacbeth, line 11: This seems like a lot of duplication. Couldn't we have these args provided via the command line and then run the script twice (once for `current` then once for `fallback`)?

Makefile
```diff
@@ -315,9 +315,40 @@ ifeq ('$(browser)','chrome-mv3')
...
+	node scripts/fetchMV3Blocklists.mjs
+	touch $(MV3_BLOCKLIST_TARGETS)
+
+$(BUILD_DIR)/data/bundled/current-mv3-tds.json: $(INTERMEDIATES_DIR)/current-mv3-tds.json
+	cp $< $@
+
+$(BUILD_DIR)/data/bundled/fallback-mv3-tds.json: $(INTERMEDIATES_DIR)/fallback-mv3-tds.json
+	cp $< $@
+
+$(MV3_CURRENT_RULESET_TARGETS) &: $(INTERMEDIATES_DIR)/current-mv3-tds.json $(INTERMEDIATES_DIR)/surrogates.json
```
sammacbeth, line 334: I'm not sure what the `&` does here, but it spits out a load of warnings for me when I run `make` (on a mac):
```
Makefile:335: warning: overriding commands for target `&'
Makefile:325: warning: ignoring old commands for target `&'
Makefile:338: warning: overriding commands for target `&'
Makefile:335: warning: ignoring old commands for target `&'
```

Makefile
```diff
@@ -315,9 +315,40 @@ ifeq ('$(browser)','chrome-mv3')
...
+	npx ddg2dnr tds $^ $(MV3_CURRENT_RULESET_TARGETS)
+
+$(MV3_FALLBACK_RULESET_TARGETS) &: $(INTERMEDIATES_DIR)/fallback-mv3-tds.json $(INTERMEDIATES_DIR)/surrogates.json
+	npx ddg2dnr tds $^ $(MV3_FALLBACK_RULESET_TARGETS)
```
sammacbeth, line 338: These commands launch `chromium` for me during the build, which was unexpected. Is that meant to happen? If so, can we avoid it?

sammacbeth (changes requested):
This looks generally ok, but I think we can tidy up some of the duplication in scripts.

I only see the logic for fetching lists here though, but these lists are not yet used by the MV3 extension. I guess that will come in a follow-up, but can we avoid putting the static lists in the build until they're in use? Otherwise we're adding 10M of bloat to the release which is not used.

packages/ddg2dnr/README.md
kzar: That's right, you have to specify the first if you want to specify the second. I think that's fine honestly.

kzar:
> I only see the logic for fetching lists here though, but these lists are not yet used by the MV3 extension. I guess that will come in a follow-up, but can we avoid putting the static lists in the build until they're in use? Otherwise we're adding 10M of bloat to the release which is not used.

Yea, I did explain that in the README and commit message, but we're not using the files yet. If you prefer, I can keep it disabled for now.

Makefile
kzar: The `&:` syntax is for a [rule with grouped target](https://www.gnu.org/software/make/manual/html_node/Multiple-Targets.html#Rules-with-Grouped-Targets). It's used for when one command is used to generate multiple target files at the same time, when used Make knows to only run the command once to generate all the files.

I will try to debug what's going wrong. What version of Make are you running?

scripts/updateMV3Blocklists.mjs
kzar: We could run the script twice to update the two files like you describe, or we could do it like this. I don't think it makes too much difference, but I went with this way for a couple of reasons:
1. If the update for one blocklist fails, we're not updating the etag for the other one.
2. We're not reading and writing to the etag.json file extra times.

Makefile
kzar: But the `npx ddg2dnr smarter-encryption $< $@` line doesn't for you?

kzar: Oh right, I remember. We're using Chrome via Puppeteer to access the [isRegexSupported](https://developer.chrome.com/docs/extensions/reference/declarativeNetRequest/#method-isRegexSupported) API. That's non-trivial to duplicate ourselves.

Makefile
sammacbeth:
```bash
$ make --version
GNU Make 3.81
Copyright (C) 2006  Free Software Foundation, Inc.
This is free software; see the source for copying conditions.
There is NO warranty; not even for MERCHANTABILITY or FITNESS FOR A
PARTICULAR PURPOSE.

This program built for i386-apple-darwin11.3.0
```

scripts/updateMV3Blocklists.mjs
sammacbeth: Making this atomic makes sense, but it would be nice if the file was more extendable, if we wanted to add more blocklists here. e.g. we could specify an array of lists at the top and fetch them all:
```javascript
const lists = [{
  url: 'https://staticcdn.duckduckgo.com/trackerblocking/v6/current/extension-mv3-tds.json',
  downloadPath: './build/.intermediates/current-mv3-tds.json',
  etagKey: 'current-mv3-tds-etag' 
}, {
 ...
}]
```

sammacbeth:
> > I only see the logic for fetching lists here though, but these lists are not yet used by the MV3 extension. I guess that will come in a follow-up, but can we avoid putting the static lists in the build until they're in use? Otherwise we're adding 10M of bloat to the release which is not used.
> 
> Yea, I did explain that in the README and commit message, but we're not using the files yet. If you prefer, I can keep it disabled for now.

Ok, let's keep this out of the build output (at least for release builds) until these files are being used.

scripts/updateMV3Blocklists.mjs
kzar: Done.

Makefile
kzar: Seems that support for the `&:` syntax was added with 4.3, but macOS is stuck with an outdated version of Make due to licensing 🤦. I have avoided that by using [the empty target trick](https://www.gnu.org/software/make/manual/html_node/Empty-Targets.html). Is that working better for you now?

kzar:
> > > I only see the logic for fetching lists here though, but these lists are not yet used by the MV3 extension. I guess that will come in a follow-up, but can we avoid putting the static lists in the build until they're in use? Otherwise we're adding 10M of bloat to the release which is not used.
> > 
> > 
> > Yea, I did explain that in the README and commit message, but we're not using the files yet. If you prefer, I can keep it disabled for now.
> 
> Ok, let's keep this out of the build output (at least for release builds) until these files are being used.

OK, I will comment the relevant lines out before merging. I have left them uncommented for now while we test. (Also I have updated the README already.)

sammacbeth (approved):
Thanks for making those changes!

kimmillet:
Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

PR #2156, opened 2023-07-31
Update the uninstall URL each time the ATB setting is updated

kzar:
When a user uninstalls the extension, an survey is displayed which
aims to find out what the user's reason was. Some parameters are
included with that request (e.g. the browser name and version) to
help us diagnose issues. The ATB setting is included as well, so it's
important that we update the uninstall URL each time the ATB setting
is changed. So far, changes to the ATB setting were only reflected
after the next time the updateUninstallURL event fired, which led to a
delay of up to 10 minutes.

**Reviewer:**

## Automated tests:
- [ ] Unit tests
- [ ] Integration tests

###### Reviewer Checklist:

###### PR Author Checklist:


shared/js/background/settings.js
```diff
@@ -124,6 +126,9 @@ function updateSetting (name, value) {
 
     settings[name] = value
     syncSettingTolocalStorage()
+    setTimeout(() => {
```
jonathanKingston, line 129: What's the timeout for?

kzar: Since the events are dispatched synchronously otherwise. I figure that if someone updates a setting and waits for the update to finish, it would be undesirable to also wait for all the update listeners to finish. I don't think it's too important for now, but I wanted to avoid a foot-gun for the next person using this.

jonathanKingston: This won't guarantee it's stored though, so you might end up creating more problems.

I would suggest making `syncSettingTolocalStorage` return the promise and await/then it here if you want to provide any async guarantee.

unit-test/background/atb.js
```diff
@@ -287,3 +287,63 @@ describe('complex install workflow cases', () => {
...
+        expect(uninstallUrl).toEqual(null)
+
+        // Before the atb is set.
+        uninstallUrl = await atb.getSurveyURL()
+        checkUninstallUrl(null, null)
+
+        // After the atb is set.
+        stubLoadJSON({ returnedAtb: 'v119-8' })
+        await atb.setInitialVersions()
+        await new Promise(resolve => setTimeout(resolve, 0))
```
jonathanKingston, line 326: This could race, would listening for the even be cleaner?

jdorweiler (approved):

shared/js/background/settings.js
kzar: Good idea, Done.

unit-test/background/atb.js
kzar: I thought about that too. I think this is OK, since for the unit tests the event is firing synchronously. I'm going to leave it alone for now considering we're in a rush here, but if the test starts failing we can always adjust it later.

PR #2157, opened 2023-08-01
Referrer cleanup and test fixes to merge latest C-S-S

jonathanKingston:
**Reviewer:** @muodov 

## Description:

Fixing up issue caused in the latest C-S-S referrer changes. This is largely fixing the test itself that is very coupled with the integration too much.

However we needed to disable the feature in the arguments object based on checking the referrer of the tab when it's safelisted. This was an actual bug with preventing us to disable the feature.

https://app.asana.com/0/312629933896096/1205182548153047/f

## Automated tests:
- [ ] Unit tests
- [ ] Integration tests

###### Reviewer Checklist:

###### PR Author Checklist:

### DIFF
--- a/shared/js/background/helpers/arguments-object.js
+++ b/shared/js/background/helpers/arguments-object.js
@@ -25,10 +25,17 @@ export function getArgumentsObject (tabId, sender, documentUrl, sessionKey) {
         site.enabledFeatures = site.enabledFeatures.filter(feature => aboutBlankEnabled.includes(feature))
     }
 
+    site.enabledFeatures = site.enabledFeatures.filter((feature) => {
+        // Prune out the tracker allowlist feature setting as it's not needed and is large
+        if (feature === 'trackerAllowlist') return false
+
+        // Disable referrer trimming when we're not changing the referrer for the tab
+        if (feature === 'referrer' && !tab.referrer?.referrer) return false
+        return true
+    })
+
     const featureSettings = {}
     for (const feature of site.enabledFeatures) {
-        // Prune out the tracker allowlist feature setting as it's not needed and is large
-        if (feature === 'trackerAllowlist') continue
         const featureSetting = utils.getFeatureSettings(feature)
         if (Object.keys(featureSetting).length) {
             featureSettings[feature] = featureSetting
--- a/shared/js/background/helpers/arguments-object.js
+++ b/shared/js/background/helpers/arguments-object.js
@@ -17,7 +17,6 @@ export function getArgumentsObject (tabId, sender, documentUrl, sessionKey) {
     // Clone site so we don't retain any site changes
     // @ts-ignore
     const site = tabClone.site
-    const referrer = tab?.referrer || ''
     let cookie = {}
 
     // Special case for iframes that are blank we check if it's also enabled
@@ -67,7 +66,6 @@ export function getArgumentsObject (tabId, sender, documentUrl, sessionKey) {
         stringExemptionLists: utils.getBrokenScriptLists(),
         sessionKey,
         site,
-        referrer,
         platform: constants.platform,
         locale: getUserLocale(),
         assets: {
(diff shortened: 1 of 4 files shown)


shared/js/background/helpers/arguments-object.js
```diff
@@ -17,7 +17,6 @@ export function getArgumentsObject (tabId, sender, documentUrl, sessionKey) {
     // Clone site so we don't retain any site changes
     // @ts-ignore
     const site = tabClone.site
-    const referrer = tab?.referrer || ''
```
jonathanKingston, line 20: We're no longer needing this in C-S-S so let's remove it as it wasn't ever documented or intended to be required.

shared/js/background/helpers/arguments-object.js
```diff
@@ -26,10 +25,17 @@ export function getArgumentsObject (tabId, sender, documentUrl, sessionKey) {
         site.enabledFeatures = site.enabledFeatures.filter(feature => aboutBlankEnabled.includes(feature))
     }
 
+    site.enabledFeatures = site.enabledFeatures.filter((feature) => {
+        // Prune out the tracker allowlist feature setting as it's not needed and is large
+        if (feature === 'trackerAllowlist') return false
```
jonathanKingston, line 30: Moving this has the same effect as pruning below, merging it makes the below code easier to follow if feature pruning is only here.

shared/js/background/helpers/arguments-object.js
```diff
@@ -26,10 +25,17 @@ export function getArgumentsObject (tabId, sender, documentUrl, sessionKey) {
...
+
+        // Disable referrer trimming when we're not changing the referrer for the tab
+        if (feature === 'referrer' && !tab.referrer?.referrer) return false
```
jonathanKingston, line 33: When the `tab.referrer` is blank we're expecting the feature to be disabled, this happened previously inside C-S-S prior to https://github.com/duckduckgo/content-scope-scripts/commit/5492c11b5c71cb5b1f1db5a131424a51f3f977f1

This was an actual bug once we've merged that code in C-S-S

unit-test/background/reference-tests/referrer-trimming-tests.js
```diff
@@ -91,16 +93,17 @@ for (const setName of Object.keys(testSets)) {
                         enumerable: true,
                         writeable: true
                     })
-                    FakeDocument.prototype.location = {
```
jonathanKingston, line 94: Removed this purely as we're not using it, instead we're using `document.URL` inside C-S-S

unit-test/background/reference-tests/referrer-trimming-tests.js
```diff
@@ -91,16 +93,17 @@ for (const setName of Object.keys(testSets)) {
...
                     const orgDocument = globalThis.Document
 
                     // replacing real document and Document with fake ones
                     globalThis.Document = FakeDocument
                     spyOnProperty(document, 'referrer', 'get').and.returnValue(test.referrerValue)
+                    spyOnProperty(document, 'URL', 'get').and.returnValue(test.frameURL || test.siteURL)
 
-                    jsReferrerProtection.init({ referrer: tab.referrer })
+                    const args = getArgumentsObject(1, { url: test.siteURL, frameId: 0 }, test.siteURL, 'abc123')
+                    if (!isFeatureBroken(args, 'referrer')) {
```
jonathanKingston, line 104: We're doing this in other tests, it's not ideal as mentioned on MM but changing this to instantiate the whole of C-S-S is large rewrite.

We're essentially mocking the message from the extension which used the `getArgumentsObject` and calling the `isFeatureBroken` method from C-S-S to validate if we should trigger the calling of the method.

shared/js/background/helpers/arguments-object.js
muodov: Do we still need https://github.com/duckduckgo/content-scope-scripts/commit/acc38de473f162e406d51c70ce0dcac75f58bbb6 then?

muodov (approved):

shared/js/background/helpers/arguments-object.js
jonathanKingston: Yeah I think so, the browser might choose to sever the referrer itself or other extensions. It's pretty unrelated to this issue.

PR #2201, opened 2023-08-23
Group dependabot PRs

sammacbeth:
See https://github.blog/changelog/2023-06-30-grouped-version-updates-for-dependabot-public-beta/

This PR configures dependabot to group PRs for non-internal packages.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -8,6 +8,12 @@ updates:
     target-branch: "main"
     labels:
       - "dependencies"
+    groups:
+      external-dependencies:
+        exclude-patterns:
+         - "puppeteer"
+         - "@duckduckgo/*"
+         - "privacy-test-pages"
   - package-ecosystem: "github-actions"
     directory: "/"
     schedule:


.github/dependabot.yml
```diff
@@ -8,6 +8,12 @@ updates:
     target-branch: "main"
     labels:
       - "dependencies"
+    groups:
+      dev-dependencies:
+        exclude-patterns:
+         - "puppeteer"
+         - "@duckduckgo/*"
+         - "privacy-test-pages"
```
jonathanKingston, line 16: Do these apply? They're normal dependencies? Is it just belt and braces in case we move them?

jonathanKingston: Oh I guess `dev-dependencies` is just a group name rather than having distinct meaning. Should we rename the group name?

sammacbeth: Yeah, the idea is that we'll get individual PRs for `@duckduckgo/*` and `privacy-test-pages`, as those will be worth reviewing individually. Everything else will be bundled into a single version bump PR each week for `dev-dependencies`.

jonathanKingston: Yeah fair, I'd still rename the group name given the collision with npm naming but whatever works 👍🏻

jonathanKingston (approved):

PR #2306, opened 2023-10-26
Add seen before day on site breakage reports

jonathanKingston:
**Reviewer:**

## Description:

https://app.asana.com/0/inbox/1199237043596759/1205588216029401/1205802892084402

## Automated tests:
- [ ] Unit tests
- [ ] Integration tests

###### Reviewer Checklist:

###### PR Author Checklist:


shared/js/background/events.js
```diff
@@ -433,6 +434,8 @@ browserWrapper.createAlarm('clearExpiredHTTPSServiceCache', { periodInMinutes: 6
 browserWrapper.createAlarm('rotateUserAgent', { periodInMinutes: 24 * 60 })
 // Rotate the sessionKey
 browserWrapper.createAlarm('rotateSessionKey', { periodInMinutes: 24 * 60 })
+// Expire site breakage reports
+browserWrapper.createAlarm('clearExpiredBrokenSiteReportTimes', { periodInMinutes: 24 * 60 })
```
sammacbeth, line 438: I think this is not going to fire unless the browser is open for 24 hours. I think a simpler solution will be to just run a clear on extension start, then rerun whenever a broken site report is sent.

unit-test/background/reference-tests/broken-site-reporting-tests.js
```diff
@@ -62,7 +62,7 @@ for (const setName of Object.keys(testSets)) {
                 addActionRequests(test.adAttributionRequests, 'ad-attribution')
                 addActionRequests(test.noActionRequests, 'none')
 
-                breakageReportForTab({
+                await breakageReportForTab({
```
sammacbeth, line 65: Should we add a test that the parameter is not being sent on the first report, then it is present on the second?

sammacbeth (changes requested):

shared/js/background/events.js
jonathanKingston: That doesn't clear the data for local users though which was the main concern. I think on start and every 60 might be a better compromise (I think I left a comment saying it was every 60 anyway and was my intent) I think the one above should do the same too.

unit-test/background/reference-tests/broken-site-reporting-tests.js
jonathanKingston: Not as a reference test as we'd be waiting about for all other implementations to ship. Let me check if we have others already.

shared/js/background/broken-site-report.js
```diff
@@ -88,6 +88,60 @@ const requestCategoryMapping = {
...
+    return hashHex
+}
+
+async function computeLastSentDay (urlString) {
+    const url = new URL(urlString)
+    const time = new Date()
+    // Round time to nearest day
+    time.setHours(0, 0, 0, 0)
+    // Output time as a string in the format YYYY-MM-DD
+    const dayOutput = time.toISOString().split('T')[0]
```
sammacbeth, line 107: This is incorrect for timezones east of GMT. For example for me (in CEST), the date is set to 11pm yesterday. Can we just skip the `setHours` call and do `new Date().toISOString().split('T')[0]` as a one-liner?

sammacbeth (changes requested):

shared/js/background/broken-site-report.js
jonathanKingston: Love this so much haha (weird bugs are the best). Cheers will fix.

sammacbeth (approved):

PR #2309, opened 2023-10-30
Refactor TDS list loading

sammacbeth:
Simplifies the loading of blocklists and config.

### DIFF
--- a/integration-test/request-blocking.spec.js
+++ b/integration-test/request-blocking.spec.js
@@ -6,6 +6,21 @@ import { TEST_SERVER_ORIGIN } from './helpers/testPages'
 const testHost = 'privacy-test-pages.site'
 const testSite = `https://${testHost}/privacy-protections/request-blocking/`
 
+async function forDynamicRulesLoaded (backgroundPage) {
+    // We don't have anything we can listen to to know we've finished loading all rules, so just
+    // poll the dynamic rules until it looks like they've been added.
+    while (true) {
+        const ruleCount = await backgroundPage.evaluate(async () => {
+            const rules = await chrome.declarativeNetRequest.getDynamicRules()
+            return rules.length
+        })
+        if (ruleCount > 10) {
+            break
+        }
+        await new Promise(resolve => setTimeout(resolve, 100))
+    }
+}
+
 async function runRequestBlockingTest (page, url = testSite) {
     const pageRequests = []
     page.on('request', async (req) => {
@@ -43,10 +58,13 @@ async function runRequestBlockingTest (page, url = testSite) {
 }
 
 test.describe('Test request blocking', () => {
-    test('Should block all the test tracking requests', async ({ page, backgroundPage, context, backgroundNetworkContext }) => {
+    test('Should block all the test tracking requests', async ({ page, backgroundPage, context, backgroundNetworkContext, manifestVersion }) => {
         await overridePrivacyConfig(backgroundNetworkContext, 'serviceworker-blocking.json')
         await forExtensionLoaded(context)
         await forAllConfiguration(backgroundPage)
+        if (manifestVersion === 3) {
+            await forDynamicRulesLoaded(backgroundPage)
+        }
         const [testCount, pageRequests] = await runRequestBlockingTest(page)
 
         // Verify that no logged requests were allowed.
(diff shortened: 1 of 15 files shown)


shared/js/background/components/resource-loader.js
```diff
@@ -0,0 +1,198 @@
...
+
+        if (!this.remoteUrl && !this.localUrl) {
+            throw new Error('invalid config: need a remote or local URL (or both)')
+        }
+
+        this.ready = this.checkForUpdates()
+
+        if (this.updateIntervalMinutes) {
+            const alarmName = `updateResource: ${this.name}`
+            alarms.create(alarmName, { periodInMinutes: this.updateIntervalMinutes })
```
kzar, line 47: I think creating the alarm without first checking it doesn't exist might cause an exception for MV3? (Have you tested this on MV3?)

kzar: (The old code you're replacing uses the `browserWrapper.createAlarm` wrapper which takes care of that. Perhaps use that here?

integration-test/request-blocking.spec.js
```diff
@@ -6,6 +6,21 @@ import { TEST_SERVER_ORIGIN } from './helpers/testPages'
 const testHost = 'privacy-test-pages.site'
 const testSite = `https://${testHost}/privacy-protections/request-blocking/`
 
+async function forDynamicRulesLoaded (backgroundPage) {
```
kzar, line 9: I think this should have DNR in the name, like `forDynamicDNRRulesLoaded` and live in `helpers/backgroundWait.js`.

This is kind of a hack, but if it's good enough for now fair enough:
1. I can't see how better to do it without some fairly large changes.
2. The related helpers use similar hacks at the moment.
3. So long as we use this helper in the tests, changing how it works in the future won't require messing with the tests anyway.

shared/js/background/components/resource-loader.js
```diff
@@ -0,0 +1,198 @@
...
+    constructor (config, { settings }) {
+        this.settings = settings
+        this.name = config.name
+        this.remoteUrl = config.remoteUrl
+        this.localUrl = config.localUrl
+        this.updateIntervalMinutes = config.updateIntervalMinutes || 0
+        this.format = config.format || 'json'
+        this.data = null
+        /** @type {OnUpdatedCallback[]} */
+        this._onUpdatedListeners = []
```
kzar, line 37: The code this was taken from had a comment suggesting we should use [EventTarget](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget) here instead (like we do in `shared/js/background/settings.js`). I think that still stands, why not change that while refactoring this code? _(If you prefer, we also have EventEmitter2 available, but I figure EventTarget is preferable as it's a built-in API.)_

If we did that, we could then also add a second emitter for `onAfterUpdated` which would fire after all the update listeners are done. We could then use that for our tests, to avoid the need for some of the backgroundWait.js hacks. _We might want to give that emitter a name to signify it's for tests only though, e.g. `_testOnlyOnAfterUpdated`._

integration-test/request-blocking.spec.js
kzar: _(Also see this comment https://github.com/duckduckgo/duckduckgo-privacy-extension/pull/2309/files#r1537480355, I had an idea for another way to do this.)_

shared/js/background/components/resource-loader.js
sammacbeth: Good shout - I forgot about this edge case. I didn't see any exceptions in MV3 (seems like this has been fixed now in Chrome), but re-registering does push the time of the next alarm back, so this change is important to ensure those alarms actually run! Fixed in https://github.com/duckduckgo/duckduckgo-privacy-extension/pull/2309/commits/38c199def268eae31127c7dde81adc49a4b51c8d

integration-test/request-blocking.spec.js
sammacbeth: Done and moved (https://github.com/duckduckgo/duckduckgo-privacy-extension/pull/2309/commits/9d43936ea5712fe31efc1c8643e225b92ecbbc60)

shared/js/background/components/resource-loader.js
kzar: So I think `onConfigUpdate()` in `dnr-config-rulesets.js` returns a promise that resolves after the work updating the dynamic rules is done, so I think that yes the `onAfterUpdated` event could wait for all the `onUpdate` listeners to have been called and any returned Promises to have been resolved. It doesn't seem too hard, unless I'm missing something?

shared/js/background/components/resource-loader.js
```diff
@@ -0,0 +1,207 @@
...
+        if (!this.remoteUrl && !this.localUrl) {
+            throw new Error('invalid config: need a remote or local URL (or both)')
+        }
+
+        this.ready = this.checkForUpdates()
+
+        if (this.updateIntervalMinutes) {
+            const alarmName = `updateResource: ${this.name}`
+            createAlarm(alarmName, { periodInMinutes: this.updateIntervalMinutes })
+            alarms.onAlarm.addListener((alarmEvent) => {
```
kzar, line 48: Let's use `browserWrapper.createAlarm` that already takes care of the checking logic. No point duplicating it here.

kzar (changes requested):
Changes look good, but let's discuss the after update event a little more.

PR #2325, opened 2023-11-09
MV3: Request content-script config on first tick

sammacbeth:
The navigator-interface test is often flakey on MV3. This seems to be because, on a slow machine, extension startup and content-script registration are slow. The means that we actually load the test page before the content-scripts are registered.

Now that we exposed MV3 content-script registration via a component, we can wait for that to complete before loading the test page.


sammacbeth:
It seems this doesn't fully fix it - it looks like in some cases we're just too slow to get the config to C-S-S. Probably MV3 SW is slower to respond to messages than the old background.

kzar (approved):

PR #2378, opened 2023-12-08
Merge autofill code into a component

sammacbeth:
The prime motivation here was to add guards around use of the `contextMenus` API, so the extension doesn't instantly crash on platforms that don't support this API (Firefox for Android and Safari). However, I also took the liberty of refactoring the autofill code to move as much of it as possible together into a single file, as it is current spread over quite a few places, making it difficult to track what's going on.

Unfortunately, the diff is not very helpful, but all this is, is the code from `email-utils` put into a class, and all the message-handler functions, previously in `message-handlers.js` are moved to be member functions of that class. Then I also merged the initialization code from `events.js` and `startup.js` into the constructor and init functions of this class.


shared/js/background/components/email-autofill.js
```diff
@@ -0,0 +1,353 @@
...
+
+        sendPixelRequest(getFullPixelName(pixel, browserName), { duck_address_last_used: lastAddressUsedAt, cohort: userData.cohort })
+        if (shouldUpdateLastUsed) {
+            this.settings.updateSetting('lastAddressUsedAt', currentDate())
+        }
+    }
+
+    getIncontextSignupDismissedAt () {
+        const permanentlyDismissedAt = this.settings.getSetting('incontextSignupPermanentlyDismissedAt')
+        // TODO: inject this dependency (after TDS refactor lands)
```
GioSensation, line 218: There are a couple of these. I'm not quite sure who's supposed to follow-up. Is it an action item to whomever is doing that TDS refactor?

sammacbeth: That's for me, referring to https://github.com/duckduckgo/duckduckgo-privacy-extension/pull/2309

shared/js/background/components/email-autofill.js
```diff
@@ -0,0 +1,353 @@
...
+            if (alarmEvent.name === REFETCH_ALIAS_ALARM) {
+                this.fetchAlias()
+            }
+        })
+        // message handlers
+        registerMessageHandler('getAddresses', this.getAddresses.bind(this))
+        registerMessageHandler('sendJSPixel', this.sendJSPixel.bind(this))
+        registerMessageHandler('getAlias', this.getAlias.bind(this))
+        registerMessageHandler('refreshAlias', this.refreshAlias.bind(this))
+        registerMessageHandler('getEmailProtectionCapabilities', getEmailProtectionCapabilities)
```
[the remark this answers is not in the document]
sammacbeth: No, as this is a plain function, rather than a class member function.

PR #2480, opened 2024-03-19
Detect internal users and add an extra section to settings for them.

sammacbeth:
**Reviewer:**

## Description:
This mirrors how we detect internal users on other platforms (for example [apple platforms](https://github.com/duckduckgo/BrowserServicesKit/blob/main/Sources/BrowserServicesKit/InternalUserDecider/InternalUserDecider.swift#L74-L76), in order to add a `isInternalUser` flag to settings. This can be used to expose additional settings in the options menu.

## Steps to test this PR:
1. Open settings - note that the sections are as before.
2. Visit `use-login.duckduckgo.com` and login with DDG SSO.
3. Return to the settings page - there should not be a new 'Interal settings' section.


kzar (approved):
Changes LGTM, but as mentioned in Asana I want to double check about a privacy review before we merge this.

PR #2485, opened 2024-03-27
Custom config override

sammacbeth:
## Description:
Allows the config URL to be overridden for a session.
 - Also opens a connection to the debugger server to trigger updates on the config is changed.
 - Adds an option to the settings menu to allow set/unset of this config URL.

Screenshot 2024-03-27 at 13 36 45

## Steps to test this PR:
1. Visit use-login.duckduckgo.com and login to get recognised as an internal user.
2. Open the settings page.
3. The 'Custom config URL' setting should be displayed.
4. Type in a URL and hit 'Set'. The config at that URL should be loaded.
5. In settings, click 'Clear' to return back to the original config URL.

### DIFF
--- a/shared/js/ui/views/internal-options.js
+++ b/shared/js/ui/views/internal-options.js
@@ -55,6 +55,7 @@ InternalOptionsView.prototype = window.$.extend({}, ViewParent.prototype, {
             [this.$setconfigurl, 'click', this._clickSetting],
             [this.$configurl, 'input', this._onURLChange]
         ])
+        this._onURLChange()
     },
     _buttonState: function () {
         const currentValue = this.model.debuggingSettings?.configURLOverride
--- a/shared/js/ui/views/internal-options.js
+++ b/shared/js/ui/views/internal-options.js
@@ -32,7 +32,7 @@ function template () {
                 <li>
                     <p class="options-info">Custom config URL</p>
                     <input class="allowlist-url js-options-config-url" type="text" placeholder="Privacy Configuration URL" value="${this.model.debuggingSettings?.configURLOverride}" />
-                    <button class="allowlist-add float-right js-options-set-config-url" ${buttonDisabled ? 'disabled' : ''}>${buttonText}</button>
+                    <button class="custom-config-button float-right js-options-set-config-url" ${buttonDisabled ? 'disabled' : ''}>${buttonText}</button>
                 </li>
             </ul>
         </section>`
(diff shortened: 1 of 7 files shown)


shared/js/ui/views/internal-options.js
```diff
@@ -36,11 +45,58 @@ export default function InternalOptionsView (ops) {
...
 }
 
 InternalOptionsView.prototype = window.$.extend({}, ViewParent.prototype, {
-
+    setup: function () {
+        this._cacheElems('.js-options', ['config-url', 'set-config-url'])
+        this.bindEvents([
+            [this.$setconfigurl, 'click', this._clickSetting],
+            [this.$configurl, 'input', this._onURLChange]
+        ])
```
shakyShane, line 57: Should we align with other implementations, and support things like 'reload now' <- I found myself wanting to do that when testing :)

image

shared/js/ui/views/internal-options.js
```diff
@@ -25,6 +29,11 @@ function template () {
                         Internal-only settings for debugging the extension.
                     </p>
                 </li>
+                <li>
+                    <p class="options-info">Custom config URL</p>
+                    <input class="allowlist-url js-options-config-url" type="text" placeholder="Privacy Configuration URL" value="${this.model.debuggingSettings?.configURLOverride}" />
+                    <button class="allowlist-add float-right js-options-set-config-url" ${buttonDisabled ? 'disabled' : ''}>${buttonText}</button>
+                </li>
```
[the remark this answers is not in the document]
sammacbeth: Should be fixed by https://github.com/duckduckgo/duckduckgo-privacy-extension/pull/2485/commits/438937b0015190226629d2ce770f9fabdb99b0b5

shared/js/ui/views/internal-options.js
sammacbeth: The config should reload automatically when using the new debugger tool, but for consistency with other platforms I've added this by adding another state to the 'set'/'clear' button.

shakyShane: > The config should reload automatically when using the new debugger tool, but for consistency with other platforms I've added this by adding another state to the 'set'/'clear' button.


Ahh that's good then - I think it's a nice change because then this is useful outside of the new debugger interface too :)

shakyShane (approved):
@sammacbeth there's a failure in the integration tests, but I think it's unrelated :)

PR #2509, opened 2024-04-16
Fix the onboarding flow for the Chrome MV3 build

kzar:
The onboarding code has some logic to display a welcome banner the
first time the user performs a DuckDuckGo search after installing the
extension. Until now, that code did not work for Chrome MV3 builds,
since it attempted to inject a script into the "main world" by
appending a <script> element - that does not work with MV3 due to the
stricter Content-Security-Policy used. Let's use the
scripting.executeScript API instead to inject the required code into
the "main world".

Note: This code is somewhat convoluted and could likely be simplified
      in the future. But for now, with the upcoming MV3 migration
      deadline, let's make the minimum changes required to get this
      working!

**Reviewer:** @sballesteros, @sammacbeth 

## Steps to test this PR:
1. Install the extension
2. Ensure DDG is the default browser if you're running Linux
3. Perform a search from the URL bar
4. Ensure the purple onboarding banner is displayed and a CSP warning isn't shown in the console
5. Perform another search from the URL bar and ensure the onboarding banner is not displayed this time

Note: Test _both_ with MV2 and MV3 builds of the extension!

### DIFF
--- a/integration-test/onboarding.spec.js
+++ b/integration-test/onboarding.spec.js
@@ -2,7 +2,7 @@ import { test, expect } from './helpers/playwrightHarness'
 import backgroundWait from './helpers/backgroundWait'
 
 test.describe('onboarding', () => {
-    test('should manage the onboarding state and inject a script that calls window.onFirstSearchPostExtensionInstall on the first search post extension', async ({ context, backgroundPage, page }) => {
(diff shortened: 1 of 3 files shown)


sammacbeth (approved):
This worked for me in testing.

I think it would be helpful to file a follow-up to tidying up this code (post MV3 migration):
 - Make it a standalone 'component' (to move listeners of our events.js), and keep code in one place.
 - Use build flags to only include the component on chrome builds (i.e. `BUILD_TARGET !== 'firefox'`).

PR #2510, opened 2024-04-19
Remove GPC declarativeNetRequest rule when GPC feature is disabled

kzar:
When the user selects to disable Global Privacy Control (GPC) in the
options page, we must take care to remove the corresponding
declarativeNetRequest rule. Otherwise, the GPC headers will be added
to requests against the user's wishes.

**Reviewer:** @sammacbeth 
**CC:** @jdorweiler 

## Steps to test this PR:
1. Navigate to the GPC test page with the MV3 build https://privacy-test-pages.site/privacy-protections/gpc/
2. Run the tests and save the results
3. Turn off the GPC feature in the options page
4. Run the tests again and save the results
5. Remove the MV3 extension and install the MV2 extension
6. Repeat the tests with/without GPC enabled and compare the results to the ones saved earlier. Make sure they match exactly.

## Automated tests:
- [x] Unit tests
- [x] Integration tests


shared/js/background/dnr-gpc.js
```diff
@@ -0,0 +1,54 @@
...
+    }
+
+    await chrome.declarativeNetRequest.updateSessionRules({
+        removeRuleIds, addRules
+    })
+}
+
+settings.onSettingUpdate.addEventListener(
+    'GPC', () => { ensureGPCHeaderRule() }
+)
```
sammacbeth, line 54: Let's wrap this in a component - this will allow us to avoid using globals (`settings` and `tdsStorage` can be passed to the component constructor), and avoid calling `settings.onSettingsUpdate` on module import. This will make it easy for us to write a simple unittest for this component.

packages/ddg2dnr/lib/gpc.js
```diff
@@ -1,32 +0,0 @@
-/** @module gpc */
-const {
-    resourceTypes,
-    generateDNRRule
-} = require('./utils')
-
-const GPC_HEADER_PRIORITY = 40000
-
-function generateGPCheaderRules (config) {
```
sammacbeth, line 9: Couldn't this library still have a function for creating the DNR rule, but be agnostic to extension internals like config? i.e. `generateGPCheaderRule(exceptions)`. The extension-side code can then check settings and config then pass the relevant arguments to this function call?

sammacbeth (changes requested):

sammacbeth (approved):
As discussed, I'll approve this to unblock testing and lets address the comments in a follow-up.

PR #2513, opened 2024-04-24
Ensure onFirstSearchPostExtensionInstallCalls is called correctly

kzar:
**Reviewer:** @sammacbeth 
**CC:** @sballesteros 

When fixing the onboarding flow for the Chrome MV3 build[1], I made a
mistake. For MV3 builds, the page's onFirstSearchPostExtensionInstall
function was being called like:

    onFirstSearchPostExtensionInstall({
        isAddressBarQuery: false,
        showCounterMessaging: true,
        showWelcomeBanner: true
    }, {
        hadFocusOnStart: true
    })

Instead of like:

    onFirstSearchPostExtensionInstall({
        hadFocusOnStart: true,
        isAddressBarQuery: false,
        showCounterMessaging: true,
        showWelcomeBanner: true
    })

(I had missed a subtle Object.assign call.)

Let's fix that here, and also add test coverage for that call to catch
similar mistakes in the future.

1 - https://github.com/duckduckgo/duckduckgo-privacy-extension/commit/2e4441d633715b455e24fa9311af19b2aeb05230

## Automated tests:
- [ ] Unit tests
- [x] Integration tests

### DIFF
--- a/integration-test/onboarding.spec.js
+++ b/integration-test/onboarding.spec.js
@@ -1,6 +1,22 @@
 import { test, expect } from './helpers/playwrightHarness'
 import backgroundWait from './helpers/backgroundWait'
 
+function stubOnFirstSearchPostExtensionInstallOnInit (page) {
+    return page.addInitScript(() => {
+        window.afterInitCall = new Promise(resolve => {
+            Object.defineProperty(
+                window,
+                'onFirstSearchPostExtensionInstall',
+                {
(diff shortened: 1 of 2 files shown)


sammacbeth (approved):
Looks good. I ran the tests locally and it works. Lets see if the test runner issue is resolved soon so we can verify this runs fine on CI before merging.

PR #2518, opened 2024-04-26
[localize]: add Japanese

tachibana-shin:
**Reviewer:**

## Description:

Add Japanese

## Automated tests:
- [ ] Unit tests
- [ ] Integration tests

###### Reviewer Checklist:

###### PR Author Checklist:


sammacbeth:
Thank you for these contributions. Unfortunately, we don't take external localization contributions, as there is quite a bit more to it than just editing the files here:
- Adding this translation file requires us to also translate store listings pages and static assets on those pages.
- The extension also uses strings from our cross-platform dependencies (e.g. privacy-dashboard, autofill, content-scope-scripts), which also need to be translated.
- Adding a new language adds an ongoing maintenance cost, so we need to be sure in advance that the user numbers for that language justify this cost.

PR #2524, opened 2024-04-30
Update the build etc scripts and docs to reflect the Chrome MV3 migration

kzar:
We are migrating Chrome extension users over onto the manifest
v3 (MV3) build, in time for the June deadline[1]. Once that's
finished, only Firefox users will remain on the MV2 build. Let's
rename the build, release, and test scripts and related documentation
to reflect that change.

Note: We unfortunately can't remove the chrome-mv2 build target
      entirely, since Playwright does not yet support testing Firefox
      extensions[2]. To test the MV2 code via the integration tests,
      using the chrome-mv2 build is the best we can do for now.

1 - https://developer.chrome.com/blog/resuming-the-transition-to-mv3/
2 - https://github.com/microsoft/playwright/issues/7297

**Reviewer:** @sammacbeth


sammacbeth (approved):
Looking through, I think you got everything that needs changing here. Hopefully, nothing blows up when we run the next release 🤞

PR #2563, opened 2024-06-20
Refactor the devtools code, fixing Chrome MV3 support

kzar:
The devtools-panel.html page is for testing disabling/enabling
certain privacy protections, to help debug website
issues. Unfortunately the page did not work yet for Chrome MV3 builds,
let's fix that here. Since some refactoring is required anyway, let's
tidy things up a bit at the same time.

Some of the main changes:
 - Instead of messing with the extension's (MV2) state of the various
   privacy protections directly, we must now update the underlying
   configurations. That way the MV2 state for those features updates
   automatically and the MV3 decarativeNetRequest rules are also
   regenerated as necessary.
 - To access the ResourceLoader instances necessary for that, the
   devtools code needs to be split out into the Devtools component.
 - Instead of reloading pages right away after toggling protections,
   it should wait until after the changes have finished
   applying. Otherwise, the decarativeNetRequest rule changes for the
   MV3 build often won't have been applied in time. This results in
   requests being incorrectly blocked/allowed. The corresponding
   buttons must also be disabled while this is happening.
 - When the devtools panel page recreates the messaging connection, it
   should also recreate the message listeners. Otherwise the page will
   stop working after the background ServiceWorker restarts for Chrome
   MV3 builds.

**Reviewer:** @sammacbeth 
**CC:** @jdorweiler

## Steps to test this PR:
I spent quite a while testing the devtools-panel.html page on both Chrome MV3 and Firefox MV2 builds, everything _seems_ to be working correctly. Can you give it a test against your privacy-protections-debugger tool though? I want to make sure I didn't mess that integration up with these changes.

## Automated tests:
- [ ] Unit tests
- [ ] Integration tests

###### Reviewer Checklist:

###### PR Author Checklist:

### DIFF
--- a/shared/js/background/components/resource-loader.js
+++ b/shared/js/background/components/resource-loader.js
@@ -168,8 +168,9 @@ export default class ResourceLoader extends EventTarget {
      *  contents: any;
      *  etag?: string;
      * }} result
+     * @param {boolean} waitForUpdateProcessing
      */
-    async _updateData (result) {
+    async _updateData (result, waitForUpdateProcessing = false) {
         console.log(`Loaded ${this.name}:`, result)
         const updated = !!(this.data === null || this.etag)
         this.data = result.contents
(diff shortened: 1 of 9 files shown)


shared/js/background/components/resource-loader.js
```diff
@@ -234,7 +243,7 @@ export default class ResourceLoader extends EventTarget {
      */
     async modify (func) {
         await this.ready
-        const newValue = func(this.data)
+        const newValue = func(this.data) || this.data
```
kzar, line 246: (Just so the `.modify()` calls in the Devtools component don't have to pointlessly return the reference to this.data back again.)

sammacbeth (commented):
Looks good, thanks for this, and I like the refactors. Just a couple of points to discuss on the resource loader interface.

shared/js/background/components/resource-loader.js
```diff
@@ -168,8 +168,9 @@ export default class ResourceLoader extends EventTarget {
      *  contents: any;
      *  etag?: string;
      * }} result
+     * @param {boolean} waitForUpdateProcessing
      */
-    async _updateData (result) {
+    async _updateData (result, waitForUpdateProcessing = false) {
```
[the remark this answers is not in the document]
kzar: I did that initially but I found it caused a deadlock. I didn't run it through a debugger to see exactly why, for now it seemed simpler to avoid waiting by default.

shared/js/background/components/resource-loader.js
kzar: Fair enough, Done.

sammacbeth (approved):

PR #2588, opened 2024-07-19
Adjust ad click attribution pixel logic

kzar:
We recently added "pixel" requests to the mechanism we rely on to
support ad click attribution in DuckDuckGo products. These help
DuckDuckGo validate that the mechanism's logic is working correctly.

A few further adjustments are required:

1. The 'm_ad_click_active' pixel should only fire (and the
   'm_pageloads_with_ad_attribution' counter incremented) when a request
   is actually allowed by the mechanism.
2. The browser name (e.g. "chrome", "edge", etc.) should be appended
   to the pixel names.
3. The extension's version number (e.g. "2024.7.10") should be
   included with some of the pixels.

Further reading:
 - https://github.com/duckduckgo/duckduckgo-privacy-extension/commit/f8755949344728145dcd9da970b7b1bfc8e7d926
 - https://duckduckgo.com/duckduckgo-help-pages/privacy/web-tracking-protections/#duckduckgo-private-search-ads
 - https://improving.duckduckgo.com/

**Reviewer:** @sammacbeth, @jonathanKingston 

## Automated tests:
- [x] Unit tests
- [x] Integration tests

###### Reviewer Checklist:

###### PR Author Checklist:

### DIFF
--- a/shared/js/background/pixels.js
+++ b/shared/js/background/pixels.js
@@ -1,5 +1,6 @@
 /* global BUILD_TARGET */
 import load from './load'
+import { getBrowserName } from './utils'
 
 /**
  *
@@ -20,8 +21,11 @@ export function sendPixelRequest (pixelName, params = {}) {
         return
     }
 
+    const browserName = getBrowserName() || 'unknown'
+
     const randomNum = Math.ceil(Math.random() * 1e7)
     const searchParams = new URLSearchParams(Object.entries(params))
-    const url = getURL(pixelName) + `?${randomNum}&${searchParams.toString()}`
+    const url = getURL(`${pixelName}_extension_${browserName}`) +
+        `?${randomNum}&${searchParams.toString()}`
     return load.url(url)
 }
--- a/integration-test/click-attribution.spec.js
+++ b/integration-test/click-attribution.spec.js
@@ -20,10 +20,11 @@ for (let i = 0; i < testCases.length; i++) {
 }
 
 test.describe('Ad click blocking', () => {
+    let extensionVersion
     const backgroundPixels = []
     let clearBackgroundPixels
 
-    test.beforeEach(async ({ context, backgroundNetworkContext }) => {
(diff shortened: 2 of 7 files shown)


shared/js/background/pixels.js
```diff
@@ -20,8 +21,11 @@ export function sendPixelRequest (pixelName, params = {}) {
         return
     }
 
+    const browserName = getBrowserName() || 'unknown'
+
     const randomNum = Math.ceil(Math.random() * 1e7)
     const searchParams = new URLSearchParams(Object.entries(params))
-    const url = getURL(pixelName) + `?${randomNum}&${searchParams.toString()}`
+    const url = getURL(`${pixelName}_extension_${browserName}`) +
```
sammacbeth, line 28: This seems to be a different naming convention to on the browser site report pixel, which only sends browser name (does not send `extension`), but does not change `firefox` to `moz`. Can we align the behavior for all pixels? We might need to sync with @GioSensation if this will affect current autofill pixels.

sammacbeth (changes requested):

shared/js/background/pixels.js
kzar: Pixels are not sent for DDG extension users on Firefox, so I don't think `firefox` vs `moz` is an issue.

My preference is to stick with the "_extension_BROWSERNAME" suffix convention @GioSensation set up for the autofill pixels here, to avoid messing with those pixels. I also checked with @englehardt  and @lrjball that the suffix works for them.

If you're against that suffix, what are you asking for instead?

GioSensation: This does affect autofill pixels, but the number of panels to update is relatively low. The `extension_BROWSERNAME` convention makes it easier to distinguish from mobile, so that you can group all mobile and all extensions separately. It's not critical, but makes it slightly easier. I'm not entirely sure what's the goal, as @kzar was asking.

sammacbeth (approved):

jonathanKingston (approved):

PR #2609, opened 2024-08-07
Update typescript dependency to 5.5.4

kzar:
Some changes were required to get the type checking passing again:
 - Functions that return declarativeNetRequest rules with/without rule
   IDs need to be more clearly annotated, so that it's clear to
   typescript when a rule's ID is available.
 - Edge cases where getNextSessionRuleId() fails need to be handled
   properly, adding rules with an undefined/null ID won't work.
 - The DNRRule type needed to be replaced with the
   chrome.declarativeNetRequest.Rule rule type, and with that some
   string type rule properties needed to be cast to the correct enum
   types.

See https://devblogs.microsoft.com/typescript/announcing-typescript-5-4/#notable-behavioral-changes

## Automated tests:
- [ ] Unit tests
- [ ] Integration tests

###### Reviewer Checklist:

###### PR Author Checklist:

### DIFF
--- a/packages/ddg2dnr/lib/cookies.js
+++ b/packages/ddg2dnr/lib/cookies.js
@@ -1,4 +1,9 @@
-const { generateRequestDomainsByTrackerDomain, getTrackerEntryDomain, storeInLookup } = require('./utils')
+const {
+    castDNREnum,
+    generateRequestDomainsByTrackerDomain,
+    getTrackerEntryDomain,
+    storeInLookup
+} = require('./utils')
 
 const COOKIE_PRIORITY = 40000
 
@@ -10,7 +15,7 @@ const COOKIE_PRIORITY = 40000
  * @returns {import('./utils').RulesetResult} Cookie blocking rules
  */
 function generateCookieBlockingRuleset (tds, excludedCookieDomains, siteAllowlist, startingRuleId = 1) {
-    /** @type {import('./utils').DNRRule[]} */
+    /** @type {chrome.declarativeNetRequest.Rule[]} */
     const rules = []
     /** @type {Map<string, { domains: Set<string>, trackerDomains: Set<string> }>} */
     const entityDomainMapping = new Map()
@@ -66,9 +71,15 @@ function generateCookieBlockingRuleset (tds, excludedCookieDomains, siteAllowlis
             id: startingRuleId++,
             priority: COOKIE_PRIORITY,
             action: {
-                type: 'modifyHeaders',
-                requestHeaders: [{ header: 'cookie', operation: 'remove' }],
-                responseHeaders: [{ header: 'set-cookie', operation: 'remove' }]
+                type: castDNREnum('modifyHeaders'),
+                requestHeaders: [{
+                    header: 'cookie',
+                    operation: castDNREnum('remove')
+                }],
+                responseHeaders: [{
+                    header: 'set-cookie',
+                    operation: castDNREnum('remove')
+                }]
             },
             condition: {
                 requestDomains: Array.from(trackerDomains),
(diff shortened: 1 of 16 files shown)


types/declarative-net-request/index.d.ts
```diff
@@ -0,0 +1,20 @@
+type DNRRule = import('../../packages/ddg2dnr/lib/utils.js').DNRRule;
+type DNRRuleWithID = import('../../packages/ddg2dnr/lib/utils.js').DNRRuleWithID;
```
jonathanKingston, line 2: I'd rather we didn't define our own type globals within .d.ts files if we can help it. Especially as it means we can't easily move to a class to represent this data which IMO would be much cleaner/simpler.

Is `DNRRule` still used on it's own (in this file and elsewhere)?

packages/ddg2dnr/lib/utils.js
```diff
@@ -170,8 +169,14 @@ function storeInLookup (lookup, key, values) {
...
+ *
+ * @overload
+ * @param {Omit<generateDNRRuleDetails, 'id'>} ruleDetails
+ * @returns {DNRRule}
+ *
+ * @overload
  * @param {generateDNRRuleDetails} ruleDetails
- * @return {DNRRule}
+ * @returns {DNRRuleWithID}
  */
```
jonathanKingston, line 180: Is the `DNRRule` in this function the only remaining one left? (I've not checked).

types/declarative-net-request/index.d.ts
```diff
@@ -0,0 +1,20 @@
...
+
+// Augment the chrome.declarativeNetRequest API type definitions as provided by
+// @types/chrome to allow for DNRRule/DNRRuleWithID to be used. This is
+// necessary since working with the original Rule type that @types/chrome
+// provides is tricky from JS code, as that uses enums for some of the argument
+// properties. Enums unfortunately aren't ergonomic to use from JavaScript code,
+// assigning a correct string value for the enum isn't enough.
+declare namespace chrome.declarativeNetRequest {
```
jonathanKingston, line 10: Maybe @shakyShane knows but perhaps there's a way to coerce our type into theirs, seems like we aren't doing it in many places so if possible that seems preferred than overriding this which might get outdated etc.

jonathanKingston (commented):
Bunch of comments and questions. Maybe pass it to the type-overlord @shakyShane

types/declarative-net-request/index.d.ts
kzar: I couldn't see an easy way to do that. Lots of the Rule type properties are enums, whereas our DNRRuleWithID type goes with strings instead. I figured augmenting the interface for `updateDynamicRules` and `updateSessionRules` to allow our type as well was a pretty clean option for now, but open to suggestions if you can see a better way.

packages/ddg2dnr/lib/utils.js
kzar: No, some folks decided not to use `generateDNRRule()` so the rule types are scattered around unfortunately.

jonathanKingston: I'd consider filing that as a issue to clean up, and migrating `generateDNRRule` to be a class constructor; that way the type def can go away also.

types/declarative-net-request/index.d.ts
shakyShane: I added an idea over in https://github.com/duckduckgo/duckduckgo-privacy-extension/commit/c5f2cdf0d17d18cd2d8fc65bc5e4c8c32923c951

shakyShane (changes requested):
Please see my ideas over in https://github.com/duckduckgo/duckduckgo-privacy-extension/commit/c5f2cdf0d17d18cd2d8fc65bc5e4c8c32923c951 and adopt/change as you see fit :)

packages/ddg2dnr/lib/cookies.js
```diff
@@ -66,9 +71,15 @@ function generateCookieBlockingRuleset (tds, excludedCookieDomains, siteAllowlis
             id: startingRuleId++,
             priority: COOKIE_PRIORITY,
             action: {
-                type: 'modifyHeaders',
-                requestHeaders: [{ header: 'cookie', operation: 'remove' }],
-                responseHeaders: [{ header: 'set-cookie', operation: 'remove' }]
+                type: castDNREnum('modifyHeaders'),
```
shakyShane, line 74: nice!

shakyShane (approved):
nice! thanks for making those changes ❤️

jonathanKingston (approved):

PR #2726, opened 2024-09-19
Update privacy-dashboard dependency to 6.0.0

kzar:
With this update, the way privacy-dashboard communicates with the extension has
changed. Instead of sending messages using the chrome.runtime.sendMessage API,
the chrome.runtime.connect API is used instead. In the future, that will allow
us to listen for the disconnect event as a way to know when the popup UI has
been closed by the browser (e.g. the user clicked away).

**Reviewer:** @shakyShane 
**CC:** @sammacbeth 

## Steps to test this PR:
1. Check that the popup UI functions correctly, try things like allowlisting a page, or switching tabs, and make sure the popup closes itself. Ensure the website details populated in the popup UI looks right.

## Automated tests:
- [ ] Unit tests
- [ ] Integration tests

###### Reviewer Checklist:

###### PR Author Checklist:

### DIFF
--- a/shared/js/background/companies.js
+++ b/shared/js/background/companies.js
@@ -1,6 +1,7 @@
 const TopBlocked = require('./classes/top-blocked')
 const Company = require('./classes/company')
 const browserWrapper = require('./wrapper')
+const { postPopupMessage } = require('./popupMessaging')
 
 const Companies = (() => {
     let companyContainer = {}
@@ -87,8 +88,7 @@ const Companies = (() => {
             totalPagesWithTrackers = 0
             lastStatsResetDate = Date.now()
             Companies.syncToStorage()
-            const resetDate = Companies.getLastResetDate()
-            browserWrapper.notifyPopup({ didResetTrackersData: resetDate })
+            postPopupMessage({ messageType: 'didResetTrackersData' })
         },
 
         getLastResetDate: () => lastStatsResetDate,
--- a/shared/js/background/events.js
+++ b/shared/js/background/events.js
@@ -20,6 +20,7 @@ import { clearExpiredBrokenSiteReportTimes } from './broken-site-report'
 import {
     sendPageloadsWithAdAttributionPixelAndResetCount
 } from './classes/ad-click-attribution-policy'
+import { popupConnectionOpened, postPopupMessage } from './popupMessaging'
 const utils = require('./utils')
 const experiment = require('./experiments')
 const settings = require('./settings')
@@ -285,7 +286,7 @@ browser.webRequest.onHeadersReceived.addListener(
  */
 // message popup to close when the active tab changes.
 browser.tabs.onActivated.addListener(() => {
-    browserWrapper.notifyPopup({ closePopup: true })
+    postPopupMessage({ messageType: 'closePopup' })
 })
 
 // Include the performanceWarning flag in breakage reports.
@@ -350,6 +351,13 @@ browser.runtime.onMessage.addListener((req, sender) => {
 
     console.error('Unrecognized message to background:', req, sender)
     return false
+})
+
+// Handle popup UI (aka privacy dashboard) messaging.
+browser.runtime.onConnect.addListener(port => {
+    if (port.name === 'privacy-dashboard') {
+        popupConnectionOpened(port, messageHandlers)
+    }
 })
 
 /*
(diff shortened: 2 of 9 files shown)


shared/js/background/popupMessaging.js
```diff
@@ -0,0 +1,58 @@
...
+ * Note: If the ServiceWorker dies while there is an open connection, the popup
+ *       will take care of re-opening the connection.
+ *
+ * @param {Port} port
+ * @param {Record<
+ *     string,
+ *     (options: any, sender: any, message: any) => any
+ * >} messageHandlers
+ */
+export function popupConnectionOpened (port, messageHandlers) {
```
kzar, line 22: _(Message handlers passed in here instead of imported, to reduce circular dependencies.)_

shared/js/background/companies.js
```diff
@@ -87,8 +88,7 @@ const Companies = (() => {
             totalPagesWithTrackers = 0
             lastStatsResetDate = Date.now()
             Companies.syncToStorage()
-            const resetDate = Companies.getLastResetDate()
-            browserWrapper.notifyPopup({ didResetTrackersData: resetDate })
+            postPopupMessage({ messageType: 'didResetTrackersData' })
```
kzar, line 91: _Shane confirmed the `resetDate` isn't used at all, and in fact including it will throw a type error now._

shared/js/background/message-handlers.js
```diff
@@ -58,8 +59,8 @@ export async function setLists (options) {
     }
 
     try {
-        notifyPopup({ closePopup: true })
-        reloadCurrentTab()
+        postPopupMessage({ messageType: 'closePopup' })
+        await reloadCurrentTab()
```
sammacbeth, line 63: In the `reloadCurrentTab` implementation we're not awaiting the [`browser.tabs.reload`](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/tabs/reload) call. Do we want to wait here for the tab to complete reload, or just that the reload has started?

shared/js/background/popupMessaging.js
```diff
@@ -0,0 +1,58 @@
...
+    activePort = port
```
sammacbeth, line 23: Is it possible at all to open two popups concurrently?

I just tested in Chrome and if I inspect a popup in one window, then I can open a second window and open a second popup at the same time. I wonder what will happen here? Will we just post data to the most recently opened, or are they going to fight over reconnects?

sammacbeth (commented):
Looks mostly nice and clean. I just had a couple of comments:
 - Architecturally, if we can put the listener registration in a component. If we ever want to ship a headless version of the extension, we'll want to skip that registration.
 - There's potentially an edge-case when multiple popups are open. I think we should just check things don't blow up when that happens.

shared/js/background/popupMessaging.js
kzar: In that situation, I think incoming (popup to extension) messages will be received OK, but replies and outgoing messages will be sent to the second popup. The disconnect listeners also take care to check that they really are the active connection, before clearing the active connection variable.

Just tested and it behaved like I expected, somewhat broken in that disabling protections in the first popup results in the protections being disabled for the second popup - but nothing blows up that I can see.

_(This is an extreme edge-case to worry about, I doubt a user would trigger it + our popup UI often breaks in these kinds of situations already when the current tab can't be determined.)_

shared/js/background/events.js
```diff
@@ -352,6 +353,13 @@ browser.runtime.onMessage.addListener((req, sender) => {
     return false
 })
 
+// Handle popup UI (aka privacy dashboard) messaging.
+browser.runtime.onConnect.addListener(port => {
+    if (port.name === 'privacy-dashboard') {
+        popupConnectionOpened(port, messageHandlers)
+    }
+})
```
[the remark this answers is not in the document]
kzar: I considered using a "component" for this messaging, but I don't think that abstraction works too well for cases where lots of code outside of any component needs to send messages in to the component. I think this way is cleaner until more of the callsites for postPopupMessage are in components themselves. At that point refactoring will be pretty simple too, so it's not a big deal doing it this way for now IMO.

shared/js/background/message-handlers.js
kzar: Good catch, I'll fix that now :+1: .

sammacbeth (approved):

PR #2731, opened 2024-09-23
Add support for the "toggle reports" feature

kzar:
When the user clicks to disable protections for a website, the "toggle reports"
feature prompts them to also send a website breakage report to us. This helps
increase the number of user reports we receive for website breakage issues, in
turn helping us fix or mitigate them. The feature also takes care to not prompt
the user too often, at the time of writing only prompting one to three
times (depending on the user's response) every 48 hours.

**Reviewer:** @sammacbeth 
**CC:** @shakyShane 

## Steps to test this PR:
1. Try disabling protections for a website, the "simplified breakage report" (aka "toggle reports") UI should show, prompting you to send a breakage report.
2. The UI should only show after < three accepted responses, and not after any declined responses.
3. A breakage report request should be made in the extension's background when accepted.
4. Run `dbg.resetToggleReports()` in the extension's background console to reset the count.

## Automated tests:
- [x] Unit tests
- [ ] Integration tests

###### Reviewer Checklist:

###### PR Author Checklist:

### DIFF
--- a/shared/js/background/broken-site-report.js
+++ b/shared/js/background/broken-site-report.js
@@ -1,15 +1,21 @@
 /**
+ * When web tracking protections interfere with essential usability, for
+ * example, if users can't sign in to a popular website, they can report that
+ * breakage to us anonymously in our apps and extensions. This is part of our
+ * tooling for those anonymous broken site reports.
  *
- * This is part of our tool for anonymous broken site reports
- * Learn more at https://duck.co/help/privacy/atb
- *
+ * Learn more at: https://duckduckgo.com/duckduckgo-help-pages/privacy/web-tracking-protections/#remotely-configured-exceptions
  */
+
+const browser = require('webextension-polyfill')
 const load = require('./load')
 const browserWrapper = require('./wrapper')
 const settings = require('./settings')
 const parseUserAgentString = require('../shared-utils/parse-user-agent-string')
 const { getURLWithoutQueryString } = require('./utils')
 const { getURL } = require('./pixels')
+const tdsStorage = require('./storage/tds').default
+const tabManager = require('./tab-manager')
 const maxPixelLength = 7000
 
 /**
@@ -19,23 +25,22 @@ const maxPixelLength = 7000
  * @param {string} querystring
  *
  */
-export function fire (querystring) {
-    let url = constructUrl(querystring, false)
+export function fire (pixelName, querystring) {
+    let url = constructUrl(pixelName, querystring, false)
 
     // If we're over the max pixel length, truncate the less important params
     if (url.length > maxPixelLength) {
-        url = constructUrl(querystring, true)
+        url = constructUrl(pixelName, querystring, true)
     }
 
     // Send the request
     load.url(url)
 }
 
-function constructUrl (querystring, truncate) {
+function constructUrl (pixelName, querystring, truncate) {
     const randomNum = Math.ceil(Math.random() * 1e7)
-    const pixelName = 'epbf'
     const browserInfo = parseUserAgentString()
-    const browser = browserInfo?.browser
+    const browserName = browserInfo?.browser
     const extensionVersion = browserWrapper.getExtensionVersion()
     const atb = settings.getSetting('atb')
 
@@ -55,8 +60,8 @@ function constructUrl (querystring, truncate) {
     }
     // build url string
     let url = getURL(pixelName)
-    if (browser) {
-        url += `_${browser.toLowerCase()}`
+    if (browserName) {
+        url += `_${browserName.toLowerCase()}`
     }
     // random number cache buster
     url += `?${randomNum}&`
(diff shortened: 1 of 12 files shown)


shared/js/background/broken-site-report.js
```diff
@@ -241,6 +252,50 @@ export async function breakageReportForTab ({
...
+ *
+ * @param {Object} arg
+ * @prop {string} pixelName
+ * @prop {import("./classes/tab") | undefined} arg.currentTab
+ * @prop {string | undefined} arg.category
+ * @prop {string | undefined} arg.description
+ * @prop {string | undefined} arg.reportFlow
+ *   String detailing the UI flow that this breakage report came from.
+ */
+export async function sendBreakageReportForCurrentTab (
```
sammacbeth, line 271: This function is only used in a component and a message handler - this seems a good candidate to be in a component itself so `settings` and `tdsStorage` can be injected. We can use the `registerMessageHandler` helper from the component to replace the statically defined `submitBrokenSiteReport` handler

shared/js/background/components/toggle-reports.js
```diff
@@ -0,0 +1,265 @@
...
+    /**
+     * Check if the toggle report UI flow should been displayed display for the
+     * currently focused tab.
+     *
+     * @param {import('../classes/tab')?} currentTab
+     * @returns {Promise<boolean>}
+     */
+    static async shouldDisplay (currentTab = null) {
+        currentTab = currentTab || (
+            await tabManager.getOrRestoreCurrentTab()
```
sammacbeth, line 234: Why are we not injecting tabManager when creating the component and using `this.tabManager`?

Also, as fas as I can tell, the `currentTab` argument is never provided, so lets remove it.

shared/js/background/components/toggle-reports.js
```diff
@@ -0,0 +1,265 @@
...
+        times.push({ timestamp: Date.now(), accepted })
+        settings.updateSetting('toggleReportTimes', times)
+
+        if (accepted) {
+            await sendBreakageReportForCurrentTab({
+                pixelName: 'protection-toggled-off-breakage-report',
+                reportFlow: 'on_protections_off_dashboard_main'
+            })
+
+            await resolveAfterDelay(2000)
```
sammacbeth, line 159: Is this delay part of the spec for this flow, or a number we've picked to make the flow feel nicer?

unit-test/background/toggle-reports.js
```diff
@@ -0,0 +1,329 @@
...
+        settingsStorage = new Map()
+        spyOn(settings, 'ready').and.returnValue(Promise.resolve())
+        spyOn(settings, 'getSetting').and.callFake(
+            name => settingsStorage.get(name)
+        )
+        spyOn(settings, 'updateSetting').and.callFake(
+            (name, value) => {
+                settingsStorage.set(name, value)
+            }
+        )
```
sammacbeth, line 43: I think properly injecting `settings`, `tdsStorage` and `tabManager` into the `ToggleReports`/`BreakageReports` components would eliminate the need for all these spies.

sammacbeth (commented):
Looks good, but I think there would be some benefits from adopting a bit more of the component model. Happy to sync if it's not clear what I'm asking for.

shared/js/background/components/toggle-reports.js
kzar: Because this is a static method, which is in turn because it needs to be called from the "setLists" message handler.

shared/js/background/components/toggle-reports.js
kzar: Without the delay, the "Thanks for the report" message won't be visible. It's not spec'd as far as I know, but we are discussing how long the delay should be exactly in the ship review.

shared/js/background/broken-site-report.js
kzar: I had a go at moving the broken-site-report module into its own component today, but it's a fairly big change that I won't have time to tidy up and test properly for now. I don't think the changes here make refactoring that in the future any harder either, so I propose I file a follow-up task for this in Asana.

shared/js/background/broken-site-report.js
```diff
@@ -1,15 +1,21 @@
...
+ * breakage to us anonymously in our apps and extensions. This is part of our
+ * tooling for those anonymous broken site reports.
  *
- * This is part of our tool for anonymous broken site reports
- * Learn more at https://duck.co/help/privacy/atb
- *
+ * Learn more at: https://duckduckgo.com/duckduckgo-help-pages/privacy/web-tracking-protections/#remotely-configured-exceptions
  */
+
+const browser = require('webextension-polyfill')
```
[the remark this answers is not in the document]
kzar: I suggest we do that at the same time as moving the module into its own component.

unit-test/background/toggle-reports.js
kzar: Even if broken-site-report is also a componenent, `shouldDisplay` still needs to be static (see above) and therefore so does `clearExpiredResponses` and `countResponses`. So we would still need to import the `tabManager` and `settings` in the toggle-reports component. Therefore, we would still need these spies.

sammacbeth (approved):

PR #2743, opened 2024-10-10
Bump CI to node 20

sammacbeth:
## Description:
Some of our dependencies are [dropping support for Node 18](https://github.com/duckduckgo/duckduckgo-privacy-extension/actions/runs/11214035345/job/31168291389?pr=2742). Node 20 is the current LTS, let's use that.

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -12,7 +12,7 @@ jobs:
       - uses: actions/checkout@v4
       - uses: actions/setup-node@v4
         with:
-          node-version: 18
+          node-version: 20
           cache: 'npm'
       - run: npm run install-ci
       - run: npm run lint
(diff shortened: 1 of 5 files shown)


kzar (changes requested):
I guess we should update minimum node and npm versions in package.json too?

package.json
```diff
@@ -1,7 +1,7 @@
 {
     "name": "duckduckgo-privacy-extension",
     "engines": {
-        "node": ">=18.0.0",
+        "node": ">=20.0.0",
         "npm": ">=8.0.0"
```
kzar, line 5:
```suggestion
        "npm": ">=10.0.0"
```

kzar (approved):
Left a suggestion for npm version (based on https://nodejs.org) but otherwise LGTM.

PR #2764, opened 2024-10-17
Move DNR module side-effects into a component.

sammacbeth:
**Reviewer:** @kzar 

## Description:
A couple of the DNR modules were registering listeners on import. This PR moves that sites into a component, so we have control over when those listeners get registered, and to what objects.

### DIFF
--- a/shared/js/background/components/dnr.js
+++ b/shared/js/background/components/dnr.js
@@ -1,0 +1,56 @@
+import browser from 'webextension-polyfill'
+import ATB from '../atb'
+import { flushSessionRules } from '../dnr-session-rule-id'
+import { clearInvalidDynamicRules } from '../dnr-utils'
+import { refreshUserAllowlistRules } from '../dnr-user-allowlist'
+import { ensureGPCHeaderRule } from '../dnr-gpc'
+import { onConfigUpdate } from '../dnr-config-rulesets'
+
+/**
+ * @typedef {import('./tds.js').default} TDS
+ * @typedef {import('../settings.js')} Settings
+ * @typedef {import('./trackers.js').default} Trackers
+ */
+
+export default class DNR {
+    /**
+     * @param {{
+    *  settings: Settings;
+    *  tds: TDS;
+    * }} options
+    */
+    constructor ({ settings, tds }) {
+        this.featureName = 'DNR'
+        this.settings = settings
+        this.tds = tds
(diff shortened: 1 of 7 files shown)


shared/js/background/components/dnr.js
```diff
@@ -0,0 +1,59 @@
...
+import { refreshUserAllowlistRules } from '../dnr-user-allowlist'
+import { ensureGPCHeaderRule } from '../dnr-gpc'
+import { onConfigUpdate } from '../dnr-config-rulesets'
+
+/**
+ * @typedef {import('./tds.js').default} TDS
+ * @typedef {import('../settings.js')} Settings
+ * @typedef {import('./trackers.js').default} Trackers
+ */
+
```
kzar, line 14: I think with the name being "DNR" it's not super clear what the component is for, I could imagine trying to explain to someone in a code review why putting the DNR logic for a new privacy feature in here is a bad idea, and them arguing "well it's called DNR, so why not?".

Not sure what you think, and I'm not sure what a better name would be. Perhaps a good option could be to add a nice comment to explain what the DNR component is for? I figure if we explain to folks it's just for instrumenting things like updating the rulesets on install/update, rather than to contain the feature-specific logic, it might be enough. WDYT?

shared/js/background/dnr-user-allowlist.js
```diff
@@ -103,10 +103,9 @@ export async function refreshUserAllowlistRules (allowlistedDomains) {
 
 /**
  * Retrieve a normalized and sorted list of user denylisted domains.
- * @returns {Promise<string[]>}
+ * @returns {string[]}
  */
-export async function getDenylistedDomains () {
-    await settings.ready()
```
kzar, line 109: Why stop waiting for settings to be ready, when we're using settings?

sammacbeth: This function was never called in a context where settings could not be ready - most of the DNR functions don't start running until config or the blocklist is loaded, and those require settings to be ready. Here, removing the `settings.ready` call, allows this function to be sync.

shared/js/background/components/dnr.js
sammacbeth: Good point. I renamed the class and added a comment to describe what it does.

kzar (approved):

PR #2839, opened 2024-12-06
Improve lint workflow

muodov:
**Reviewer:**

## Description:

https://app.asana.com/0/892838074342800/1208921863159441/f

## Automated tests:
- [ ] Unit tests
- [ ] Integration tests

###### Reviewer Checklist:

###### PR Author Checklist:


.github/workflows/checks.yml
```diff
@@ -15,7 +15,7 @@ jobs:
                   node-version: 20
                   cache: 'npm'
             - run: npm run install-ci
-            - run: npm run lint
```
kzar, line 18: I would prefer not to have it automatically run `lint-fix`, but I think having the guidance here is nice. My vote would be to tweak this to just explain to the user to try running `npm run lint-fix`. WDYT @sammacbeth ?

kzar: _(Thinking about it, if we _don't_ go with my suggestion above perhaps we should even just combine the `lint` and `lint-fix` commands. Otherwise, we'd have two commands that do basically the same thing.)_

muodov: @kzar note that lint-fix is executed only to generate and print the diff in CI logs. It doesn't actually commit anything and the CI check will still fail

sammacbeth: I think this is fine as-is: this will only run lint-fix if there is a failure to show a better error message in the logs.

sammacbeth (approved):

muodov:
@sammacbeth fyi the test failures seem unrelated, but they block merging

kzar:
> @sammacbeth fyi the test failures seem unrelated, but they block merging

I think it was the unresolved conversations that blocked merging, but no matter it's merged now. Thanks again!

PR #2892, opened 2025-01-10
Schema for ad-click attribution pixels

sammacbeth:
Adds schema for pixels sent for the ad-click attribution functionality.


jonathanKingston (approved):
I'm not super familiar with these pixels as they came in after I wrote the code. However everything looks as expected, wait for Steve / Dave if you want more assurance.

englehardt (approved):
Made one suggestion, but otherwise LGTM

PR #2905, opened 2025-01-17
A/B/N Experiment metrics

sammacbeth:
Implements metrics for A/B/N experiments, as per https://app.asana.com/0/1208889145294658/1208747415972722

Metrics implemented:
- `app_use` (triggered when the background service worker starts up).
- `search` (on SERP landing).
- `brokenSiteReport` (when the user has sent a broken site report).

Note, metrics are not active in Firefox.

### DIFF
--- a/shared/js/background/components/abn-experiments.js
+++ b/shared/js/background/components/abn-experiments.js
@@ -17,7 +17,7 @@ import { sendPixelRequest } from '../pixels';
 /**
  * @returns {ExperimentMetric[]}
  */
-export function generateBuildRetentionMetrics() {
+export function generateRetentionMetrics() {
     return ['app_use', 'search'].flatMap((metric) => {
         const metrics = [0, 1, 2, 3, 4, 5, 6, 7].map((conversionWindowStart) => ({
             metric,
@@ -133,7 +133,7 @@ export default class AbnExperimentMetrics {
         const cohort = this.remoteConfig.getCohort(featureName, subFeatureName);
         if (cohort && !cohort.enrolledAt) {
             cohort.enrolledAt = Date.now();
-            cohort.metrics = (metrics || generateBuildRetentionMetrics()).map((m) => ({ ...m, counter: 0, sent: false }));
+            cohort.metrics = (metrics || generateRetentionMetrics()).map((m) => ({ ...m, counter: 0, sent: false }));
             sendPixelRequest(`experiment_enroll_${subFeatureName}_${cohort.name}`, {
                 enrollmentDate: new Date(cohort.enrolledAt).toISOString().slice(0, 10),
             });
--- a/shared/js/background/components/abn-experiments.js
+++ b/shared/js/background/components/abn-experiments.js
@@ -71,7 +71,7 @@ export class SearchMetric {
         browser.webRequest.onCompleted.addListener(
             async (details) => {
                 const params = new URL(details.url).searchParams;
-                if (params.has('q') && (params.get('q')?.length || 0) > 0) {
+                if (params.get('q')?.length) {
                     await abnMetrics.remoteConfig.ready;
                     abnMetrics.onMetricTriggered('search');
                 }
--- a/shared/js/background/components/abn-experiments.js
+++ b/shared/js/background/components/abn-experiments.js
@@ -143,9 +143,16 @@ export default class AbnExperimentMetrics {
     }
 
     /**
-     *
+     * Triggered when a given metric is triggered.
+     * 
+     * Checks all active, enrolled subfeature experiments to find matching metrics that should be 
+     * sent for them. With those that match, their counter is incremented by `value`, and if they
+     * exceed the threshold, a pixel is sent.
+     * 
+     * After incrementing counters and potentially sending pixels, the updated cohort state is 
+     * written back to settings.
      * @param {string} metric
-     * @param {number} value
+     * @param {number} [value]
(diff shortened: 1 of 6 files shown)


shared/js/background/components/abn-experiments.js
```diff
@@ -0,0 +1,221 @@
...
+
+export class AppUseMetric {
+    /**
+     * Metric that fires once on creation.
+     * @param {{
+     * abnMetrics: AbnExperimentMetrics
+     * }} opts
+     */
+    constructor({ abnMetrics }) {
+        // trigger on construction: happens whenever the service worker is spun up, which should correlate with browser activity.
```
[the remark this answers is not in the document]
sammacbeth: Yes, the SW triggering is a bit unpredictable, but I think it's OK for this use-case, because:
1. These metrics are only used by the A/B/N framework, which rate limits the actual pixels that trigger per conversion internal. This means that overtriggering is not an issue: triggers after the first in a day will be ignored.
2. The `app_use` metric on mobile is tied to app launches, but on desktop we're mostly concerned about retention, i.e. that they didn't uninstall/disable the extension. This implementation will just periodically trigger when the browser is open, which seems a reasonable analogue to 'use'. Other ways of measuring would likely have similar false positives and negatives.

unit-test/background/abn-framework.js
```diff
@@ -125,3 +160,191 @@ describe('choseCohort', () => {
...
+        );
+    });
+
+    it('onMetricTriggered can trigger multiple matching metrics', () => {
+        const { remoteConfig, abnMetrics } = constructMockComponents();
+        remoteConfig.updateConfig(mockExperimentConfig);
+        abnMetrics.markExperimentEnrolled(feature, subFeature);
+        // modify enrollment to be 6 days ago
+        const cohort = remoteConfig.getCohort(feature, subFeature);
+        cohort.enrolledAt = Date.now() - 1000 * 60 * 60 * 25 * 6;
```
kzar, line 255: Is `Date.now() - 1000 * 60 * 60 * 25 * 6;` a typo? (I figure it should be `* 25`, as in 24 hours in a day.) Same elsewhere, also maybe have a `millisecondsInDay` or similar constant in this file to reuse?

shared/js/background/components/abn-experiments.js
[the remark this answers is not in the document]
kzar: OK fair enough, that makes sense. Thanks for explaining :+1: .

shared/js/background/components/abn-experiments.js
```diff
@@ -0,0 +1,221 @@
...
+     * @param {string} featureName
+     * @param {string} subFeatureName
+     * @param {ExperimentMetric[]} [metrics]
+     */
+    markExperimentEnrolled(featureName, subFeatureName, metrics) {
+        const cohort = this.remoteConfig.getCohort(featureName, subFeatureName);
+        if (cohort && !cohort.enrolledAt) {
+            cohort.enrolledAt = Date.now();
+            cohort.metrics = (metrics || generateRetentionMetrics()).map((m) => ({ ...m, counter: 0, sent: false }));
+            sendPixelRequest(`experiment_enroll_${subFeatureName}_${cohort.name}`, {
```
kzar, line 157: Are we sure that `subFeatureName` and `cohort.name` will never have special characters that could screw up the pixel request's URL? If not (or just to be safe), we could encode the pixel name in `sendPixelRequest()`?

shared/js/background/components/abn-experiments.js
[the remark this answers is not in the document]
kzar: Could you add a note section to the comment here to explain though? It seems like folks could be confused by it, or think it's a bug in the future.

unit-test/background/abn-framework.js
sammacbeth: No - I used 25 hours per day so that the events land safely within a time window. Using 24 hours would make it potentially unclear which bucket the metric should fall into (i.e. event happens _exactly_ 24 hours after enrollment, is that day 0 or day 1?). I've added a brief comment to say why 25 hour is used.

shared/js/background/components/abn-experiments.js
sammacbeth: `subFeatureName` must conform to `/^[a-zA-Z0-9]+$/`, which is enforced via unit-tests in the privacy config: https://github.com/duckduckgo/privacy-configuration/blob/main/tests/config-tests.js#L68-L75

`cohort.name` doesn't have the same validation, so could be problematic. I'll check with native teams to see if they are doing some sanitization on the client, or we should just add config validation to protect it.

kzar: Thanks, glad you're checking that with them. Perhaps it's also worth adding a `encodeURIComponent(pixelName)` call in the shared `sendPixel()` function to help protect other folks writing pixels in the future too?

sammacbeth: Added config validation in https://github.com/duckduckgo/privacy-configuration/pull/2688.

I don't think adding `encodeURIComponent` in sendPixel will be of much use:
- It doesn't protect against `_`, which gets converted to a namespace separator on the backend.
- We generally shouldn't be sending variables we don't control in the pixel name, so this would only be catching an suppressing coding mistakes, which could end up hiding bugs. I'd rather fail fast here.

kzar (approved):

PR #2918, opened 2025-01-24
Additional metrics for breakage testing

sammacbeth:
(stacked on top of #2906)

Adds additional metrics that can be used in A/B/N tests as proxies for breakage:
- Privacy dashboard open.
- Protection toggled on/off.
- 2x and 3x refresh within a time window.

To do this, I've moved messaging to a component, allowing others to subscribe to the incoming messages, which we use to trigger appropriate metrics.


kzar (changes requested):

shared/js/background/metrics.js
```diff
@@ -0,0 +1,188 @@
...
+                refreshCounter.push(Date.now());
+
+                // reload-twice-within-12-seconds
+                const tMinus12s = Date.now() - 12000;
+                if (refreshCounter.filter((t) => t > tMinus12s).length === 2) {
+                    abnMetrics.onMetricTriggered('2xRefresh');
+                }
+                // reload-three-times-within-20-seconds
+                const tMinus20s = Date.now() - 20000;
+                if (refreshCounter.filter((t) => tMinus20s).length === 3) {
```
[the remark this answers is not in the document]
sammacbeth: Good catch - I added tests and found another bug too.

unit-test/background/metrics.js
```diff
@@ -0,0 +1,84 @@
...
+                detail: {
+                    tabId,
+                    timeStamp,
+                },
+            }),
+        );
+    }
+}
+
+fdescribe('RefreshMetric', () => {
```
kzar, line 20: I've not seen `fdescribe` before, how's that different from `describe`?

kzar (approved):

unit-test/background/metrics.js
sammacbeth: Oops - that tells the test running to only run that batch of tests. Removed.

PR #2939, opened 2025-02-05
Refactor broken site report message handlers

sammacbeth:
**Reviewer:**

## Description:
Some breakage reports were coming through with config version `undefined` on Firefox (https://app.asana.com/0/0/1209244730607066/f). This is likely caused by a race condition in the breakage report flow if the background is not fully loaded yet.

To fix this, this PR:
- Moves the sendBreakageReport function to a component.
- Removes static access to page context information, using components instead.
- Allows us to wait for the correct config lifecycle event before getting config version.
- Updates the toggle reports component to use this new entrypoint for sending reports.

## Automated tests:
- [ ] Unit tests
- [ ] Integration tests

###### Reviewer Checklist:

###### PR Author Checklist:

### DIFF
--- a/shared/js/background/broken-site-report.js
+++ b/shared/js/background/broken-site-report.js
@@ -321,6 +321,8 @@ export async function breakageReportForTab({
 export async function sendBreakageReportForCurrentTab({ pixelName, currentTab, category, description, reportFlow }) {
     await settings.ready();
     await tdsStorage.ready('config');
+    // wait for onload callbacks (to ensure that config has been correctly processed)
+    await tdsStorage.config.allLoadingFinished
 
     const tab = currentTab || (await tabManager.getOrRestoreCurrentTab());
     if (!tab) {
--- a/shared/js/background/broken-site-report.js
+++ b/shared/js/background/broken-site-report.js
@@ -10,15 +10,12 @@
  * @typedef {import('@duckduckgo/privacy-dashboard/schema/__generated__/schema.types').DataItemId} DisclosureParamId
  */
 
-const browser = require('webextension-polyfill');
 const load = require('./load');
 const browserWrapper = require('./wrapper');
 const settings = require('./settings');
 const parseUserAgentString = require('../shared-utils/parse-user-agent-string');
 const { getCurrentTab, getURLWithoutQueryString } = require('./utils');
 const { getURL } = require('./pixels');
-const tdsStorage = require('./storage/tds').default;
-const tabManager = require('./tab-manager');
 const maxPixelLength = 7000;
 
 /**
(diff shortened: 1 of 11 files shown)


shared/js/background/broken-site-report.js
```diff
@@ -321,6 +321,8 @@ export async function breakageReportForTab({
 export async function sendBreakageReportForCurrentTab({ pixelName, currentTab, category, description, reportFlow }) {
     await settings.ready();
     await tdsStorage.ready('config');
```
kzar, line 323: Any reason to wait for both `tdsStorage.ready('config')` and `tdsStorage.config.allLoadingFinished`?

shared/js/background/components/dashboard-messaging.js
```diff
@@ -0,0 +1,92 @@
...
+import { breakageReportForTab, getDisclosureDetails } from '../broken-site-report';
+import { dashboardDataFromTab } from '../classes/privacy-dashboard-data';
+import { registerMessageHandler } from '../message-handlers';
+import { getCurrentTab } from '../utils';
+import { isFireButtonEnabled } from './fire-button';
+
+/**
+ * Message handlers for communication from the dashboard to the extension background.
+ *
+ * Note, additional message handles for toggle reports is in the separate ToggleReports component.
```
kzar, line 11:
```suggestion
 * Note, additional message handlers for toggle reports is in the separate ToggleReports component.
```

shared/js/background/components/dashboard-messaging.js
```diff
@@ -0,0 +1,92 @@
...
+    /**
+     * Only the dashboard sends this message, so we import the types from there.
+     * @param {import('@duckduckgo/privacy-dashboard/schema/__generated__/schema.types').BreakageReportRequest} breakageReport
+     * @param {string} [pixelName]
+     * @param {string} [reportFlow]
+     * @returns {Promise<void>}
+     */
+    async submitBrokenSiteReport(breakageReport, pixelName = 'epbf', reportFlow = undefined) {
+        // wait for config and TDS so we can get etags and config version
+        await Promise.all([this.tds.remoteConfig.allLoadingFinished, this.tds.tds.ready]);
```
kzar, line 40: Is it just this `await Promise.all([this.tds.remoteConfig.allLoadingFinished, this.tds.tds.ready]);` that's resolving the race condition, or are there other changes here that do too? I'm thinking we should split this into two PRs, one for the race condition and one for the refactor.

shared/js/background/components/dashboard-messaging.js
kzar, line 11: I think there are other dashboard messages that we didn't move here? E.g. for the fire button. Maybe we should mention some of those in the comment here, or move the handlers into this component too? I think the docs here might help https://duckduckgo.github.io/privacy-dashboard/modules/Browser_Extensions_integration.html

kzar (changes requested):
Overall I like the changes, but can we split the race condition fix out from the refactor? Also, I'd like to avoid weakening the ToggleReports unit tests.

unit-test/background/toggle-reports.js
```diff
@@ -148,40 +144,7 @@ describe('ToggleReports', () => {
...
-                        noActionRequests: '',
-                        adAttributionRequests: '',
-                        ignoredByUserRequests: '',
-                    },
-                },
-            ],
-            1,
-            2,
-        );
+        await expectReports([[{}, 'protection-toggled-off-breakage-report', 'on_protections_off_dashboard_main']], 1, 2);
```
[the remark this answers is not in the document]
sammacbeth: I see your point, but what would that actually test? It would actually just be a test of `dashboardMessaging.submitBrokenSiteReport`, that it passes the arguments given here correctly to the pixel request. That would be better encapsulated in a separate unit-test.

Part of the advantage of this refactor, is that we can mock the boundaries of `ToggleReport` better - rather than having to intercept the outgoing pixel to test that a report was sent, we can just mock the interface for sending reports.

shared/js/background/components/dashboard-messaging.js
sammacbeth: I've updated that docstring to include all messages and where they are being handled. Thanks for pointing to those docs, super helpful!

unit-test/background/toggle-reports.js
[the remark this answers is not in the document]
sammacbeth: Note, I've added some new tests for `dashboardMessaging.submitBrokenSiteReport` explicitly, so I think everything should be now covered. The `toggle-reports` tests test everything up to the `submitBrokenSiteReport` call, then the `dashboard-messaging` tests check that call to pixel sent.

shared/js/background/components/dashboard-messaging.js
kzar: Ace, looks good. Actually let's include that link in the docstring too for folks?

shared/js/background/components/dashboard-messaging.js
```diff
@@ -0,0 +1,116 @@
...
+ *  - getToggleReportOptions
+ *  - rejectToggleReport
+ *  - sendToggleReport
+ *  - seeWhatIsSent
+ *
+ * Static message handlers:
+ *  - openOptions
+ *  - search
+ *  - setLists
+ */
```
kzar, line 36:
```suggestion
 * See https://duckduckgo.github.io/privacy-dashboard/modules/Browser_Extensions_integration.html
 *
 */
```

shared/js/background/message-handlers.js
```diff
@@ -323,9 +279,6 @@ export function addDebugFlag(message, sender, req) {
  * @param {(options: any, sender: any, req: any) => any} func
  */
 export function registerMessageHandler(name, func) {
-    if (messageHandlers[name]) {
```
kzar, line 326: How come we're removing this check?

sammacbeth: This was being problematic in tests, where creating a component in subsequent test cases threw an exception and failed. If we're worried about naming collisions in message handlers, we could just log here.

PR #2950, opened 2025-02-12
Send enrollment dates in ET, and align conversion windows with ET days.

sammacbeth:
**Reviewer:**

## Description:
Our A/B/N framework assumes that:
 1. Enrollment dates are sent in ET (UTC-5).
 2. Conversation windows are calculated over calendar days (in ET), rather than rolling 24-hour windows from enrollment.

Our initial implementation used UTC dates (default for javascript `Date` objects), and using rolling windows, so this PR corrects that by:
1. Converting the stored enrollment timestamps to a date in ET when sending.
2. Setting the enrollment date to midnight in ET when joining an experiment. This means that conversion windows will get day-aligned without us having to change the rolling-window implementation.

This also contains a fix for a potential race with the `app_use` metric. If it is triggered too early, we may not yet have the config processed and ready to handle metrics.

## Automated tests:
- [ ] Unit tests
- [ ] Integration tests

###### Reviewer Checklist:

###### PR Author Checklist:

### DIFF
--- a/shared/js/background/components/abn-experiments.js
+++ b/shared/js/background/components/abn-experiments.js
@@ -89,10 +89,10 @@ export default class AbnExperimentMetrics {
         const cohort = this.remoteConfig.getCohort(featureName, subFeatureName);
         if (cohort && !cohort.enrolledAt) {
             // We set the enrollment timestamp as the start of the current day in ET, so that all conversion windows align with ET date changes.
-            cohort.enrolledAt = startOfDayET(Date.now());
+            cohort.enrolledAt = startOfDayEST(Date.now());
             cohort.metrics = (metrics || generateRetentionMetrics()).map((m) => ({ ...m, counter: 0, sent: false }));
             sendPixelRequest(`experiment_enroll_${subFeatureName}_${cohort.name}`, {
-                enrollmentDate: getDateStringET(cohort.enrolledAt),
+                enrollmentDate: getDateStringEST(cohort.enrolledAt),
             });
             // updated stored cohort metadata
             this.remoteConfig.setCohort(featureName, subFeatureName, cohort);
@@ -121,7 +121,7 @@ export default class AbnExperimentMetrics {
                 if (!cohort?.metrics || !cohort?.enrolledAt) {
                     return;
                 }
-                const enrollmentDate = getDateStringET(cohort.enrolledAt);
+                const enrollmentDate = getDateStringEST(cohort.enrolledAt);
                 const daysSinceEnrollment = Math.floor(((timestamp || Date.now()) - cohort.enrolledAt) / (1000 * 60 * 60 * 24));
                 // Find metrics for this experiment that match at this point in time.
                 // i.e. we are within the conversion window, and haven't sent the pixel yet.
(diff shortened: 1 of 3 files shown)


shared/js/background/components/abn-experiments.js
```diff
@@ -155,3 +156,30 @@ export default class AbnExperimentMetrics {
...
+    return new Date(timestamp).toLocaleDateString('sv', { timeZone: '-05:00' });
+}
+
+/**
+ * Calculates the unix timestamp for the start of the current day in Eastern Time (UTC-5) from the provided
+ * timestamp.
+ * @param {number} now
+ * @returns {number} Timestamp for midnight on the current day in ET.
+ */
+export function startOfDayET(now = Date.now()) {
```
kzar, line 178: Do we want to account for daylight savings? In other words, are we using EST here, so always -5, or should we make the offset -4 or -5 depending on the date?

kzar (commented):
Mostly looks good, but just want to check about daylight savings.

kzar:
Do we need to worry about devices that are already enrolled as well?

shared/js/background/components/abn-experiments.js
sammacbeth: The spec is to follow DST (https://app.asana.com/0/0/1209368634935660/1209390565320394/f), but I'm not sure of a single way to handle that in JS, as there is no simple way to set the timezone of a Date object. To do it robustly, we'd probably have to pull in a 3rd-party library, which seems overkill just for this.

Given that, I got an OK that we can just to UTC-5 for simplicity: https://app.asana.com/0/0/1209368634935660/1209390565320405/f

sammacbeth:
> Do we need to worry about devices that are already enrolled as well?

No, before we release this we'll shut down the current experiment, then start a new one, so no need to fix existing enrollments.

shared/js/background/components/abn-experiments.js
kzar: I've not tested it, but after a bit of searching apparently passing  `timeZone: 'America/New_York'` instead of `timeZone: '-05:00'` could work. With similar logic we could figure out the offset for the date as well:

```javascript
function etOffset(date) {
  return new Date(date).toLocaleString('en-US', {
    timeZone: 'America/New_York',
    timeZoneName: 'short'
  }).endsWith('EST') ? -5 : -4;
}
```

I tested that function on Firefox Dev Edition 135.0b9 at least, and it seems to work. As I change the months, I see the "EDT" vs "EST" ending changing, and therefore the offset changing. But if we go with this approach, it's definitely worth testing it on Chrome as well.

Not saying this is a great solution, like you say the best way would be to use a proper date library but that will likely be pretty large. If you'd rather just go with the hardcoded -5 that's OK with me (since the data folks are OK with that), but could you update the comments to explain it's EST not ET and maybe explain the limitation?

kzar (approved):

shared/js/background/components/abn-experiments.js
kzar: @sammacbeth Out of interest, did that suggestion work? (Just curious for next time, I've not used the Date API in that way before.)

sammacbeth: @kzar Yes, testing in Chrome it does seem to work. I decided against using it though because I think there would still be difficult edge cases. For example, we are still assuming that days are 24 hours long over the experiment period, which will not be the case when the experiment starts around a DST change. Sticking with EST is predictably wrong during DST, so it's easy to correct that error on the server side.

PR #3081, opened 2025-05-29
Experimentation C-S-S

jonathanKingston:
Patch: https://app.asana.com/1/137249556945/project/72649045549333/task/1209784651783580?focus=true

**Reviewer:**

## Description:

Adds support for experiments in C-S-S.

## Automated tests:
- [x] Unit tests
- [x] Integration tests

###### Reviewer Checklist:

###### PR Author Checklist:


integration-test/conditional-matching-experiments.spec.js
```diff
@@ -0,0 +1,46 @@
...
+    test('applies correct API manipulation for experiment cohort', async ({
+        context,
+        backgroundPage,
+        page,
+        backgroundNetworkContext,
+        manifestVersion,
+    }) => {
+        // inject.js is outdated in C-S-S meaning this suite doesn't work.
+        // We either should move to using firefox or start building Chrome again.
+        test.skip(manifestVersion === 2, 'MV2 is not supported for this suite, inject.js is outdated');
```
jonathanKingston, line 21: See: https://app.asana.com/1/137249556945/project/1201614831475344/task/1210394080628545?focus=true

TL;DR the mv2 test is using outdated code and so it won't work.

shared/js/background/tab-manager.js
```diff
@@ -29,9 +29,9 @@ class TabManager {
      * 1. When a new tab is opened. See onUpdated listener below
      * 2. When we get a new main_frame request
      */
-    create(tabData) {
+    create(tabData, components) {
```
jonathanKingston: Passing these as a global as using import instead does make it work via utils.js

However: this makes the build system interpret tab.js as ESM and that requires a whole lot of refactoring to match the correct wiring for types all over the codebase.

Where Tab is currently `/* @param {require('path/tab.js')} */` to `/* @param {import {tab} from 'path/tab.js' } */`

This is a simpler step to allow us to get this working but eventually allowing globalThis.components in utils.js would remove some boilerplate.

shared/js/background/components/remote-config.js
```diff
@@ -269,7 +276,7 @@ export function isSubFeatureEnabled(featureName, subFeatureName, config) {
  * @returns {number}
  */
 function getSubFeatureRolloutPercent(subFeature) {
-    if (!subFeature.rollout) {
+    if (!subFeature?.rollout?.steps) {
```
sammacbeth: This seems redundant?
1. The typedoc doesn't say that `subFeature` can be null, so the first `?` is not needed.
2. `steps` is [non-optional](https://github.com/duckduckgo/privacy-configuration/blob/main/schema/feature.ts#L28) on rollout, so the second `?` is only needed if we're passing a broken `SubFeature` object.

shared/js/background/classes/tab.js
```diff
@@ -41,6 +45,24 @@ class Tab {
...
+    /**
+     * Responsible for storing the experiments that ran on the page, so stale tabs don't report the wrong experiments.
+     * @param {AbnExperimentMetrics=} abnMetrics
+     */
+    initExperiments(abnMetrics) {
+        if (!abnMetrics) return;
+        abnMetrics.automaticallyEnrollCurrentContentScopeExperiments();
+        const experiments = abnMetrics.getCurrentCohorts();
+        // Order by key
+        experiments.sort((a, b) => a.subfeature.localeCompare(b.subfeature));
```
sammacbeth: Could we not just provide sorted output from `getCurrentCohorts`?

shared/js/background/components/abn-experiments.js
```diff
@@ -156,6 +156,32 @@ export default class AbnExperimentMetrics {
                 }
             });
     }
+
+    /**
+     * Enroll all current experiments for the 'contentScopeExperiments' feature.
+     */
+    automaticallyEnrollCurrentContentScopeExperiments() {
+        const featureName = 'contentScopeExperiments';
```
sammacbeth: Can we pull this out as a constant at the top of the file?

shared/js/background/tab-manager.js
```diff
@@ -22,6 +26,8 @@ class TabManager {
         this.tabContainer = {};
         /** @type {Record<string, Tab>} */
         this.swContainer = {};
+        /** @type {AbnExperimentMetrics=} */
+        this.abnMetrics = globalThis.components?.abnMetrics;
```
sammacbeth: I think this won't work: `TabManager` is instantiated on import, which will be before `globalThis.components` is constructed. We could either:
 1. Move it to a getter function, so it will be evaluated lazily at access-time.
 2. Set it in the `TabTracker` component before we wire-up all the listeners: https://github.com/duckduckgo/duckduckgo-privacy-extension/blob/main/shared/js/background/components/tab-tracking.js#L21

I think I prefer 2. in this case.

shared/js/background/components/remote-config.js
jonathanKingston: Yeah I had a test that had a rollout with missing steps. I fixed it (and made sure the test files are now typed).

I'll remove and it should still work as you mention :)

sammacbeth (approved):

PR #3262, opened 2025-10-16
Remove failing ATB majorVersion < 500 test

kzar:
This test was first added around five years ago[1], and since the majorVersion
seems to be generated based on the current date relative to the test date, it
has started exceeding 500. Let's just remove it.

While we're at it, let's bump the floor check since we know majorVersion is now
above 500.

1 - https://github.com/duckduckgo/duckduckgo-privacy-extension/commit/9a43e2da975c83c2aecabff3f9c14b4a2b9d540e

**Reviewer:** @moollaza

### DIFF
--- a/unit-test/background/atb-utils.js
+++ b/unit-test/background/atb-utils.js
@@ -7,12 +7,8 @@ describe('utils.getCurrentATB', () => {
         expect(result.majorVersion % 1).toEqual(0);
     });
 
-    it('should return a majorVersion greater than 25', function () {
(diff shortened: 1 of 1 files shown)


unit-test/background/atb-utils.js
```diff
@@ -11,10 +11,6 @@ describe('utils.getCurrentATB', () => {
         expect(result.majorVersion).toBeGreaterThan(25);
```
moollaza: If this is for today's date then I think we could increase the floor to 500?

PR #3275, opened 2025-10-24
Add support for RequestBlocklist feature

kzar:
The RequestBlocklist feature allows for arbitrary requests to be blocked. Being
able to do that will be useful for resolving some website breakage issues, where
the necessary fix is to block a request (rather than to allow one).

This change includes MV2 and MV3 implementations of the feature, unit tests and
integration tests. See the reference tests:
https://github.com/duckduckgo/privacy-reference-tests/tree/main/request-blocklist

**Reviewer:** @jdorweiler 

## Steps to test this PR:
1. Build both the Firefox and Chrome extension: `npm run dev-firefox` and `npm run dev-chrome`.
2. Install those builds in Firefox and Chrome respectively.
3. Navigate to https://reuters.com and ensure images (.jpg requests) load, even after holding shift and clicking refresh.
4. In the extension settings, set the custom config URL of `https://gist.githubusercontent.com/kzar/4cbfeff5d75772b2482a9124a6ea00b7/raw/9b76111d22957a033b8b3deadedd5a2066014c83/extension-config.json`
5. Shift-reload the reuters.com tab again, ensure the .jpg image requests are now blocked, and lots of images are shown as grey placeholders.

If you want to do more testing with your own configuration, here's an example `request-blocklist.json` feature config:
```json
{
    "_meta": {
        "description": "Request blocking rules necessary to fix website breakage. Note: Tracker request blocking is handled elsewhere.",
        "sampleExcludeRecords": {
            "domain": "example.com",
            "reason": "site breakage"
        }
    },
    "state": "enabled",
    "settings": {
        "blockedRequests": {
            "reuters.com": {
                "rules": [
                    {
                        "rule": "reuters.com/*.jpg",
                        "domains": ["reuters.com"],
                        "reason": ["EXAMPLE"]
                    }
                ]
            }
        }
    },
    "exceptions": []
}
```

## Automated tests:
- [x] Unit tests
- [x] Integration tests

### DIFF
--- a/integration-test/request-blocklist.spec.js
+++ b/integration-test/request-blocklist.spec.js
@@ -1,0 +1,54 @@
+import { test, expect } from './helpers/playwrightHarness';
+import { forAllConfiguration, forExtensionLoaded, forDynamicDNRRulesLoaded } from './helpers/backgroundWait';
+import { overridePrivacyConfig, overrideTds } from './helpers/testConfig';
+import { runRequestBlockingTest } from './helpers/requests';
+
+const testHost = 'privacy-test-pages.site';
+const testSite = `https://${testHost}/privacy-protections/request-blocking/`;
+
+function expectBlocked(protectionsEnabled, url) {
+    return protectionsEnabled && new URL(url).pathname.endsWith('.jpg');
+}
+
+test.describe('Test Request Blocklist feature', () => {
+    test('Should block the .jpg requests', async ({ page, backgroundPage, context, backgroundNetworkContext, manifestVersion }) => {
+        await overrideTds(backgroundNetworkContext, 'empty-tds.json');
+        await overridePrivacyConfig(backgroundNetworkContext, 'request-blocklist.json');
+        await forExtensionLoaded(context);
+        await forAllConfiguration(backgroundPage);
+        if (manifestVersion === 3) {
+            await forDynamicDNRRulesLoaded(backgroundPage);
+        }
+
+        for (const protectionsEnabled of [true, false]) {
+            // Disable protections after the first time.
+            if (!protectionsEnabled) {
+                await backgroundPage.evaluate(async (domain) => {
+                    /* global dbg */
+                    dbg.tabManager.setList({ list: 'allowlisted', domain, value: true });
+                }, testHost);
+            }
+
+            // Load and run the request blocking test page.
+            const { testCount, pageRequests } = await runRequestBlockingTest(page, testSite);
+            expect(testCount).toBeGreaterThan(0);
+
+            // Verify that the .jpg image requests were blocked as expected.
+            for (const { url, status } of pageRequests) {
+                // TODO: Figure out why the video.mp4 request is reported as
+                //       blocked regardless.
+                if (new URL(url).pathname.endsWith('video.mp4')) {
+                    continue;
+                }
+
+                expect(status, `URL: ${url}, Allowlisted: ${!protectionsEnabled}`).toEqual(
+                    expectBlocked(protectionsEnabled, url) ? 'blocked' : 'allowed',
+                );
+            }
+
+            await page.reload();
+        }
+
+        await page.close();
+    });
+});
--- a/packages/ddg2dnr/lib/requestBlocklist.js
+++ b/packages/ddg2dnr/lib/requestBlocklist.js
@@ -1,0 +1,97 @@
+/** @module requestBlocklist */
+
+const { getDomain } = require('tldts');
+
+const { processPlaintextTrackerRule, generateDNRRule } = require('./utils');
+
+const EXPECTED_RULE_KEYS = new Set(['domains', 'reason', 'rule']);
+const PRIORITY = 20000;
+
+function validRule(rule) {
+    // Missing or unknown rule properties.
+    const ruleKeys = Object.keys(rule);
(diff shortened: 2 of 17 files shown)


packages/ddg2dnr/lib/requestBlocklist.js
```diff
@@ -0,0 +1,96 @@
...
+
+    for (const [trackerDomain, entry] of Object.entries(entries)) {
+        const { rules } = entry;
+        if (!rules?.length) {
+            continue;
+        }
+
+        // Entries must use eTLD+1.
+        if (getDomain(trackerDomain) !== trackerDomain) {
+            continue;
```
jdorweiler, line 58: Maybe a log message here to help with debugging

packages/ddg2dnr/lib/requestBlocklist.js
```diff
@@ -0,0 +1,96 @@
...
+
+        for (const rule of rules) {
+            if (!validRule(rule)) {
+                continue;
+            }
+
+            let { rule: trackerRule, domains: initiatorDomains, reason } = rule;
+            const { urlFilter } = processPlaintextTrackerRule(trackerDomain, trackerRule);
+
+            if (!initiatorDomains?.length || initiatorDomains.includes('<all>')) {
```
jdorweiler, line 69: check initiatorDomains is array first. `!Array.isArray(initiatorDomains)`

packages/ddg2dnr/lib/requestBlocklist.js
```diff
@@ -0,0 +1,96 @@
+/** @module requestBlocklist */
+
+const { getDomain } = require('tldts');
+
+const { processPlaintextTrackerRule, generateDNRRule } = require('./utils');
+
+const EXPECTED_RULE_KEYS = new Set(['domains', 'reason', 'rule']);
+const PRIORITY = 20000;
```
jdorweiler, line 8: Maybe a comment here so we remember this has to be higher than https://github.com/duckduckgo/duckduckgo-privacy-extension/blob/e48ed182a1c4d2f172a2b3a116e5a74b54e534a3/packages/ddg2dnr/lib/tds.js#L25

packages/ddg2dnr/lib/requestBlocklist.js
```diff
@@ -0,0 +1,96 @@
...
+        for (const rule of rules) {
+            if (!validRule(rule)) {
+                continue;
+            }
+
+            let { rule: trackerRule, domains: initiatorDomains, reason } = rule;
+            const { urlFilter } = processPlaintextTrackerRule(trackerDomain, trackerRule);
+
+            if (!initiatorDomains?.length || initiatorDomains.includes('<all>')) {
+                initiatorDomains = null;
```
jdorweiler, line 70: Is this correct to set to `null` vs empty array?

integration-test/request-blocklist.spec.js
```diff
@@ -0,0 +1,54 @@
...
+
+            // Load and run the request blocking test page.
+            const { testCount, pageRequests } = await runRequestBlockingTest(page, testSite);
+            expect(testCount).toBeGreaterThan(0);
+
+            // Verify that the .jpg image requests were blocked as expected.
+            for (const { url, status } of pageRequests) {
+                // TODO: Figure out why the video.mp4 request is reported as
+                //       blocked regardless.
+                if (new URL(url).pathname.endsWith('video.mp4')) {
```
jdorweiler, line 40: What is this for?

kzar: It seems that the test page has an issue where the video.mp4 request is always reported as being blocked, even when it isn't.

packages/ddg2dnr/lib/requestBlocklist.js
kzar: IMO [the rulePriorities.js test case and comment there](https://github.com/duckduckgo/duckduckgo-privacy-extension/pull/3275/files#diff-07f60b69c4a2b5d712ce09e38d77b81fe9643d52a711868123b86de314d716aaR83-R94) does enough to explain the relative priorities. But that said, I don't mind adding a comment here if you feel strongly about it?

packages/ddg2dnr/lib/requestBlocklist.js
kzar: `initiatorDomains?.length` will be `0` for an empty array, or `undefined` for non-array values (or values that don't have a length property anyway), thanks to the `?.` optional chaining syntax. I don't mind adding that extra check if you feel strongly, but I don't think it will do much.

packages/ddg2dnr/lib/requestBlocklist.js
kzar: I think either is acceptable, at least looking at [the type specified for generateDNRRuleDetails.initiatorDomains](https://github.com/duckduckgo/duckduckgo-privacy-extension/blob/main/packages/ddg2dnr/lib/utils.js#L154). [The logic there](https://github.com/duckduckgo/duckduckgo-privacy-extension/blob/main/packages/ddg2dnr/lib/utils.js#L264-L266) seems to ignore both falsey values and empty arrays in practice too.

packages/ddg2dnr/lib/requestBlocklist.js
kzar: Good shout, done. _(Sorry for the long line, our prettier config forced that 🙃.)_

PR #3286, opened 2025-11-03
Add "unknown" as possible value of browser suffix in pixel schema

kzar:
When firing pixel requests, sometimes the user's current browser is unknown. We
should include that possibility in the schema.

See https://app.asana.com/1/137249556945/task/1211814665268441?focus=true


ladamski (changes requested):
Looks like the `GH_RO_PAT` token that https://github.com/duckduckgo/duckduckgo-privacy-extension/blob/main/.github/workflows/pixel-validation.yml depends on is no longer present, if you update it to `GH_PAT` this should work.

Issue #3197, opened 2025-08-21
Space doubles search term


seniordevel:
If this is the wrong place for my report, please at least tell me where to post it!

This report affects the **web version of DDG only** and **also affects Firefox only** (including latest official release: 142.0) ... did not try the extension.

Since some (quite long) time, not only I have this problem specifically at duckduckgo.com but with Firefox only (other browsers work well).

When I type a space after my searchterm, I do not get searchterm+space (what I would expect), but searchterm+space+searchterm.

So if my search term would be 'test', I get 'test test' instead of 'test ' only.

This only happens after doing a DuckDuckGo search with a _single_ search term. 

Steps to reproduce:

```
1. With Firefox (latest official version), go to duckduckgo.com

2. Type test as search query, press search button (this pre-step is important)
    
3. Touch at the end of the word test in the search field. 

4.Append a single space character 

=> You should see 'test test'

```
This currently happens with Firefox and DDG only.

On Android, I am using Openboard keyboard, but there was neither an update nor did I touch it's settings. There are no extensions installed with Firefox, it is native.

Nevertheless I tried another keyboard to be sure. But almost the same issue there too (however steps to reproduce are slightly different).

With the subdomain https://html.duckduckgo.com/, there is NO issue.
Thus this issue probably arised due to modifications in DDG javascript.

I consider this behavior as very annoying, since often I do a second refined search, with another keyword appended. 

Obviously, more users are affected by this issue, please see Reddit where I posted first:
https://www.reddit.com/r/duckduckgo/comments/1mwe1gj/duckduckgo_issue_space_doubles_search_term/

I am on Android 11, another user on Reddit using  Android 14.

Docga1835:
I am having this issue too. Except now my space bar doesn't work much at all with duckduckgo.com. I am using Fennec F-Droid and Heliboard. No issues with Heliboard as Heliboard works perfectly everywhere except the duckduckgo search engine in my browser. The enter/submit key also doesn't work to send my search; I have to manually tap the search icon next to the text bar.

sammacbeth:
Thanks for filing this. I'm trying to reproduce the issue but currently cannot. Can you confirm the repro steps? I tried the following:

1. Using Firefox on Android, visit duckduckgo.com.
2. Type a search, e.g. 'test' and submit.
3. On the results page, edit the search to add a space.

For me, this does not change the search to 'test test' (I still get 'test').

seniordevel:
[...]

PR #3304, opened 2025-11-22
Fix space doubling in omnibox search terms

notAryan10:
## Description:
This fixes issue #3197 where spaces in search terms were being doubled.

fixed a bug where search terms with spaces were getting messed up.
Earlier, the code was manually sticking pieces of the URL together, which caused spaces to get double-encoded.

Now we use URL and URLSearchParams, which basically handle all the encoding automatically

### DIFF
--- a/shared/js/background/components/omnibox-search.js
+++ b/shared/js/background/components/omnibox-search.js
@@ -11,8 +11,11 @@ export default class OmniboxSearch {
                     currentWindow: true,
                     active: true,
                 });
+                const url = new URL('https://duckduckgo.com');
+                url.searchParams.set('q', text);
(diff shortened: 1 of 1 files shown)


sammacbeth:
While the patch looks correct, the linked issue is not an extension issue, as far as I can tell. Does this patch actually fix an open issue?

notAryan10:
> While the patch looks correct, the linked issue is not an extension issue, as far as I can tell. Does this patch actually fix an open issue?

The patch doesn't address the reported issue. The problem needs to be fixed in the main DuckDuckGo website's frontend code not in the privacy extension 

also the issue is in the website's search input handling

If you like i could work on that too...

PR #3354, opened 2026-01-28
Workaround some integration test failures

kzar:
I found issues running two of our integration tests:

1. The conditional matching experiment tests seem to fail since the test page
   entirely fails to load. That was related to the `routeFromLocalhost(page);`
   call, so let's get rid of the routeFromLocalhost call  for now to get the
   test running again.
2. One of the canvas fingerprinting test cases is failing for me consistently,
   let's skip that test for now.


integration-test/conditional-matching-experiments.spec.js
jonathanKingston:
```suggestion
// TODO: Re-enable this when it is working again.
// import { routeFromLocalhost } from './helpers/testPages';
```

jonathanKingston (approved):

PR #3330, opened 2025-12-18
Playwright test route error

jonathanKingston:
**Reviewer:** @kzar 

https://app.asana.com/1/137249556945/project/1199237043598868/task/1212423549860686?focus=true

## Description:
Fixes Playwright test failures ("Target page, context or browser has been closed") caused by unhandled route callbacks firing during teardown.

This PR addresses the issue by:
1.  Implementing `await context.unrouteAll({ behavior: 'ignoreErrors' })` and `await page.unrouteAll({ behavior: 'ignoreErrors' })` during fixture teardown to gracefully remove all active routes before pages/contexts close.
2.  Ensuring all route handlers consistently `return` or `await` their `route.fulfill()` or `route.continue()` promises to prevent detached async operations that could outlive the test and cause errors.

## Steps to test this PR:
1.  Run `npm ci`
2.  Run `npm run test:integration -- --shard=2/4` (or the full `npm run test:integration` suite) to confirm the error no longer occurs.

## Automated tests:
- [ ] Unit tests
- [x] Integration tests

###### Reviewer Checklist:
- [ ] **Ensure the PR solves the problem**
- [ ] **Review every line of code**
- [ ] **Ensure the PR does no harm by testing the changes thoroughly**
- [ ] **Get help if you're uncomfortable with any of the above!**
- [ ] Determine if there are any quick wins that improve the implementation

###### PR Author Checklist:
- [ ] Get advice or leverage existing code
- [ ] Agree on technical approach with reviewer (if the changes are nuanced)
- [ ] Ensure that there is a testing strategy (and documented non-automated tests)
- [ ] Ensure there is a documented monitoring strategy (if necessary)
- [ ] Consider systems implications

---
<a href="https://cursor.com/background-agent?bcId=bc-3881ccad-26e3-4055-bd4a-441c0f8bd1d1"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/open-in-cursor-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/open-in-cursor-light.svg">Open in Cursor</picture></a>&nbsp;<a href="https://cursor.com/agents?id=bc-3881ccad-26e3-4055-bd4a-441c0f8bd1d1"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/open-in-web-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/open-in-web-light.svg">Open in Web</picture></a>

---

> [!NOTE]
[...]

PR #3357, opened 2026-01-28
Improve Playwright harness reliability and error messages

kzar:
Unfortunately the Playwright tests still aren't running entirely reliably. At
least part of the issue is the Playwright harness, for example sometimes it
fails to find the extension's background ServiceWorker. Let's do what we can to
reduce those issues, and let's also improve the output when we still run into
problems.

Note: H/T to @jonathanKingston, who spotted the missing context.route
      awaits with https://github.com/duckduckgo/duckduckgo-privacy-extension/pull/3330.
      Those seemed to cause more error message noise related to route callbacks when
      the tests flake.

### DIFF
--- a/integration-test/helpers/playwrightHarness.js
+++ b/integration-test/helpers/playwrightHarness.js
@@ -112,7 +112,7 @@ export const test = base.extend({
                 try {
                     background = await context.waitForEvent('serviceworker', { timeout: 5000 });
                 } catch (e) {
-                    ([background] = context.serviceWorkers());
+                    [background] = context.serviceWorkers();
                     if (!background) {
                         // Playwright sometimes fails in this way, which unfortunately is a common
                         // source of test flakes.


integration-test/helpers/playwrightHarness.js
```diff
@@ -104,11 +104,25 @@ export const test = base.extend({
...
         if (manifestVersion === 3) {
             // See https://playwright.dev/docs/service-workers
             let [background] = context.serviceWorkers();
-            if (!background) background = await context.waitForEvent('serviceworker');
-            context.route('**/*', routeHandler);
+            if (!background) {
+                try {
+                    background = await context.waitForEvent('serviceworker', { timeout: 5000 });
+                } catch (e) {
+                    ([background] = context.serviceWorkers());
```
kzar:
```suggestion
                    [background] = context.serviceWorkers();
```

PR #3360, opened 2026-01-29
Add breakageData parameter to breakage reports

GuiltyDolphin:
https://app.asana.com/1/137249556945/project/1202561462274611/task/1213015238129926?focus=true

**CC:** @jdorweiler
**Depends on:** 

## Description:

Adds the `breakageData` parameter to breakage reports.

The value of this parameter is generated in the breakage reporting feature of Content Scope Scripts and should be passed along as-is to breakage reports.

## Steps to test this PR:

1. Make sure you're using the latest Content Scope Scripts
2. Add logging around `submitBrokenSiteReport` with the `breakageData` from CSS (or otherwise monitor breakage payload)
3. Ensure the `breakageReporting` feature is enabled via config
4. Visit a site and edit the page text to include "disable your adblocker"
5. Make a breakage report, check:
    - you can see the `breakageData` parameter in the response from CSS
    - you can see `&breakageData=` in the breakage report request, and that this exactly matches the value received from CSS without any double-encoding 

## Automated tests:
- [X] Unit tests
- [ ] Integration tests

###### Reviewer Checklist:

###### PR Author Checklist:

---

> [!NOTE]
> **Low Risk**
> Small, additive change to pixel URL construction and breakage reporting payloads; main risk is malformed URLs or unexpected truncation behavior if `breakageData` is large.
> 
> **Overview**
> Breakage report pixels now include an optional `breakageData` query parameter sourced from content-scope-scripts and **passed through without double-encoding**.
> 
> To support this, `broken-site-report.fire()`/URL construction accepts a set of *pre-encoded* params that are appended directly (and still omitted when truncating), `DashboardMessaging` forwards `breakageData` into `pageParams`, and unit tests assert presence/absence and no re-encoding.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 3443c412fc8ad0caccd01c06f6cca75bc7817d18. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/shared/js/background/broken-site-report.js
+++ b/shared/js/background/broken-site-report.js
@@ -60,22 +60,32 @@ const PARAM_IDS = [
  *
  * Fire a pixel
  *
+ * @param {string} pixelName
  * @param {string} querystring
+ * @param {Record<string, string>} [encodedParams]
  *
  */
-export function fire(pixelName, querystring) {
-    let url = constructUrl(pixelName, querystring, false);
+export function fire(pixelName, querystring, encodedParams = {}) {
+    let url = constructUrl(pixelName, querystring, false, encodedParams);
 
     // If we're over the max pixel length, truncate the less important params
     if (url.length > maxPixelLength) {
-        url = constructUrl(pixelName, querystring, true);
+        url = constructUrl(pixelName, querystring, true, encodedParams);
     }
 
     // Send the request
     load.url(url);
 }
 
-function constructUrl(pixelName, querystring, truncate) {
+/**
+ *
+ * @param {string} pixelName
+ * @param {string} querystring
+ * @param {boolean} truncate
+ * @param {Record<string, string>} [encodedParams]
+ * @returns
+ */
+function constructUrl(pixelName, querystring, truncate, encodedParams = {}) {
     const randomNum = Math.ceil(Math.random() * 1e7);
     const browserInfo = parseUserAgentString();
     const browserName = browserInfo?.browser;
@@ -113,6 +123,11 @@ function constructUrl(pixelName, querystring, truncate) {
             searchParams.delete(key);
         }
     });
+    // encodedParams are pre-encoded. Append directly without re-encoding
+    for (const [key, value] of Object.entries(encodedParams)) {
+        if (truncate && truncatableFields.includes(key)) continue;
+        extraParams += `&${key}=${value}`;
+    }
     url += `${searchParams.toString()}${extraParams}`;
     return url;
 }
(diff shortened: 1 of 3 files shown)


shared/js/background/broken-site-report.js
```diff
@@ -113,6 +123,11 @@ function constructUrl(pixelName, querystring, truncate) {
             searchParams.delete(key);
         }
     });
+    // encodedParams are pre-encoded. Append directly without re-encoding
+    for (const [key, value] of Object.entries(encodedParams)) {
+        if (truncate && truncatableFields.includes(key)) continue;
+        extraParams += `&${key}=${value}`;
```
jonathanKingston, line 129: Can't we add this to `searchParams` instead here, rather than hand encoding?

GuiltyDolphin: @jonathanKingston if we add to `searchParams` then the string will get double-encoded.

We're not actually doing any encoding here--the string has already been encoded by CSS, and per the intent of that, we should pass it along exactly as-is.

E..g,

```
const qs = new URLSearchParams();
qs.set('param', '%25');
qs.toString();
// param=%2525 -- bad!
```

jonathanKingston: I see, I know this is pedantic but we could unpack and then do `set`. I'm open to doing as-is, just usually this is a bit of a smell doing your own param `=` etc.

GuiltyDolphin: Doing this as-is should fit the intended semantics better (value to be passed along as-is). Unencoding then re-encoding could lead to unexpected results e.g., `foo,bar` -> `foo,bar` -> `foo%2Cbar` (!= original `foo,bar`).

Would be good if the param builder allowed you to say "don't encode this param!", but couldn't find any support for that.

shared/js/background/broken-site-report.js
```diff
@@ -264,6 +279,8 @@ export async function breakageReportForTab({
     const locale = tab.locale;
     const contentScopeExperiments = tab.contentScopeExperiments;
     const detectorData = pageParams.detectorData ? JSON.stringify(pageParams.detectorData) : undefined;
+    // breakageData is pre-encoded by content-scope-scripts, pass as-is without re-encoding
+    const breakageData = pageParams.breakageData !== undefined ? pageParams.breakageData : undefined;
```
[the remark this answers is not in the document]
GuiltyDolphin: CSS either provides a string or does not provide the parameter at all. Doesn't send `null`.

shared/js/background/broken-site-report.js
jonathanKingston: yeah don't worry about it too much :).
Thanks for humouring me.

jonathanKingston (approved):

jonathanKingston:
🚢 !

PR #3363, opened 2026-01-30
Get the fire animation integration test case passing more reliably

kzar:
Instead of waiting for exactly three seconds before checking that the page was
redirected, instead wait for the redirection to finish.


integration-test/fire-button.spec.js
jonathanKingston, line 242: Should we be worried about this error?

Image

kzar: Oh interesting, I didn't spot that. I think there's a separate flake here that's worth digging into in the future. In my latest big test run (with all the prepared changes applied), the Fire button tests are now some of the least reliable. So this flake you spotted could explain why. I'll take a look this weekend if I get a chance.

PR #3364, opened 2026-01-31
Ensure manuallyWaitForFunction Promise doesn't attempt to resolve twice

kzar:
The (work in progress) Firefox Playwright harness is failing sometimes, when the
manuallyWaitForFunction function attempts to resolve/reject the returned Promise
more than once. Let's prevent that from happening, ready for when we add the new
harness.


jdorweiler (approved):
Looks good

PR #3392, opened 2026-02-06
Get the integration tests for storage blocking running more reliably

kzar:
The storage blocking integration tests were failing quite often. It turned out
that the test page[1] loads several iframes, and gives up waiting for them to
load after one second. When they don't load quickly enough,
`waitForAllResults()` never resolves, and the tests will timeout after 30
seconds.

Usually, the answer would be to route the iframe requests locally using
Playwright. Unfortunately, in this instance we can't do that because doing so
prevents the extension's cookie protections from being applied. Instead, let's
workaround the issue for now by having the testpage wait for up to 10 seconds
for those iframes to load instead.

1 - https://privacy-test-pages.site/privacy-protections/storage-blocking

### DIFF
--- a/integration-test/storage-blocking.spec.js
+++ b/integration-test/storage-blocking.spec.js
@@ -4,6 +4,7 @@ import { overridePrivacyConfig, overrideTds } from './helpers/testConfig';
 import { TEST_SERVER_ORIGIN, routeFromLocalhost } from './helpers/testPages';
 
 const testPageDomain = 'privacy-test-pages.site';
+const testPageParams = '?timeout=10000';
 const thirdPartyDomain = 'good.third-party.site';
 const thirdPartyTracker = 'broken.third-party.site';
 const thirdPartyAd = 'convert.ad-company.site';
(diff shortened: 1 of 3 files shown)


integration-test/storage-blocking.spec.js
```diff
@@ -26,6 +51,7 @@ test.describe('Storage blocking Tests', () => {
         await backgroundWait.forExtensionLoaded(context);
         await backgroundWait.forAllConfiguration(backgroundPage);
         await routeFromLocalhost(page);
+        await increaseIframeTimeout(page);
         await page.bringToFront();
         await page.goto(`https://${testPageDomain}/privacy-protections/storage-blocking/?store`, { waitUntil: 'networkidle' });
```
jonathanKingston: I think it might be cleaner just to set a `timeout` argument here, we own the test and so I think it'd be less 'spooky action at a distance'.

Happy to approve if you don't want to do the effort.

jonathanKingston: (ddg-workflow in the cloud might be able to do both parts here if you show it the patch).

kzar: That's a good call-out, I have drafted the test pages change here https://github.com/duckduckgo/privacy-test-pages/pull/406

kzar: Done.

jonathanKingston (approved):

PR #3393, opened 2026-02-07
Get the HTTPS loop protection tests running more reliably

kzar:
The HTTPS loop protection feature aims to protect the user from infinite
redirection loops caused by upgrading HTTP navigations to HTTPS. It works by
activating if seven such redirections happening within a three second window.

When the integration tests were running slowly, the test page's redirections
didn't happen within that three second window, so the tests would fail. Let's
add a check, so that when the tests are running too slowly to work, we skip them
instead.

### DIFF
--- a/integration-test/https-loop-protection.spec.js
+++ b/integration-test/https-loop-protection.spec.js
@@ -10,14 +10,37 @@ test.describe('Loop protection', () => {
         await backgroundWait.forAllConfiguration(backgroundPage);
         await routeFromLocalhost(page);
 
+        // HTTPS loop detection considers REQUEST_REDIRECT_LIMIT upgrades within
+        // MAINFRAME_RESET_MS to be a loop.
+        const [REQUEST_REDIRECT_LIMIT, MAINFRAME_RESET_MS] = await backgroundPage.evaluate(() => [
+            globalThis.dbg.HttpsRedirects.REQUEST_REDIRECT_LIMIT,
+            globalThis.dbg.HttpsRedirects.MAINFRAME_RESET_MS,
+        ]);
+
+        const upgradeTimes = [];
+        context.on('request', (request) => {
+            if (request.url().startsWith('https://') && request.url().includes('http-only.html')) {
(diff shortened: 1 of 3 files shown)


integration-test/https-loop-protection.spec.js
```diff
@@ -10,14 +10,37 @@ test.describe('Loop protection', () => {
...
-        });
 
-        // The expected outcome of this test is we land back on http:// rather than upgrading forever to https://
+        // When the tests are running too slowly, the redirections don't happen
+        // quickly enough to trigger the loop protection feature.
+        // Note: Use the request after REQUEST_REDIRECT_LIMIT, since that's when
+        //       the feature's canRedirect check happens.
+        const followingRequest = REQUEST_REDIRECT_LIMIT + 1;
+        test.skip(
+            !(upgradeTimes.length > followingRequest && upgradeTimes[followingRequest] - upgradeTimes[0] < MAINFRAME_RESET_MS),
```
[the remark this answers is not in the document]
kzar: _(After a lot of debugging and running the tests locally, I am sure this is the right timestamp to use. If we use REQUEST_REDIRECT_LIMIT, we still get a lot of failures under load.)_

shared/js/background/classes/https-redirects.js
```diff
@@ -15,6 +13,9 @@ const manifestVersion = browserWrapper.getManifestVersion();
  */
 
 class HttpsRedirects {
+    static MAINFRAME_RESET_MS = 3000;
```
jonathanKingston, line 16: Ouch and we're hitting 3s often in CI?

jonathanKingston (approved):

shared/js/background/classes/https-redirects.js
kzar: I don't think it happens too often on CI, but it was pretty consistent when I ran the flakey stress test locally. After the first 10 or 20 runs, my laptop started throttling and I was getting overall like 80% failure rate or something. They pass consistently for me locally usually, even with a 10x run though.

PR #3395, opened 2026-02-10
Add an integration test for empty-config handling

kzar:
We want to make sure that if an empty (but valid) remote configuration or
blocklist are loaded, the extension handles that OK and doesn't break. Let's add
a test that covers that situation now.


kzar:
> Changes look solid to me, some nice simplifications in there. Not sure what's going on with the pixel schema validation check, mind following up on that before we merge?

@dharb  Thanks! Don't worry, that failure is just because the branch is on my fork, some of the CI jobs like that won't run from forks unfortunately.

PR #3439, opened 2026-03-12
No AI in Searches toggle on options page

sammacbeth:
## Description:
Adds a single **No AI in Searches** privacy option in the settings UI and removes the **Show embedded Tweets** option.

When enabled, searches are routed through the existing `alternativeSearch` setting to `noai.duckduckgo.com`; when disabled, searches remain on `duckduckgo.com`.

### What changed
- Added the **No AI in Searches** toggle in options
- Removed the **Show embedded Tweets** option from the options model/view/template stack
- Added supporting copy in `shared/locales/en/options.json` and synced generated browser locale manifests
- Updated background redirect handling for `alternativeSearch` in MV2/MV3 code paths (`before-request.js`, `atb.js`, `dnr-utils.js`)
- Added integration coverage in `integration-test/search-choice.spec.js`

### Options page refinements included in this branch
- Added description text under the toggle
- Standardized heading spacing
- Reordered sections to match design (Unprotected Sites above Email Protection)
- Added divider treatment around Unprotected Sites / Email Protection sections
- Included CSS/layout tweaks to prevent overlap and click-target regressions

[Final options page](https://cursor.com/agents/bc-25aa6ce9-35fb-410f-b339-5c115147640c/artifacts?path=%2Fopt%2Fcursor%2Fartifacts%2Foptions_divider_fix.png)

## Steps to test this PR:
1. Build the extension (`make dev browser=chrome type=dev`).
2. Load the extension in a browser.
3. Navigate to the extension's options page.
4. Verify the **No AI in Searches** toggle appears with description text.
5. Toggle it on, then search on duckduckgo.com — should redirect to noai.duckduckgo.com.
6. Toggle it off — searches should stay on duckduckgo.com.
7. Verify section ordering, spacing, and dividers on the options page match design.
8. Verify allowlist area remains clickable and layout does not overlap in edge cases.

## Automated tests:
- [x] Integration tests (`integration-test/search-choice.spec.js` — 4 Playwright tests)

###### Reviewer Checklist:

###### PR Author Checklist:

<sub>To show artifacts inline, <a href=\"https://cursor.com/dashboard/cloud-agents#team-pull-requests\">enable</a> in settings.</sub>


integration-test/atb.spec.js
```diff
@@ -202,4 +202,26 @@ test.describe('search workflow', () => {
         expect(newSetAtb).toEqual(todaysAtb);
         expect(atb).toEqual('v123-1');
     });
+
+    test('should redirect searches to the alternative search subdomain when enabled', async ({ backgroundPage, page }) => {
+        await backgroundPage.evaluate(() => globalThis.dbg.settings.updateSetting('alternativeSearch', 'noai'));
```
kzar: Would be good to reuse the `setAlternativeSearch()` helper in this file (same below), should we share that from a helpers module?

integration-test/search-choice.spec.js
```diff
@@ -0,0 +1,94 @@
...
+/**
+ * Update the alternativeSearch setting and wait for the DNR rule to be applied.
+ * The setting update is async (dispatches event after storage sync), so the
+ * DNR rule isn't installed immediately.
+ */
+async function setAlternativeSearch(backgroundPage, value) {
+    await backgroundPage.evaluate(async (val) => {
+        globalThis.dbg.settings.updateSetting('alternativeSearch', val);
+        // Wait for the storage sync + event dispatch + DNR rule install to complete.
+        await new Promise((resolve) => setTimeout(resolve, 1000));
```
kzar: I suspected the one second wait here wouldn't be enough to make the tests pass reliably, so I checked and unfortunately could reproduce quite a few flakes as I feared.

After some trial and error, Claude found a more reliable approach, by first updating the `"alternativeSearch"` listener to wait for the ATB:

```diff
diff --git a/shared/js/background/atb.js b/shared/js/background/atb.js
index 2291d02d7..8d264418c 100644
--- a/shared/js/background/atb.js
+++ b/shared/js/background/atb.js
@@ -298,7 +298,11 @@ settings.ready().then(() => {
         ATB.setOrUpdateATBdnrRule(atb);
     });
     settings.onSettingUpdate.addEventListener('set_atb', updateUninstallURL);
-    settings.onSettingUpdate.addEventListener('alternativeSearch', () => {
+    settings.onSettingUpdate.addEventListener('alternativeSearch', async () => {
+        // setOrUpdateATBdnrRule short-circuits when atb is empty, which can
+        // happen if the SW recycled before the initial atb storage write
+        // flushed. Recover before installing the search rule.
+        if (!settings.getSetting('atb')) await ATB.setInitialVersions();
         ATB.setOrUpdateATBdnrRule(settings.getSetting('atb'));
     });
 });
```
Then by updating the atb tests to use the `setAlternativeSearch` helper as well:

```diff
iff --git a/integration-test/atb.spec.js b/integration-test/atb.spec.js
index 9043ea24b..0b48d8186 100644
--- a/integration-test/atb.spec.js
+++ b/integration-test/atb.spec.js
@@ -1,6 +1,7 @@
 import { test, expect, mockAtb } from './helpers/playwrightHarness';
 import backgroundWait, { forSetting } from './helpers/backgroundWait';
 import { logPageRequests } from './helpers/requests';
+import { setAlternativeSearch } from './helpers/setAlternativeSearch';
 
 test.describe('install workflow', () => {
     test('postinstall page: should open the postinstall page correctly', async ({ context, page }) => {
@@ -204,7 +205,7 @@ test.describe('search workflow', () => {
     });
 
     test('should redirect searches to the alternative search subdomain when enabled', async ({ backgroundPage, page }) => {
-        await backgroundPage.evaluate(() => globalThis.dbg.settings.updateSetting('alternativeSearch', 'noai'));
+        await setAlternativeSearch(backgroundPage, 'noai');
 
         await page.goto('https://duckduckgo.com/?q=alternative-search-test', { waitUntil: 'networkidle' });
 
@@ -215,7 +216,7 @@ test.describe('search workflow', () => {
     });
 
     test('should keep searches on duckduckgo.com when alternative search is disabled', async ({ backgroundPage, page }) => {
-        await backgroundPage.evaluate(() => globalThis.dbg.settings.updateSetting('alternativeSearch', ''));
+        await setAlternativeSearch(backgroundPage, '');
 
         await page.goto('https://duckduckgo.com/?q=alternative-search-disabled-test', { waitUntil: 'networkidle' });
```

And then by finally updating the `setAlternativeSearch helper to wait for the DNR rules to be ready (where relevant):

```javascript
export async function setAlternativeSearch(backgroundPage, value) {
    await backgroundPage.evaluate(async (val) => {
        // Matches SEARCH_REDIRECT_RULE_ID in shared/js/background/dnr-utils.js
        const SEARCH_REDIRECT_RULE_ID = 20009;
        globalThis.dbg.settings.updateSetting('alternativeSearch', val);
        if (chrome.declarativeNetRequest?.getDynamicRules) {
            const want = val !== '';
            // eslint-disable-next-line no-constant-condition
            while (true) {
                const rules = await chrome.declarativeNetRequest.getDynamicRules();
                if (rules.some((r) => r.id === SEARCH_REDIRECT_RULE_ID) === want) return;
                await new Promise((resolve) => setTimeout(resolve, 25));
            }
        }
    }, value);
}
```

I suppose instead of polling with getDynamicRules, a cleaner option might be to await the `settings.updateSetting('alternativeSearch', val)` call, and then make sure the plumbing there so that the returned Promise doesn't resolve until any corresponding work like adding dynamic rules is done, but I'm guessing that might be complicated in practice, if it requires a lot of plumbing or risks causing deadlocks and things like that.

I don't feel too strongly exactly how we do it anyway, so long as the tests pass reliably. I was reproducing the flakes with commands like:

```bash
npm run playwright -- integration-test/atb.spec.js integration-test/search-choice.spec.js --repeat-each=20 --workers=10 --reporter=dot --retries=0
npm run playwright-mv2 -- integration-test/atb.spec.js integration-test/search-choice.spec.js --repeat-each=20 --workers=10 --reporter=dot --retries=0
```

(*FWIW you might need to up the workers + repeat counts to reproduce the flakes if your computer is more powerful, since I generally find flakes happen more when the device is struggling and things slow down.*)

shared/js/ui/models/privacy-options.js
```diff
@@ -21,7 +21,11 @@ PrivacyOptions.prototype = window.$.extend({}, Parent.prototype, {
         if (Object.hasOwnProperty.call(this, k)) {
             this[k] = !this[k];
             console.log(`PrivacyOptions model toggle ${k} is now ${this[k]}`);
-            this.sendMessage('updateSetting', { name: k, value: this[k] });
+            if (k === 'noAiMode') {
```
kzar: I don't quite follow why we need this change, I figure we could store a bool to say noai is enabled, and therefore keep the existing logic?

shared/js/ui/models/privacy-options.js
```diff
@@ -32,13 +36,13 @@ PrivacyOptions.prototype = window.$.extend({}, Parent.prototype, {
         ]);
 
         this.httpsEverywhereEnabled = settings.httpsEverywhereEnabled;
-        this.embeddedTweetsEnabled = settings.embeddedTweetsEnabled;
```
kzar, line 35: Seems unrelated?

integration-test/search-choice.spec.js
```diff
@@ -0,0 +1,94 @@
...
+        expect(page.url()).toContain('duckduckgo.com/?q=test');
+        expect(page.url()).not.toContain('noai.duckduckgo.com');
+    });
+
+    // Playwright's context.route() intercepts start.duckduckgo.com before the
+    // extension's redirect mechanism can fire (DNR on MV3, webRequest on MV2).
+    // Unlike duckduckgo.com/?q= which also matches ATB's rules and goes through
+    // the network stack, start.duckduckgo.com has no other matching rule. The
+    // redirect works in actual Chrome; this is a test infrastructure limitation.
+    test.skip('redirects start.duckduckgo.com when alternativeSearch is enabled', async ({ context, backgroundPage, page }) => {
```
kzar: Are you sure that this skip is not masking a real issue? For example, I figure the current logic in before-request.js wouldn't match subdomains, e.g. for a URL like https://start.duckduckgo.com/?q=test.

shared/js/background/atb.js
```diff
@@ -232,10 +233,28 @@ const ATB = (() => {
...
+                        id: SEARCH_REDIRECT_RULE_ID,
+                        priority: ALTERNATIVE_SEARCH_PRIORITY,
+                        actionType: 'redirect',
+                        redirect: {
+                            transform: {
+                                host: `${alternativeSearchSubdomain}.duckduckgo.com`,
+                            },
+                        },
+                        resourceTypes: ['main_frame'],
+                        requestDomains: ['duckduckgo.com'],
```
kzar: Nit: I don't think this `requestDomains` condition does anything, since we already have the `urlFilter` condition anchored to the same domain.

shared/js/background/before-request.js
```diff
@@ -141,7 +141,17 @@ function handleRequest(requestData) {
         // add atb params only to main_frame
         const atbParametersAdded = ATB.addParametersMainFrameRequestUrl(mainFrameRequestURL);
 
-        if (urlParametersRemovedForThisRequest || ampRedirected || atbParametersAdded) {
+        // apply alternative search redirect (e.g. noai.duckduckgo.com)
```
kzar: Seems inconsistent with the MV3 equivalent rule? E.g. I figure `https://duckduckgo.com/?ia=images` would be redirected on MV3 but not on MV2?

integration-test/atb.spec.js
```diff
@@ -202,4 +202,26 @@ test.describe('search workflow', () => {
...
+
+        await page.goto('https://duckduckgo.com/?q=alternative-search-test', { waitUntil: 'networkidle' });
+
+        const redirectedUrl = new URL(page.url());
+        expect(redirectedUrl.hostname).toEqual('noai.duckduckgo.com');
+        expect(redirectedUrl.pathname).toEqual('/');
+        expect(redirectedUrl.searchParams.get('q')).toEqual('alternative-search-test');
```
kzar, line 215: Should we add an assertion that the ATB param is correct too? (Same below.)

kzar (changes requested):

shared/js/ui/models/privacy-options.js
sammacbeth: We removed this when updating the settings design. This toggle hasn't done anything for a long time. https://app.asana.com/1/137249556945/task/1213721679545335/comment/1214231032235778?focus=true

shared/js/ui/models/privacy-options.js
sammacbeth: Fixed this by making the setting a boolean.

integration-test/search-choice.spec.js
sammacbeth: This is an LLM relic. As you say, we're not redirecting subdomains. I've removed this.

shared/js/background/before-request.js
sammacbeth: I've updated the MV2 rule to match MV3 behavior. I don't think you can currently get anything except the homepage without specifying `q`, but still we should make them behave the same.

shared/js/ui/models/privacy-options.js
kzar: OK got it, sounds good to me to remove it to be fair.

integration-test/atb.spec.js
```diff
@@ -202,4 +203,28 @@ test.describe('search workflow', () => {
...
+        expect(redirectedUrl.searchParams.get('atb')).toMatch(/^v[\d-]+$/);
+    });
+
+    test('should keep searches on duckduckgo.com when no AI search is disabled', async ({ backgroundPage, page }) => {
+        await setUseNoAiSearch(backgroundPage, false);
+
+        await page.goto('https://duckduckgo.com/?q=alternative-search-disabled-test', { waitUntil: 'networkidle' });
```
kzar: I did some testing today, and I found that switching `'networkidle'` to `'domcontentloaded'` for all the `page.goto()` calls in atb.spec.js reduced the flakes further and sped up the the tests. Some of them existed before your changes here, but seems worth doing?

shared/js/background/before-request.js
kzar: Thanks, SGTM :+1:

shared/js/background/atb.js
```diff
@@ -232,10 +233,27 @@ const ATB = (() => {
...
+                        id: SEARCH_REDIRECT_RULE_ID,
+                        priority: ALTERNATIVE_SEARCH_PRIORITY,
+                        actionType: 'redirect',
+                        redirect: {
+                            transform: {
+                                host: 'noai.duckduckgo.com',
+                            },
+                        },
+                        resourceTypes: ['main_frame'],
+                        urlFilter: '||duckduckgo.com/?',
```
kzar: How about we add a `excludedRequestDomains: ['noai.duckduckgo.com'],` condition to the noai redirection rule? I don't _think_ the declarativeNetRequest APIs would get stuck in a redirection loop anyway, but seems worth ruling it out?

kzar (changes requested):

shared/js/background/before-request.js
```diff
@@ -141,7 +141,17 @@ function handleRequest(requestData) {
         // add atb params only to main_frame
         const atbParametersAdded = ATB.addParametersMainFrameRequestUrl(mainFrameRequestURL);
 
-        if (urlParametersRemovedForThisRequest || ampRedirected || atbParametersAdded) {
+        // apply no AI search redirect (noai.duckduckgo.com)
+        const shouldRedirectSearch =
+            mainFrameRequestURL.hostname === 'duckduckgo.com' &&
```
kzar, line 146: FWIW I think this is slightly different for MV2 vs MV3, since here we checking hostname is exactly `duckduckgo.com` but for MV3 we're also matching subdomains. I'm not sure how important that is, but if you like I guess we could make MV3 consistent by matching only `duckduckgo.com` by using a `regexFilter` condition instead of the `urlFilter` condition, we do [something similar (but more complicated) for smarterEncryption](https://github.com/duckduckgo/duckduckgo-privacy-extension/blob/main/packages/ddg2dnr/lib/smarterEncryption.js#L7-L9). I suppose that might have the other bonus of making the `excludedRequestDomains: ['noai.duckduckgo.com'],` suggestion redundant as well?

sammacbeth: You're right, I missed an edge case here. The MV3 rule would also redirect e.g. searches on safe.duckduckgo.com to noai, which I don't think we want to do here. I've added the regex change (which will also make `excludedRequestDomains` redundant.

shared/js/background/atb.js
```diff
@@ -232,10 +233,27 @@ const ATB = (() => {
...
+                        id: SEARCH_REDIRECT_RULE_ID,
+                        priority: ALTERNATIVE_SEARCH_PRIORITY,
+                        actionType: 'redirect',
+                        redirect: {
+                            transform: {
+                                host: 'noai.duckduckgo.com',
+                            },
+                        },
+                        resourceTypes: ['main_frame'],
+                        regexFilter: '^https://duckduckgo.com/\\?.*',
```
kzar:
```suggestion
                        regexFilter: '^https://duckduckgo\\.com/\\?.*',
```

kzar (approved):
LGTM % the regexp escaping _(Note: I didn't re-test again with these latest changes.)_

PR #3588, opened 2026-07-07
Add timeouts to native messages

muodov:
Asana task: https://app.asana.com/1/137249556945/project/1163321984198618/task/1216031389682952?focus=true

**Reviewer:**

## Description:
Speculative fix to avoid theoretical infinite deadlock in native messaging

## Automated tests:
- [ ] Unit tests
- [ ] Integration tests

###### Reviewer Checklist:

###### PR Author Checklist:

---

> [!NOTE]
> **Medium Risk**
> Changes behavior for all embedded native CPM IPC; a slow-but-valid native reply past 10s would now fail and fall back to error handling/cached config instead of waiting indefinitely.
> 
> **Overview**
> Adds **`NATIVE_MESSAGE_TIMEOUT_MS` (10s)** and a **`withTimeout`** helper that races native IPC promises against a timer and rejects with a labeled error if no reply arrives.
> 
> **`CPMEmbeddedMessaging`** now wraps **`nativeMessaging.notify`**, **`nativeMessaging.request`** (queued paths), and the out-of-queue **`getResourceIfNew`** call in **`refreshRemoteConfig`** with that timeout. Hung calls surface through existing queue **`.catch`** handling (diagnostics + logging) instead of leaving **`_queue`** stuck indefinitely; timed-out **`_request`** paths do not write to the response cache.
> 
> Unit tests cover hung **`_notify`** / **`_request`** (queue unblocks, diagnostics), no cache after timeout, and **cached config fallback** when config refresh times out. Embedded extension **manifest version** bumps to **2026.7.7**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 25a255ce6806ab58cd2a7950ae65acdeebc95e76. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/shared/js/background/components/cpm-embedded-messaging.js
+++ b/shared/js/background/components/cpm-embedded-messaging.js
@@ -14,6 +14,29 @@ export const SETTING_CHECK_TTL = 10 * 1000; // 10s
 export const SITE_CHECK_TTL = 3 * 1000; // 3s (to be able to respond to Protections toggle changes)
 export const MAX_CACHE_SIZE = 100;
 
+// Upper bound for a single native message round-trip. Native messages are local IPC and normally
+// complete in milliseconds, but a dropped reply may leave the promise pending forever.
+export const NATIVE_MESSAGE_TIMEOUT_MS = 10 * 1000; // 10s
+
+/**
+ * Reject if `promise` does not settle within `timeoutMs`
+ * @template T
+ * @param {Promise<T>} promise
+ * @param {number} timeoutMs
+ * @param {string} label - method name, included in the timeout error for diagnostics
+ * @returns {Promise<T>}
+ */
+function withTimeout(promise, timeoutMs, label) {
+    /** @type {ReturnType<typeof setTimeout>} */
+    let timer;
+    const timeout = new Promise((_resolve, reject) => {
+        timer = setTimeout(() => {
+            reject(new Error(`native message "${label}" timed out after ${timeoutMs}ms`));
+        }, timeoutMs);
+    });
+    return Promise.race([promise, timeout]).finally(() => clearTimeout(timer));
+}
+
 /**
  * CPM messaging for embedded extension.
  * @implements {CPMMessagingBase}
@@ -64,7 +87,11 @@ export class CPMEmbeddedMessaging {
         this._queueSize += 1;
         const result = this._queue
             .then(() => {
-                return this.nativeMessaging.notify(method, params);
+                return withTimeout(
+                    this.nativeMessaging.notify(method, params),
+                    NATIVE_MESSAGE_TIMEOUT_MS,
+                    method,
+                );
             })
             .catch((e) => {
                 console.error('error in notification queue', e);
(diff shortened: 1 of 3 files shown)


shared/js/background/components/cpm-embedded-messaging.js
```diff
@@ -14,6 +14,29 @@ export const SETTING_CHECK_TTL = 10 * 1000; // 10s
 export const SITE_CHECK_TTL = 3 * 1000; // 3s (to be able to respond to Protections toggle changes)
 export const MAX_CACHE_SIZE = 100;
 
+// Upper bound for a single native message round-trip. Native messages are local IPC and normally
+// complete in milliseconds, but a dropped reply may leave the promise pending forever.
+export const NATIVE_MESSAGE_TIMEOUT_MS = 10 * 1000; // 10s
```
sammacbeth, line 19: This timeout seems excessive, given that a normal round trip should be milliseconds. Isn't 1s enough?

shared/js/background/components/cpm-embedded-messaging.js
```diff
@@ -64,7 +87,7 @@ export class CPMEmbeddedMessaging {
         this._queueSize += 1;
         const result = this._queue
             .then(() => {
-                return this.nativeMessaging.notify(method, params);
+                return withTimeout(this.nativeMessaging.notify(method, params), NATIVE_MESSAGE_TIMEOUT_MS, method);
```
sammacbeth, line 90: A queue of things timing out is going to block the messaging queue for a long time. e.g. if the user re-opens the browser and 20+ tabs reloading all ping. If the background never responds, then the message pipe will block for 200s+.

What was the reason to make messaging serial originally? Can't we just send these requests in parallel and then timeout the ones that fail?

sammacbeth (commented):
While this potentially unblocks a bottleneck from unreliable messaging, I'm thinking using serial messages is also unnecessary complexity.

shared/js/background/components/cpm-embedded-messaging.js
muodov: The Native Messaging API is an stdio pipe under the hood. I was not sure if sending parallel messages would be safe, so added this queue.

I think the browser reopen is not an issue: the extension is reloaded on app launch. The reports suggest that it stops working in a running instance - possibly around OS sleep/wake cycle.

shared/js/background/components/cpm-embedded-messaging.js
muodov: I was trying to be careful here, to let it heal eventually. Maybe we need more testing to check if we can reduce this (or remove the queue)

shared/js/background/components/cpm-embedded-messaging.js
sammacbeth: I would start with the assumption that messaging is safe, because it would be a bad implementation error if it were not. I'm starting to think that this messaging queue is unnecessary tech debt, and maybe making things worse rather than preventing issues. Its existence is based on the assumption that parallel messaging may break, which seems to be untested. The disadvantages are:
- breaks all messages when one fails (which we are trying to fix here)
- will slow down messages when there's a lot of message traffic (because we have to wait for each round trip)
- adds complexity to the implementation
- we're bypassing the queue anyway for config update messages, so seems we're not protecting against the case we think there may be issues with anyway?

Either let's prove that parallel messages are an issue, then we should move config updates onto the queue too, or just drop the queue and post all messages in parallel.

PR #3589, opened 2026-07-07
Drop the Puppeteer dependency

kzar:
We already use Playwright for the integration tests, so let's drop the Puppeteer
dependency. For the cases where we were still using Puppeteer, we can switch to
Playwright instead. Hopefully that should clear the way for us to start running
the unit tests against Firefox in the future as well, since we already have a
Firefox Playwright harness working for the integration tests.


sammacbeth (approved):
LGTM. Thanks for cleaning this up!

PR #3608, opened 2026-07-16
Remove cpm messaging queue

muodov:
Asana task: https://app.asana.com/1/137249556945/project/1201844467387842/task/1216559789388569?focus=true

**Reviewer:** @sammacbeth 

## Automated tests:
- [ ] Unit tests
- [ ] Integration tests

###### Reviewer Checklist:

###### PR Author Checklist:

---

> [!NOTE]
> **Medium Risk**
> Changes how all embedded CPM native traffic is ordered and timed; parallel messages and dropped queue metrics could affect native-side behavior or debugging, though errors/timeouts are still surfaced via diagnostics.
> 
> **Overview**
> **Removes the serialized native-message queue** in `CPMEmbeddedMessaging` so `_notify` and `_request` call the native layer directly and can run **in parallel** instead of blocking on prior messages.
> 
> Adds **`NATIVE_MESSAGE_TIMEOUT_MS` (20s)** and **`_withTimeout`** so hung native `notify`/`request` calls fail fast, record diagnostics, and (for timed-out cached requests) **do not cache late responses**. `refreshRemoteConfig` now uses the same timeout wrapper for `getResourceIfNew`.
> 
> Dashboard **`refreshCpmDashboardState` no longer sends `cpmQueueSize`** in `consentStatus`. Unit tests are updated for parallel behavior, timeouts, and concurrent cache misses; embedded extension version is bumped to **2026.7.16**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 8a8be4ea4265d292799a34734d4fa0c59befaeb6. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/shared/js/background/components/cpm-embedded-messaging.js
+++ b/shared/js/background/components/cpm-embedded-messaging.js
@@ -53,14 +53,14 @@ export class CPMEmbeddedMessaging {
     }
 
     /**
-     * Send a notification message to the native app.
+     * Send a fire-and-forget notification message to the native app. May reject if the native side didn't acknowledge the message.
      * @param {string} method
      * @param {Record<string, any>} params
      */
-    async _notify(method, params) {
+    async _notify(method, params, timeout = NATIVE_MESSAGE_TIMEOUT_MS) {
         const diagnosticsTabId = typeof params.tabId === 'number' ? params.tabId : null;
         try {
-            await this._withTimeout(this.nativeMessaging.notify(method, params), method);
+            await this._withTimeout(this.nativeMessaging.notify(method, params), timeout);
         } catch (e) {
             console.error('error sending native notification', e);
             this._recordDiagnosticsError(diagnosticsTabId, method);
@@ -76,9 +76,10 @@ export class CPMEmbeddedMessaging {
      * @param {string} [cacheKey]
      * @param {number} [ttl]
      * @param {number=} diagnosticsTabId
+     * @param {number} [timeout] - milliseconds to wait before abandoning the native request
      * @returns {Promise<any>}
      */
-    async _request(method, params, cacheKey, ttl, diagnosticsTabId) {
+    async _request(method, params, cacheKey, ttl, diagnosticsTabId, timeout = NATIVE_MESSAGE_TIMEOUT_MS) {
         if (cacheKey && ttl) {
             const cached = this._cache.get(cacheKey);
             if (cached && Date.now() - cached.time < ttl) {
@@ -87,7 +88,7 @@ export class CPMEmbeddedMessaging {
         }
 
         try {
-            const result = await this._withTimeout(this.nativeMessaging.request(method, params), method);
+            const result = await this._withTimeout(this.nativeMessaging.request(method, params), timeout);
             if (cacheKey && ttl) {
                 // evict the oldest entry if the cache is full
                 if (this._cache.size >= MAX_CACHE_SIZE) {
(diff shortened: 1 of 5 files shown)


shared/js/background/components/cpm-embedded-messaging.js
```diff
@@ -136,22 +144,32 @@ export class CPMEmbeddedMessaging {
...
-                          cpmErrors: cpmErrors.join(',').substring(0, 255),
-                      }
-                    : {}),
-                // add the current queue size
-                cpmQueueSize: this._queueSize,
-            },
-        });
+        const currentChain = this._dashboardNotifyChains.get(tabId);
+        const previous = currentChain ?? Promise.resolve();
+        const notification = previous.then(() =>
```
sammacbeth: I’m trying to understand why this is needed:
 - Is message delivery to native also out of order?
 - Are these messages destructive if out of order? Could message handling be adjusted to make that not the case? i.e. making the function commutative.

shared/js/background/components/cpm-embedded-messaging.js
```diff
@@ -89,37 +81,53 @@ export class CPMEmbeddedMessaging {
...
+            this._recordDiagnosticsError(diagnosticsTabId ?? null, method);
+        }
+    }
+
+    /**
+     * @param {Promise<any>} operation
+     * @param {string} method
+     * @returns {Promise<any>}
+     */
+    async _withTimeout(operation, method) {
```
sammacbeth: Having an extra parameter just to log the right thing on error is not the best API design. I’d suggest pushing this down the stack one level and adding a `timeout` option to `_request`.

I’m not sure the meaning of putting a timeout on `_notify` - it looks like the API as a promise that is confirmation of delivery, but when this times out then it’s unclear what that the contract is.

sammacbeth (changes requested):

shared/js/background/components/cpm-embedded-messaging.js
muodov: good question, I didn't test the delivery order. Let me check it.

muodov: - Yes, delivery to native can be out of order.
- we can make dashboard updates commutative within one page load, but if the page reloads, dashboard updates from the previous page could override the new ones. FrameIds don't help there because all top-level frames have frameId=0

kzar: > FrameIds don't help there because all top-level frames have frameId=0

@muodov One approach might be to have the content script generate a random ID for the page/frame when it loads, and include that in messages. Maybe that way the receiver could differentiate the messages after navigation/reload. I guess one thing to watch out for there might be the back/forward cache, e.g. if user presses back would the old random ID be used, and would that be an issue?

muodov: content script already has a random id for each frame, but we don't send it with each message at the moment. Cross-document races seem unlikely, so I think we can leave it out of scope for now.
@sammacbeth I traced the different flows, and I think the notification chains are mostly an overkill, except one case: `popupFound` and `done` messages are often coming in quick succession. Because every message carries the full dashboard state snapshot, popupFound event might override `consentManaged` flag if it's processed out of order.
Making the messages commutative is possible but would require native changes (some kind of merging of dashboard states).

shared/js/background/components/cpm-embedded-messaging.js
```diff
@@ -55,71 +53,77 @@ export class CPMEmbeddedMessaging {
...
-                console.error(`error in request queue for ${method}`, e);
-                this._recordDiagnosticsError(diagnosticsTabId ?? null, method);
-            })
-            .finally(() => {
-                this._queueSize -= 1;
-            });
-        this._queue = job;
-        return job;
+                this._cache.set(cacheKey, { time: Date.now(), value: result });
+            }
```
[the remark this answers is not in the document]
muodov: this is acceptable

shared/js/background/components/cpm-embedded-messaging.js
sammacbeth: I think we can skip that for the moment.

sammacbeth (approved):

shared/js/background/components/cpm-embedded-messaging.js
muodov: I tested it with macos, and couldn't repro this race

PR #3650, opened 2026-08-26
Chromium-embedded build target

sammacbeth:
**Reviewer:**

## Description:
We're creating a new build target for an extension version to be embedded in a chromium build. This PR sets up the initial build target, and sets up feature gating for things we don't want to include in this build (yet).

---

> [!NOTE]
> **Medium Risk**
> Introduces a new production-like build with changed telemetry/ATB behavior and browser `chrome.ddg` integration paths; existing Chrome/Firefox targets are mostly isolated behind `BUILD_TARGET` and Makefile filters.
> 
> **Overview**
> Adds a **`chromium-embedded`** MV3 build for shipping the extension inside DuckDuckGo Chromium, with a dedicated manifest, Makefile copy/bundle rules, and CI Playwright job (single shard).
> 
> The build **omits unreachable UI** (options, fire, feedback, new-tab-style pages), **skips autofill assets**, ships a slimmer JS/CSS bundle set (including **`background-chromium-embedded`**), still bundles smarter-encryption DNR rules, and **hides the dashboard settings cog** in popup CSS.
> 
> **Integration coverage** adds **`atb-disabled.spec.js`** (no post-install/ATB settings, no `atb` on search or ATB DNR rules) and updates **broken-site reports** so `atb` is absent on this target. Runtime behavior in the broader change set gates **ATB/install onboarding**, uses **Windows-oriented remote config / `navigator.duckduckgo` platform**, wires **CPM** through **`chrome.ddg`**, and registers **burn** cleanup for embedded state.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 1afd4d9e28420b43afd317ad10f9b4187baa9b24. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


krispyfries:
I think I've covered everything now so I'm setting this for review. I followed the conventions we discussed in the tech design for adding burn interop. The topmost branch of the ddg-chromium stack is [here](https://github.com/duckduckgo/ddg-cef-build-automation/tree/krispy/extension-burn-interop) and includes syncing to the latest extension in the tree.

shared/js/background/components/dashboard-messaging.js
```diff
@@ -139,6 +139,10 @@ export default class DashboardMessaging {
         const fireButtonData = {
             enabled: isFireButtonEnabled,
         };
-        return dashboardDataFromTab(tab, userData, fireButtonData);
+        // Read at call time: CPM is constructed after this component, and is
+        // absent from the MV2 builds.
+        const cpm = globalThis.components?.cpm;
```
sammacbeth: We shouldn't be using `globalThis` inside components. We should inject the CPM component in background.js. Simplest way would be to just set a property on dashboardMessaging after creating the cpm instance: `components.dashboardMessaging.cpm = components.cpm`.

shared/js/background/components/ddg-api-messaging.js
```diff
@@ -0,0 +1,63 @@
...
+ * implement the method returns `{ error: { code, message } }`, which is a reply,
+ * not a failure — callers check the reply's shape before using it.
+ *
+ * @param {string} featureName
+ * @param {string} method
+ * @param {Record<string, any>} [params]
+ * @param {number} [timeout] - milliseconds before the call is abandoned
+ * @returns {Promise<any>} the browser's reply, or null
+ */
+export async function sendToBrowser(featureName, method, params = {}, timeout = DDG_MESSAGE_TIMEOUT_MS) {
```
sammacbeth: Is there a reason we'd want this timeout and error suppression built in here? Seems like something the caller should be dealing with?

shared/js/background/onboarding.js
```diff
@@ -1,3 +1,6 @@
+// @ts-nocheck
```
sammacbeth: This change seems unrelated to the rest of the PR.

shared/js/background/components/burn.js
```diff
@@ -0,0 +1,33 @@
...
+ * persist to local storage, and the per-tab data held in memory, session
+ * storage and session DNR rules.
+ *
+ * The burn does not wait for this and may already have finished, so it is
+ * idempotent and safe to run late or twice. The event names no tab or origin,
+ * so everything goes; counts for tabs left open are rebuilt on next navigation.
+ *
+ * @param {'in-session' | 'on-exit' | 'on-startup'} trigger
+ */
+export function clearBurnedState(trigger) {
```
sammacbeth: Do we want to be able to do selective burns here? e.g. individual tab/window etc?

sammacbeth (commented):
@krispyfries Some questions and comments on your changes.

shared/js/background/components/ddg-api-messaging.js
krispyfries: I'll remove this with the explicit-contracts implementation :)

shared/js/background/onboarding.js
krispyfries: According to Claude, TSC fails without it. I'm not completely understanding its reasoning, so here's what it says:
```
Why it's needed:
- onboarding.js is listed in tsconfig.json's exclude. But exclude only controls which files tsc starts from; a file that an included file imports still gets checked.
- On main, the only importer was events.js, which is also excluded, so tsc never reached it.
- The ATB move put the import in components/setup-atb.js. That file falls under shared/js/background/**, so tsc now follows the import into onboarding.js.
```

shared/js/background/components/burn.js
krispyfries: Looking at burn in the browser I don't think this level of interop is available today - I'll leave this as-is for now and we can revisit when selective burns are available.

shared/js/background/components/dashboard-messaging.js
krispyfries: Fixed in https://github.com/duckduckgo/duckduckgo-privacy-extension/pull/3650/commits/db1dd1c05344896a1f54c19ed2193720797b1c40

shared/js/background/components/ddg-api-messaging.js
krispyfries: Removed with explicit contracts https://github.com/duckduckgo/duckduckgo-privacy-extension/pull/3650/commits/db1dd1c05344896a1f54c19ed2193720797b1c40

